PR #57, opened 2022-08-09
Add some tests for configuration

josecelano:
Hi @WarmBeer I have added some tests for the configuration.

### New dependencies

- `uuid`. I needed to create a unique id for a temporary config file like this `test_config_231863df-a1d5-4dfc-a2a9-a8b054c0b8cf.toml`.
- `serde_with`. For the `HttpTrackerConfig`, you are using a custom serializer, but the deserialization is not working. The config options `ssl_cert_path = ""` and `ssl_key_path = ""` are being deserialized like `Some("")` instead of `None`.

### Tests

- For now I have only added a few tests for saving and loading the configuration to and from the `config.toml` file.

### Changes

- I had to add some new attributes to some struts (PartialEq, Debug) to compare two configurations.
- I had to change the signature of two methods: `save_to_file` and `load_from_file`, to make those methods testable. I need to create the config file in a temporary file.

### Questions

1. I've added a tests for check default values only for one option (`should_have_a_default_value_for_the_log_level`). Do you think we should add one for each option or one test for all options, or maybe it does not makes sense to test default values directly?

2. Why are there to `impl Configuration` blocks?

3. I choose those test names in order to get this output when you execute them:

```s
 cargo nextest run
   Compiling torrust-tracker v2.3.0 (/home/josecelano/Documents/github/committer/josecelano/torrust/torrust-tracker)
    Finished test [optimized + debuginfo] target(s) in 12.91s
  Executable unittests src/lib.rs (target/debug/deps/torrust_tracker-d1fbcd787d97cc63)
  Executable unittests src/main.rs (target/debug/deps/torrust_tracker-8d19522d4b2f8cf0)
    Starting 7 tests across 2 binaries
        PASS [   0.002s] torrust-tracker config::configuration::should_be_loaded_from_a_config_file
        PASS [   0.002s] torrust-tracker config::configuration::should_be_saved_in_a_toml_config_file
        PASS [   0.002s] torrust-tracker tracker::key::tests::auth_key_from_buffer
        PASS [   0.002s] torrust-tracker tracker::key::tests::auth_key_from_string
        PASS [   0.002s] torrust-tracker tracker::key::tests::generate_expired_auth_key
        PASS [   0.002s] torrust-tracker config::configuration::should_have_a_default_value_for_the_log_level
        PASS [   0.002s] torrust-tracker tracker::key::tests::generate_valid_auth_key
------------
     Summary [   0.004s] 7 tests run: 7 passed, 0 skipped
```

You read the lines like this: `config::configuration::should_be_loaded_from_a_config_file`

I will add some more tests.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -107,9 +107,7 @@ impl Configuration {
             }
         }
     }
-}
 
-impl Configuration {
     pub fn default() -> Configuration {
         let mut configuration = Configuration {
             log_level: Option::from(String::from("info")),
--- a/src/config.rs
+++ b/src/config.rs
@@ -198,6 +198,26 @@ mod tests {
     }
 
     #[test]
+    fn configuration_should_have_default_values() {
+        use crate::Configuration;
+
+        let configuration = Configuration::default();
+
+        let toml = toml::to_string(&configuration).expect("Could not encode TOML value");
+
+        assert_eq!(toml, default_config_toml());
+    }
+
+    #[test]
+    fn configuration_should_contain_the_external_ip() {
+        use crate::Configuration;
+
+        let configuration = Configuration::default();
+
+        assert_eq!(configuration.external_ip, Option::Some(String::from("0.0.0.0")));
+    }
+
+    #[test]
     fn configuration_should_be_saved_in_a_toml_config_file() {
         use std::env;
         use crate::Configuration;
@@ -252,5 +272,14 @@ mod tests {
         let configuration = Configuration::load_from_file(&config_file_path).expect("Could not load configuration from file");
 
         assert_eq!(configuration, Configuration::default());
+    }
+
+    #[test]
+    fn configuration_error_could_be_displayed() {
+        use crate::ConfigurationError;
+
+        let error = ConfigurationError::TrackerModeIncompatible;
+
+        assert_eq!(format!("{}", error), "TrackerModeIncompatible");
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -176,7 +176,8 @@ impl Configuration {
     }
 }
 
-mod configuration {
+#[cfg(test)]
+mod tests {
 
     #[cfg(test)]
     fn default_config_toml() -> String {
@@ -216,7 +217,7 @@ mod configuration {
     }
 
     #[test]
-    fn should_have_a_default_value_for_the_log_level() {
+    fn configuration_should_have_a_default_value_for_the_log_level() {
         use crate::Configuration;
 
         let configuration = Configuration::default();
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -181,9 +169,115 @@ impl Configuration {
...
-    pub fn save_to_file(&self) -> Result<(), ()> {
+    pub fn save_to_file(&self, path: &str) -> Result<(), ()> {
         let toml_string = toml::to_string(self).expect("Could not encode TOML value");
-        fs::write("config.toml", toml_string).expect("Could not write to file!");
+        fs::write(path, toml_string).expect("Could not write to file!");
         Ok(())
     }
 }
+
+mod configuration {
```
mickvandijke: I think the module for tests should be just named `mod tests` regardless of what the name of the to be tested module is. The entire test module should then be annotated with `#[cfg(test)]`. All test functions in this module should then be annotated with `#[test]`.

Eg:
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test_func1() {
        assert!(true);
    }

    #[test]
    fn test_func2() {
        assert!(true);
    }
}
```

https://doc.rust-lang.org/book/ch11-03-test-organization.html

mickvandijke:
> ### Questions
> 1. I've added a tests for check default values only for one option (`should_have_a_default_value_for_the_log_level`). Do you think we should add one for each option or one test for all options, or maybe it does not makes sense to test default values directly?
> 2. Why are there two `impl Configuration` blocks?

1. I don't think it would be useful to test these default values.
2. I've never noticed this hehe, the functions can just be under one `impl Configuration`.

src/config.rs
josecelano: Fixed.

I forgot to add the annotation. Regarding the name, my original idea was to use "configuration_should" so that you do not have to repeat that prefix on each test, and I thought I was going to use more than one "tests" module (more than one unit to test). In the end, It was not the case. But if that's the case, maybe it makes sense to split the module into different submodules.

josecelano:
hi @WarmBeer It seems these two methods are not being used anymore:

```rust
    pub fn load(data: &[u8]) -> Result<Configuration, toml::de::Error> {
        toml::from_slice(data)
    }

    pub fn load_file(path: &str) -> Result<Configuration, ConfigurationError> {
        match std::fs::read(path) {
            Err(e) => Err(ConfigurationError::IOError(e)),
            Ok(data) => {
                match Self::load(data.as_slice()) {
                    Ok(cfg) => {
                        Ok(cfg)
                    }
                    Err(e) => Err(ConfigurationError::ParseError(e)),
                }
            }
        }
    }
```

josecelano:
I'm done with this PR.

mickvandijke (approved):

PR #67, opened 2022-08-12
Improved get_connection_id function to conform with the requirements listed in BEP-15

### LINKED ISSUE #62: Refactor: re-implement connection id for UDP tracker
BEP 15: https://www.bittorrent.org/beps/bep_0015.html

This is what the BEP 15 says about the connection ID:

> **UDP connections / spoofing**
> In the ideal case, only 2 packets would be necessary. However, it is possible to spoof the source address of a UDP packet. The tracker has to ensure this doesn't occur, so it calculates a value (connection_id) and sends it to the client. If the client spoofed it's source address, it won't receive this value (unless it's sniffing the network). The connection_id will then be send to the tracker again in packet 3. The tracker verifies the connection_id and ignores the request if it doesn't match. Connection IDs should not be guessable by the client. This is comparable to a TCP handshake and a syn cookie like approach can be used to storing the connection IDs on the tracker side. A connection ID can be used for multiple requests. A client can use a connection ID until one minute after it has received it. Trackers should accept the connection ID until two minutes after it has been send.

And this is the current implementation:

```rust
pub fn get_connection_id(remote_address: &SocketAddr) -> ConnectionId {
    match std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH) {
        Ok(duration) => ConnectionId(((duration.as_secs() / 3600) | ((remote_address.port() as u64) << 36)) as i64),
        Err(_) => ConnectionId(0x7FFFFFFFFFFFFFFF),
    }
}
```

_Originally posted by @josecelano in https://github.com/torrust/torrust-tracker/issues/60#issuecomment-1210961955_

mickvandijke:


src/protocol/utils.rs
```diff
@@ -1,9 +1,44 @@
 use std::{net::SocketAddr, time::SystemTime};
+use std::net::IpAddr;
 use aquatic_udp_protocol::ConnectionId;
 
+// todo: SALT should be randomly generated on startup
+const SALT: &str = "SALT";
```
josecelano, line 6: hi @WarmBeer should we do it in the constructor (TorrentTracker::new)? or maybe in the main:

```rust
// Initialize Torrust tracker
let salt = "SALT";
let tracker = match TorrentTracker::new(config.clone()) {
    Ok(tracker) => Arc::new(tracker, salt),
    Err(error) => {
        panic!("{}", error)
    }
};
```

I would also inject it `get_connection_id` so we can test it.

josecelano (changes requested):
I think this is a more robust implementation in terms of avoiding spoofing. But I'm not sure if generating a hash every 2 minutes could be too much CPU use. And you also have to generate it for verification.

Maybe we could fix the test for one call and compare the performance between both implementations. Maybe the other implementation is reasonable in terms of security.

src/protocol/utils.rs
mickvandijke: Hey @josecelano ,

I think the generation of the salt could be moved here:

https://github.com/torrust/torrust-tracker/blob/main/src/udp/server.rs#L27
```rust
pub async fn start(&self) {
        loop {
            let socket = self.socket.clone();
            let tracker = self.tracker.clone();

            tokio::select! {
                _ = tokio::signal::ctrl_c() => {
                    info!("Stopping UDP server: {}..", socket.local_addr().unwrap());
                    break;
                }
                Ok((valid_bytes, remote_addr)) = socket.recv_from(&mut data) => {

                    debug!("{:?}", payload);

                    let response = handle_packet(remote_addr, payload, tracker).await;
                    UdpServer::send_response(socket, remote_addr, response).await;
                }
            }
        }
    }
```

And since all connection id functions are only relevant for UDP, they can also be moved to the UDP module like you already said.

mickvandijke:
Hey @josecelano ,

I've added two benchmarks (using [Criterion](https://crates.io/crates/criterion)) to compare the speed of the two implementations for generating a connection id. 

The implementation that we currently have is of course very close to instant, so the benchmarks don't even show a time.

And this is the result I got for generating a connection id using Blake3 hashes:
Screenshot 2022-08-12 at 11 20 15

> Tested on a Macbook Air 2020, M1, 8GB

This means it would take 0.13 seconds to generate 1,000,000 connection id's using the new implementation. I think this is performant enough for our use case. Else we could potentially optimize the new implementation by skipping the Blake3 hashing and doing some light bit shifting on the same input.

mickvandijke:
You can run the benchmarks yourself by running `cargo bench` .

josecelano:
> Hey @josecelano ,
> 
> I've added two benchmarks (using [Criterion](https://crates.io/crates/criterion)) to compare the speed of the two implementations for generating a connection id.
> 
> The implementation that we currently have is of course very close to instant, so the benchmarks don't even show a time.
> 
> And this is the result I got for generating a connection id using Blake3 hashes: Screenshot 2022-08-12 at 11 20 15
> 
> > Tested on a Macbook Air 2020, M1, 8GB
> 
> This means it would take 0.13 seconds to generate 1,000,000 connection id's using the new implementation. I think this is performant enough for our use case. Else we could potentially optimize the new implementation by skipping the Blake3 hashing and doing some light bit shifting on the same input.

OK, so I guess we can merge it. We can add that explanation to the documentation and keep the other implementation in the docs. Maybe in the future, it could be a config option depending on whether you want to increase speed or security.

I rebased the base branch with your latest changes so you have to rebase this one too.

josecelano (approved):

PR #60, opened 2022-08-10
Re-implement connection ID generation for UDP tracker

josecelano:
TODO:

- [x] Add test for `verify_connection_id`
- [x] Move `get_connection_id` to `udp` module.
- [x] Move `current_time` to `clock` module.
- [x] Add "pepper" to the implementation: https://github.com/torrust/torrust-tracker/pull/60#issuecomment-1214172679
- [x] Write more documentation. Something like my comment [here](https://github.com/torrust/torrust-tracker/issues/62#issuecomment-1212077204). I think I'm going to use the Rust ```///``` inline documentation in this case.
- [x] Reimplement connection ID generation function using encryption as proposed by @da2ce7 [here](https://github.com/torrust/torrust-tracker/pull/60#issuecomment-1221020599).
- [x] Use the function `verify_connection_id` to verify the UDP announce request.
- [x] Refactor handlers to avoid creating the EncryptedConnectionIdIssuer multiple times.
- [x] Generate SALT when the server starts. We call it SERVER_SCRET now.
- [x] Generate a random secret. We are using a fixed one.
- [x] Refactor code to include some improvements from @WarmBeer [PR](https://github.com/torrust/torrust-tracker/pull/76).
- [x] Change the Blowfish crate. The one we are using is deprecated.
- [ ] Refactor code to include some improvements from @da2ce7  [PR](https://github.com/torrust/torrust-tracker/pull/77).

REMOVED FROM TODO:

- [x] Change one of the tests where I use "sleep" instead of mocking time. We need to implement a new serializer where we can inject the time. It's out of the scope of this PR and it's only ten milliseconds the delay introduced in the test execution.

PR #71, opened 2022-08-18
Memory consumption benchmark for connection ID generation (UDP tracker)

josecelano:
As we were discussing [here](https://github.com/torrust/torrust-tracker/pull/60#issuecomment-1219383694) we want to know if the new function to generate the connection ID (UDP tracker) consumes more memory (which leads to less request per second).

Following this [example](https://stackoverflow.com/a/30983834/3012842), I've created a new binary to get the memory consumption.

You have to run the binary with:

```s
cargo run --bin conn_id_benchmark -q
```

You have to comment all the tests except the sample you want to run:

```rust
/// cargo run --bin conn_id_benchmark -q
fn main() {

    // Total allocated: 99808 bytes
    mem_print();
    
    // Total allocated: 230880 bytes
    test_function();

    // Total allocated: 103008 bytes
    let address = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8080);
    test_old_get_connection_id(&address, 0);

    // Total allocated: 107104 bytes
    test_new_implementation()
}
```

The output is very big but I suppose the total memory allocated is this line:

```s
                            allocated         nmalloc   (#/sec)         ndalloc   (#/sec)       nrequests   (#/sec)           nfill   (#/sec)          nflush   (#/sec)
small:                          12896             364         0               0         0               0         0               4         0               0         0
large:                          94208               2         0               0         0               2         0               2         0               0         0
total:                         107104             366         0               0         0               2         0               6         0               0         0
```

The new implementation uses 4096 bytes (107104-103008) more than the previous one. We should check if that is what we need for each function call. It seems like a maximum memory block (4K), so it probably consumes less memory.

I've printed the result before returning the function, but even then, most of the memory used by the function could already be released.

I suppose I have to calculate the total consumed only by the function in order to know how much will decrease the number of requests we could handle per second with a certain amount of RAM.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -748,6 +748,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs_extra"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2022715d62ab30faffd124d40b76f4134a550a87792276512b18d63272333394"
+
+[[package]]
 name = "funty"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


josecelano:
If I print memory usage at the beginning and the end of the function call, I get:

- Old function: 113248 - 99808 = 13440 bytes (approx 13K)
- New function: 117344 - 99808 = 17536 bytes (approx 17K)

Which is the 4096 bytes difference I got from the total allocation difference.

1 Gigabytes = 1073741824 Bytes

Assuming a 1GB RAM Server and simplifying memory consumption:

- We can handle 1073741824/13440 = 79891,504761905 requests with the old implementation.
- We can handle 1073741824/17536 = 61230,715328467 requests with the new implementation.

That's 22% fewer requests.

PR #75, opened 2022-09-02
(#73) Add support for dynamic external ip in the config

### LINKED ISSUE #73: Add support for dynamic external ip in the config
IP Addresses can change over time, so it would be useful for tracker operators if they could set a dynamic DNS as external IP. Currently the external_ip config option only supports an IP address.

mickvandijke:
closes #73


da2ce7:
Closing as Stale

PR #77, opened 2022-09-07
Implement and compare all 3 connection cookie designs.

da2ce7:
Implemented all three connection cookie formats and compared them in the benchmark:

``` text
Running benches/bench_connection_cookie.rs (target/release/deps/bench_connection_cookie-588604952a2e2936)
Make Hashed Cookie      time:   [84.687 ns 84.792 ns 84.818 ns]

Check Hashed Cookie     time:   [85.215 ns 85.613 ns 85.713 ns]

Make Witness Cookie     time:   [80.925 ns 80.931 ns 80.932 ns]

Check Witness Cookie    time:   [78.492 ns 78.824 ns 78.907 ns]

Make Encrypted Cookie   time:   [100.87 ns 100.92 ns 100.93 ns]

Check Encrypted Cookie  time:   [126.95 ns 127.05 ns 127.48 ns]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,9 @@ hex = "0.4.3"
 percent-encoding = "2.1.0"
 binascii = "0.1"
 lazy_static = "1.4.0"
+blowfish = "0.9.1"
+cipher = "0.4.3"
+crypto-common = {version = "0.1.6", features = ["rand_core"] }
 
 openssl = { version = "0.10.41", features = ["vendored"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,13 @@ opt-level = 3
 lto = "fat"
 strip = true
 
+[dev-dependencies]
+criterion = "0.4"
+
+[[bench]]
+name = "bench_connection_cookie"
+harness = false
+
 [dependencies]
 tokio = { version = "1.7", features = [
     "rt-multi-thread",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,3 +38,34 @@ pub mod ephemeral_instance_keys {
         pub static ref RANDOM_SEED: Seed = Rng::gen(&mut ThreadRng::default());
     }
 }
+
+pub mod block_ciphers {
+
+    use blowfish::BlowfishLE;
+    use cipher::KeyInit;
+    use rand::rngs::StdRng;
+    use rand::SeedableRng;
+
+    pub type Cipher = BlowfishLE;
+
+    pub mod ephemeral_instance {
+        use super::*;
+        use crate::ephemeral_instance_keys::RANDOM_SEED;
+
+        lazy_static! {
+            pub static ref BLOCK_CIPHER_BLOWFISH: Cipher = <BlowfishLE as KeyInit>::new(&<BlowfishLE as KeyInit>::generate_key(
+                <StdRng as SeedableRng>::from_seed(*RANDOM_SEED)
+            ));
+        }
+    }
+
+    pub mod testing {
+        use super::*;
+
+        lazy_static! {
+            pub static ref TEST_BLOCK_CIPHER_BLOWFISH: Cipher = <BlowfishLE as KeyInit>::new(
+                &<BlowfishLE as KeyInit>::generate_key(<StdRng as SeedableRng>::from_seed([0u8; 32]))
+            );
+        }
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use log::info;
 use torrust_tracker::tracker::statistics::StatsTracker;
 use torrust_tracker::tracker::tracker::TorrentTracker;
-use torrust_tracker::{ephemeral_instance_keys, logging, setup, static_time, Configuration};
+use torrust_tracker::{block_ciphers, ephemeral_instance_keys, logging, setup, static_time, Configuration};
 
 #[tokio::main]
 async fn main() {
@@ -14,6 +14,11 @@ async fn main() {
 
     // Initialize the Ephemeral Instance Random Seed
     lazy_static::initialize(&ephemeral_instance_keys::RANDOM_SEED);
+
+    // Initialize the Block Ciphers
+    lazy_static::initialize(&block_ciphers::ephemeral_instance::BLOCK_CIPHER_BLOWFISH);
+    #[cfg(test)]
+    lazy_static::initialize(&block_ciphers::testing::TEST_BLOCK_CIPHER_BLOWFISH);
 
     // Initialize Torrust config
     let config = match Configuration::load_from_file(CONFIG_PATH) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "blowfish"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e412e2cd0f2b2d93e02543ceae7917b3c70331573df19ee046bcbc35e45e87d7"
+dependencies = [
+ "byteorder",
+ "cipher",
+]
+
+[[package]]
 name = "buf_redux"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -243,6 +253,16 @@ dependencies = [
  "serde 1.0.137",
  "time 0.1.44",
  "winapi",
+]
+
+[[package]]
+name = "cipher"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d1873270f8f7942c191139cb8a40fd228da6c3fd2fc376d7e92d47aa14aeb59e"
+dependencies = [
+ "crypto-common",
+ "inout",
 ]
 
 [[package]]
@@ -344,11 +364,12 @@ dependencies = [
 
 [[package]]
 name = "crypto-common"
-version = "0.1.3"
+version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57952ca27b5e3606ff4dd79b0020231aaf9d6aa76dc05fd30137538c50bd3ce8"
+checksum = "1bfb12502f3fc46cca1bb51ac28df9d618d813cdc3d2f25b9fe775a34af26bb3"
 dependencies = [
  "generic-array",
+ "rand_core",
  "typenum",
 ]
 
@@ -891,6 +912,15 @@ dependencies = [
  "autocfg",
  "hashbrown 0.11.2",
  "serde 1.0.137",
+]
+
+[[package]]
+name = "inout"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a0c10553d664a4d0bcff9f4215d0aac67a639cc68ef660840afe309b807bc9f5"
+dependencies = [
+ "generic-array",
 ]
 
 [[package]]
@@ -2335,8 +2365,11 @@ dependencies = [
  "aquatic_udp_protocol",
  "async-trait",
  "binascii",
+ "blowfish",
  "chrono",
+ "cipher",
  "config",
+ "crypto-common",
  "derive_more",
  "fern",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,12 +21,27 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "ansi_term"
@@ -61,9 +76,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "async-trait"
-version = "0.1.53"
+version = "0.1.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed6aa3524a2dfcf9fe180c51eae2b58738348d819517ceadf95789c51fff7600"
+checksum = "1e805d94e6b5001b651426cf4cd446b1ab5f319d27bab5c644f61de0a804360c"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -89,15 +104,15 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "base-x"
-version = "0.2.10"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc19a4937b4fbd3fe3379793130e42060d10627a360f2127802b10b87e7baf74"
+checksum = "4cbbc9d0964165b47557570cce6c952866c2678457aca742aafc9fb771d30270"
 
 [[package]]
 name = "base64"
-version = "0.13.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bigdecimal"
@@ -126,7 +141,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "clap",
+ "clap 2.34.0",
  "env_logger",
  "lazy_static",
  "lazycell",
@@ -169,9 +184,9 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.10.2"
+version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bf7fe51849ea569fd452f37822f606a5cabb684dc918707a0193fd4664ff324"
+checksum = "69cce20737498f97b993470a6e536b8523f0af7892a4f928cceb1ac5e52ebe7e"
 dependencies = [
  "generic-array",
 ]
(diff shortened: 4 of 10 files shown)


josecelano:
Cool @da2ce7!

I like having real speed metrics.

Do you want to include all the options in the final version or just record why we choose one of them?

da2ce7:
@josecelano rebased upon your latest version. :)

src/lib.rs
```diff
@@ -16,3 +16,31 @@ pub mod databases;
...
+            );
+        pub static ref SECRET_KEY: [u8; 32] = rand::Rng::gen(&mut rand::rngs::ThreadRng::default());
+    }
+}
+
+pub fn initialize_keys() {
+    // Initialize Torrust Blowfish Crypto Key
+    lazy_static::initialize(&keys::BLOWFISH_KEY);
+    lazy_static::initialize(&keys::TEST_BLOWFISH_KEY);
+}
```
josecelano: @da2ce7 

1. I think I would delay this initialization until needed. As I told @WarmBeer I tend to prefer keeping things as local as possible. These keys are only being used right now by the UDP tracker, and you can even not run it when you start the main app. We could move up as long as other parts of the application start using them. In fact, we could even consider using different keys.
2. I would keep this mod inside the `udp::connection`.
3. I would prefer to pass the value down instead of using a global variable.
4. I would prefer to keep the test code separated from the production code. I suppose that forces you to inject the value instead of using a global one.

src/main.rs
```diff
@@ -5,12 +5,14 @@ use torrust_tracker::logging;
...
 use torrust_tracker::tracker::tracker::TorrentTracker;
 
-extern crate crypto;
 
 #[tokio::main]
 async fn main() {
     const CONFIG_PATH: &str = "config.toml";
 
+    // Initialize Torrust Keys
+    torrust_tracker::initialize_keys();
```
josecelano: @da2ce7 As I mentioned before, for now, keys are only being used by the UPD tracker. If you do not start it, this initialization is needless.

src/protocol/clock.rs
```diff
@@ -1,20 +1,35 @@
...
-impl Clock for SystemClock {
-    fn now_as_timestamp(&self) -> u64 {
-        SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs()
+impl Clock for LocalClock {
+    fn now_as_timestamp() -> u64 {
+        let now = if cfg!(test) {
+            SystemTime::UNIX_EPOCH
+        } else {
+            SystemTime::now()
+        };
```
josecelano: @da2ce7 

The `Clock` is a very common service. I added it to be able to mock time in tests. I could end up being used a lot in other places. You could need to return a different value in other tests. This is like mocking the "now" always to the UNIX Epoch start.

On the other hand, adding test code in production code makes things harder to read. With other languages, I suppose you could overwrite the method `now_as_timestamp` with a child class: `FixClockAtUnixEpoch`, but in this case I guess you only have to alternatives:

- Do not pass the clock, just pass the time.
- Use a different implementation of the clock in the test code. We could create the `FixClock` and pass the time we want it to return when calling `now_as_timestamp`.

src/protocol/clock.rs
```diff
@@ -1,20 +1,35 @@
...
+
+        now.checked_add(period.to_owned())
+            .unwrap()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs()
+    }
```
josecelano: @da2ce7 I'm not sure if this function should belong to `Clock`. I think the only responsibility should be returning the current time. But I suppose we could keep it here and move it elsewhere if we start having too many helpers methods in the `Clock`.

josecelano: By the way, since we always return a timestamp, maybe we could simplify names:

```rust
pub trait TimestampClock {
    fn now() -> u64;
    fn time_after(period: &Duration) -> u64;
}
```

src/udp/connection/client_id.rs
```diff
@@ -1,104 +1,172 @@
...
+#[derive(PartialEq, Debug, Clone)]
+pub struct PlainClientId {
+    value: Vec<u8>,
+}
+
 #[derive(PartialEq, Debug, Clone)]
-pub struct ClientId {
-    value: [u8; 4],
+pub struct KeyedClientId {
+    value: Vec<u8>,
```
josecelano: @da2ce7 

For me, the `ClientId` is not an independent concept you can use outside of the "connection id witness" feature. I do not know why we should keep the "long" id if we only use 4 bytes to check that nobody is reusing someone else's connection id.

As I see it, the whole "connection id" is the witness. The witness contains two parts:

- A short client identifier (ClientId) to check that the use of the id comes from a legitimate client.
- The expiration date because we do not want to store it in our server.

I would keep the long version of the ClientId if we start suing elsewhere to identify clients uniquely.

src/udp/connection/client_id.rs
```diff
@@ -1,104 +1,172 @@
...
 
-impl<T: Default + Hasher> Make<T> for ClientId {
-    fn new(socket: &SocketAddr) -> Self {
-        let hash = <ClientId as Make<T>>::hash(socket);
-
-        let mut truncated_hash: [u8; 4] = [0u8; 4];
-        truncated_hash.copy_from_slice(&hash[..4]);
+impl<U> Digest<KeyedHash> for U {
+    fn hash(socket: &SocketAddr) -> Vec<u8> {
+        let secret_key: [u8; 32] = if cfg!(test) { [0u8; 32] } else { *SECRET_KEY };
```
josecelano: @da2ce7 I think `*SECRET_KEY` is a "hidden" dependency. I suppose you would not expect the `hash` method to use a key only by reading the signature.

In general, I suppose the main problem is the test code you have to add to the production code.

I've seen some frameworks which want to create shorter function calls, but they want to keep testability do something like this:

```
impl<U> Digest<KeyedHash> for U {
    fn hash(socket: &SocketAddr, secret_key: [u8; 32]) -> Vec<u8> {
        let secret_key: [u8; 32] = if cfg!(test) { [0u8; 32] } else { *SECRET_KEY };
    }
}

fn hash_facade(socket: &SocketAddr) {
    let secret_key: [u8; 32] = if cfg!(test) { [0u8; 32] } else { *SECRET_KEY };
    <PlainClientId as Digest<T>>::hash(socket, secret_key)
}
```

There is a global helper, the want you can change in tests.

More info: https://laravel.com/docs/9.x/facades

src/udp/connection/encoded_id.rs
```diff
@@ -0,0 +1,346 @@
...
+use cipher::{BlockDecrypt, BlockEncrypt, BlockSizeUser};
+
+use crate::keys::{BLOWFISH_KEY, TEST_BLOWFISH_KEY};
+use crate::protocol::clock::{Clock, LocalClock};
+
+type BlowfishArray = GenericArray<u8, <Blowfish<LittleEndian> as BlockSizeUser>::BlockSize>;
+
+use super::client_id::{HashedId, KeyedClientId, PlainClientId};
+
+pub trait EncodedId<T: HashedId> {
```
josecelano: @da2ce7 I like the name `EncodedId`. In my case I used `ConenctionIdData` becuase:

- It's not encoded. You have to encode it by calling `to_bytes()`. Maybe a better name would be `connection_id_data.encode()`.
- I wanted to make it more explicit that this is not the final connection id, it's the data you are going to put inside de connection id. In fact, you need to encode it again into an i64.

src/udp/connection/encoded_id.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        if &witness[0..4] != &encoded_id[0..4] {
+            Err("Bad Witness in Connection Id")
+        } else {
+            Ok(())
+        }
+    }
+}
+
+impl EncodedId<PlainClientId> for EncryptedEncodedId {
```
josecelano: @da2ce7 I would remove the other implementations if we finally only use the fastest version.

src/udp/connection/encoded_id.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mut block: BlowfishArray = BlowfishArray::from(id_clear_bytes);
+
+        let blowfish_key: &blowfish::Blowfish<LE> = if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };
+
+        <Blowfish<LE> as BlockEncrypt>::encrypt_block(blowfish_key, &mut block);
+
+        EncryptedEncodedId {
+            value: block.try_into().unwrap(),
+        }
+    }
```
josecelano:
```rust
fn new(client_id: PlainClientId, lifetime: Duration) -> Self {

    // 1.- Generate connection id data
    let expiry_u32: u32 = LocalClock::next_as_timestamp(&lifetime).try_into().unwrap();
    let id_clear: Vec<u8> = [&client_id.value()[0..4], &expiry_u32.to_le_bytes().as_slice()].concat();
    let id_clear_bytes: [u8; 8] = id_clear.try_into().unwrap();

    // 2.- Encrypt connection id data
    let mut block: BlowfishArray = BlowfishArray::from(id_clear_bytes);
    let blowfish_key: &blowfish::Blowfish<LE> = if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };
    <Blowfish<LE> as BlockEncrypt>::encrypt_block(blowfish_key, &mut block);

    EncryptedEncodedId {
        value: block.try_into().unwrap(),
    }
}
```

@da2ce7 

- I suppose we do not need the suffix here: `expiry_u32`
- I do not like the Rust verbosity just to concat two bytes arrays. I've been trying to get rid of it by extracting methods.

```rust
pub IdClear {
    bytes: [u8; 4];
    
    fn new(client_id: &PlainClientId, expiry: u32) -> Self {
      let client_id_witness = &client_id.value()[0..4];
      let expiry_as_bytes = expiry_u32.to_le_bytes();
      let id_clear: Vec<u8> = [client_id_witness, expiry_as_bytes].concat();
      let bytes: [u8; 8] = id_clear.try_into().unwrap();
      Self {
          bytes 
      }
    }

    fn bytes(&self) -> [u8; 8] {
        self.bytes
    }
}

fn new(client_id: PlainClientId, lifetime: Duration) -> Self {

    // 1.- Generate connection id data
    let expiry: u32 = LocalClock::next_as_timestamp(&lifetime).try_into().unwrap();
    let id_clear_bytes: [u8; 8] = IdClear::new(client_id, expiry).bytes()

    // 2.- Encrypt connection id data
    let mut block: BlowfishArray = BlowfishArray::from(id_clear_bytes);
    let blowfish_key: &blowfish::Blowfish<LE> = if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };
    <Blowfish<LE> as BlockEncrypt>::encrypt_block(blowfish_key, &mut block);

    EncryptedEncodedId {
        value: block.try_into().unwrap(),
    }
}
```

- As I mentioned before, I think this `if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };` should be moved out of production code.

src/udp/connection/encoded_id.rs
```diff
@@ -0,0 +1,346 @@
...
+        encoded_id: [u8; 8],
+        after_now: Option<Duration>,
+    ) -> Result<(), Self::Error> {
+        let mut block: BlowfishArray = BlowfishArray::from(encoded_id);
+
+        let blowfish_key: &blowfish::Blowfish<LE> = if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };
+
+        <Blowfish<LE> as BlockDecrypt>::decrypt_block(&blowfish_key, &mut block);
+
+        let client_id_bytes: [u8; 4] = block[0..4].try_into().unwrap();
```
josecelano: @da2ce7 I would rename this to:

```rust
let client_id_witness_bytes: [u8; 4] = block[0..4].try_into().unwrap();
```

and also where you use it:

```rust
let expected_client_id_witness_bytes = client_id.value()[0..4];
if &client_id_witness_bytes != &client_id.value()[0..4] {
    Err("Bad Witness in Connection Id")
} else {
    Ok(())
}
```

I would make it more explicit that we are using only 4 bytes from the client id to check that the connection id is not being used by other clients.

josecelano (changes requested):
hi @da2ce7 

I think we should not include the code we are not going to use and keep only the fastest implementation. We could keep this implementation in a different repo or branch to track why we use the encrypted version.

In general, I like the simplicity of the solution at the higher level. Still, having less abstraction forces us to see more Rust scaffolding code (for type casting) that leads to less readability, in my honest opinion. And also to less testability.

I would be happy with a similar approach but removing test code from production code and making it easier to mock global values when needed, with different values, instead of hardcoded fixed values.

src/udp/request_handler.rs
```diff
@@ -195,23 +190,25 @@ impl RequestHandler {
...
 
-        connection_id
+        ConnectionId(i64::from_le_bytes(encoded_id.value().to_owned()))
     }
 
-    pub fn is_connection_id_valid(&self, connection_id: &ConnectionId, remote_addr: &SocketAddr) -> bool {
-        let current_timestamp = current_timestamp();
+    pub fn is_connection_id_valid(connection_id: &ConnectionId, remote_addr: &SocketAddr) -> bool {
+        let client_id = KeyedClientId::new(&remote_addr);
+        let valid = HashedEncodedId::check(client_id, Duration::new(1, 0), connection_id.0.to_le_bytes(), None);
```
josecelano: @da2ce7 I think I would move the client_id_generating inside the connection mod. The client_id does make any sense outside that mod. It's not used at all. We could extract a function for that and move it to the connection mod:

```rust
fn check_connection:_id(connection_id: &ConnectionId, remote_addr: &SocketAddr) -> bool {
    let client_id = KeyedClientId::new(&remote_addr);
    let valid = HashedEncodedId::check(client_id, Duration::new(1, 0), connection_id.0.to_le_bytes(), None);
    valid
}
```

And the same for the generation.

Cargo.toml
```diff
@@ -36,7 +36,11 @@ hex = "0.4.3"
 percent-encoding = "2.1.0"
 binascii = "0.1"
 blake3 = "1.3.1"
-rust-crypto = "^0.2"
+lazy_static = "1.4.0"
+blowfish = "0.9.1"
+cipher = "0.4.3"
+byteorder = "1.4.3"
+crypto-common = {version = "0.1.6", features = ["rand_core"] }
```
josecelano: hi @da2ce7 

Is there a reason to use a different package for blowfish?

You are using:

- https://crates.io/crates/blowfish
- https://github.com/RustCrypto/block-ciphers
- 33 contributors
- 488 stars
- 102 forks
- Latest commit: 3 days ago
- It has a warning: "It's generally recommended not to use other cipher implementations in this repository besides the aes crate."

The base branch uses:

- https://crates.io/crates/rust-crypto
- https://github.com/DaGenix/rust-crypto/
- 50 contributors
- 1.2K stars
- 265 forks
- Latest commit: 6 years ago
- Used by 19.1K

src/lib.rs
da2ce7: > 1. I think I would delay this initialization until needed. As I told @WarmBeer I tend to prefer keeping things as local as possible. These keys are only being used right now by the UDP tracker, and you can even not run it when you start the main app. We could move up as long as other parts of the application start using them. In fact, we could even consider using different keys.

Generally, we want these keys to be initialized as early as possible. Global static values can be accessed by anyone at anytime, so we want to make sure they are ready before we do anything else.

For this project the overheard is so low that I don't see the problem about just initializing and forgetting.
- The problem I see is that this memory might be maliciously accessed (for example swapped to disk and accessed elsewhere), however this is less of a problem, as no persistent credentials are stored here; when this application is restart new keys are generated from random.

I think that we put in a todo to replace it with a double [arc_swap](https://docs.rs/arc-swap/latest/arc_swap/), that rotates these these keys out every hour or so.  However that is more complex than this simple solution. 

> 2. I would keep this mod inside the udp::connection.

This is unfortunately not possible. Lazy Static demands to be in the root of the crate.

> 3. I would prefer to pass the value down instead of using a global variable.

If this were some very large project with lots of crypto systems who's contexts should programmatically be kept separate from each other, this project doesn't have those design requirements.

> 4. I would prefer to keep the test code separated from the production code. I suppose that forces you to inject the value instead of using a global one.

I share the same preference; however this solution is most simple and works.

da2ce7:
> hi @da2ce7

Hello @josecelano 

> I think we should not include the code we are not going to use and keep only the fastest implementation. We could keep this implementation in a different repo or branch to track why we use the encrypted version.

I completely agree, I plan to remove all the other implementations. The reason that I have included all three implementations was:
1. To compare the implementations.
2. To see what one is the most complex, what one is the fastest, etc.
3. To see the problem from a higher-level.
4. To help me design for what sort of abstractions make sense.

> In general, I like the simplicity of the solution at the higher level. Still, having less abstraction forces us to see more Rust scaffolding code (for type casting) that leads to less readability, in my honest opinion. And also to less testability.

Yes I removed almost all the abstractions that you provided, as I'm still learning rust, I wish to use the standard library as much as possible. I do not want to re-invent the wheel.

> I would be happy with a similar approach but removing test code from production code and making it easier to mock global values when needed, with different values, instead of hardcoded fixed values.

This too I agree with. I'm not sure what sort of abstractions make the best sense yet. I plan to look into the [facade](https://laravel.com/docs/9.x/facades) design pattern as you recommended.

josecelano:
> Yes I removed almost all the abstractions that you provided, as I'm still learning rust, I wish to use the standard library as much as possible. I do not want to re-invent the wheel.
> 

Yes, I think that's something I'm still miserably failing. Specially with `std::convert::Into` trait:

"A value-to-value conversion that consumes the input value".
https://doc.rust-lang.org/std/convert/trait.Into.html

I need time to get used to those self-killing objects. I have been discussing with @WarmBeer [here](https://github.com/torrust/torrust-tracker/pull/76#pullrequestreview-1098929949).

src/udp/connection/encoded_id.rs
da2ce7: > @da2ce7 I like the name `EncodedId`. In my case I used `ConenctionIdData` becuase:
> 
>     * It's not encoded. You have to encode it by calling `to_bytes()`. Maybe a better name would be `connection_id_data.encode()`.
> 
>     * I wanted to make it more explicit that this is not the final connection id, it's the data you are going to put inside de connection id. In fact, you need to encode it again into an i64.

I have thought about the names, and I have come up with:

`ClientId -> ClientImage`
`ConnectionId -> ConnectionCookie`

Because the Id is really just a temporary lifetime cookie that we give to the clients.

src/lib.rs
mickvandijke: I much prefer to initialize the secret on UdpServer::start() as it isn’t needed anywhere else. The blowfish key can then be a field of PacketHandler.

In the future, we can use https://doc.rust-lang.org/cargo/reference/features.html to include/exclude UDP and HTTP code. It would be nice if all relevant code of these features is only part of the crate it is destined for.

This way, if someone wants to use Torrust without the UDP feature enabled. They can automatically exclude all crates/dependencies that are only needed for UDP (such as blowfish).

mickvandijke:
Very nice to see a comparison of all 3 implementations!

src/udp/connection/encoded_id.rs
josecelano: @da2ce7 I like `ConnectionCookie`. In fact, I did not like to mix it up with the connection_id in the protocol.

Regarding `ClientImage`, I do not understand it.

da2ce7:
@josecelano I have reimplemented the clock, now it takes the form a Duration, and has Into and From traits implemented for both a local clock, and a fixed clock.

josecelano:
> @josecelano I have reimplemented the clock, now it takes the form a Duration, and has Into and From traits implemented for both a local clock, and a fixed clock.

I will take a look. I've also been thinking about how to keep the global variables in your functions without having two different implementations for tests and production code. I mean things like this:

```rust
if cfg!(test) { &TEST_BLOWFISH_KEY } else { &BLOWFISH_KEY };
```

I found [this solution](https://stackoverflow.com/a/55152946/3012842).

Which is actually an old solution. Solution 2 in [this article](https://www.javaindeed.com/3-best-practices-to-test-a-code-that-calls-static-methods/).

The "facade" solution I mentioned is probably the more complex one.

da2ce7:
@josecelano and @WarmBeer 

I'm now felling that my version of the Clock Module is quite rusty:
https://github.com/da2ce7/torrust-tracker/blob/protocol-tests/src/protocol/clock.rs

What do you think?

josecelano:
> @josecelano and @WarmBeer
> 
> I'm now felling that my version of the Clock Module is quite rusty: https://github.com/da2ce7/torrust-tracker/blob/protocol-tests/src/protocol/clock.rs
> 
> What do you think?

@da2ce7 yes, it's much rustier. I've created a [PR with my feedback](https://github.com/da2ce7/torrust-tracker/pull/1).

da2ce7:
Requires rebase upon: #85

josecelano (approved):

josecelano:
hi @da2ce7 are you planning to merge this PR?

In general, I'm not fun of adding code that is not used in production but:

- We also have tests that are not used in production.
- We can consider it as documentation. In the future, we will know exactly why we are not using the other alternatives. It's also documentation for the UDP protocol.
- We could add an option to settings to use a different one. That would increase security because you do not even know which one the server is using.

da2ce7:
Closing, current implementation is good for the first release.

PR #78, opened 2022-09-08
moved RequestHandler creation

mickvandijke:


josecelano:
@WarmBeer look good to me. Thank you.

josecelano (approved):

mickvandijke:
@josecelano maybe this can be further refactored if you think the PacketHandler is too big now. I just quickly moved your code/logic to the right places and didn't change much further than that.

josecelano:
> @josecelano maybe this can be further refactored if you think the PacketHandler is too big now. I just quickly moved your code/logic to the right places and didn't change much further than that.

I think it's ok. I think there are like two levels: packets and requests. I see them as two different OSI levels:

| Levels        |
| ------------- |
| REQUESTS      |
| PACKETS      |

The low level handles the UPD packages, and the higher level handles the BitTorrent requests.

In the future, we could move some functions around to create those two levels. For example:

```rust
    async fn send_response(socket: Arc<UdpSocket>, remote_addr: SocketAddr, response: Response) {
        debug!("sending response to: {:?}", &remote_addr);

        let buffer = vec![0u8; MAX_PACKET_SIZE];
        let mut cursor = Cursor::new(buffer);

        match response.write(&mut cursor) {
            Ok(_) => {
                let position = cursor.position() as usize;
                let inner = cursor.get_ref();

                debug!("{:?}", &inner[..position]);
                UdpServer::send_packet(socket, &remote_addr, &inner[..position]).await;
            }
            Err(_) => { debug!("could not write response to bytes."); }
        }
    }
```

That function belongs to the server. It converts a Response (higher level) into a UDP Package (lower level). We could move the "packaging" logic to a new mod. I think it makes more sense to use `RequestHandler` instead of `PacketHandler` because most of the time, you are dealing with requests and responses. Maybe the server itself would be the `PacketHandler` because it's the higher service that directly speaks to the UDP port. A UDP server is basically a UPD packets handler.

But since there is not much code there and it's easy to refactor, we can do it when we agree on that view.

PR #81, opened 2022-09-09
Minimal Mockable Clock Implementation

### LINKED ISSUE #62: Refactor: re-implement connection id for UDP tracker

da2ce7:
This is a new clock that lends itself for easier testing, it as two forms: the working clock, that tracks the system time; and the stopped clock that is paused, and must be manually set on a thread-local basis.

When testing, the stopped clock is set to zero, or implicitly the unix epoch. When not testing the default value of the stopped clock is the time that the application was started.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,3 +16,14 @@ pub mod protocol;
 pub mod setup;
 pub mod tracker;
 pub mod udp;
+
+#[macro_use]
+extern crate lazy_static;
+
+pub mod static_time {
+    use std::time::SystemTime;
+
+    lazy_static! {
+        pub static ref TIME_AT_APP_START: SystemTime = SystemTime::now();
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2341,6 +2341,7 @@ dependencies = [
  "fern",
  "futures",
  "hex",
+ "lazy_static",
  "log",
  "openssl",
  "percent-encoding",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,13 @@ lto = "fat"
 strip = true
 
 [dependencies]
-tokio = { version = "1.7", features = ["rt-multi-thread", "net", "sync", "macros", "signal"] }
+tokio = { version = "1.7", features = [
+    "rt-multi-thread",
+    "net",
+    "sync",
+    "macros",
+    "signal",
+] }
 
 serde = { version = "1.0", features = ["derive"] }
 serde_bencode = "^0.2.3"
@@ -28,6 +34,7 @@ serde_with = "2.0.0"
 hex = "0.4.3"
 percent-encoding = "2.1.0"
 binascii = "0.1"
+lazy_static = "1.4.0"
 
 openssl = { version = "0.10.41", features = ["vendored"] }
 
--- a/src/api/server.rs
+++ b/src/api/server.rs
@@ -2,6 +2,7 @@ use std::cmp::min;
 use std::collections::{HashMap, HashSet};
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 
 use serde::{Deserialize, Serialize};
 use warp::{filters, reply, serve, Filter};
@@ -268,7 +269,7 @@ pub fn start(socket_addr: SocketAddr, tracker: Arc<TorrentTracker>) -> impl warp
             (seconds_valid, tracker)
         })
         .and_then(|(seconds_valid, tracker): (u64, Arc<TorrentTracker>)| async move {
-            match tracker.generate_auth_key(seconds_valid).await {
+            match tracker.generate_auth_key(Duration::from_secs(seconds_valid)).await {
                 Ok(auth_key) => Ok(warp::reply::json(&auth_key)),
                 Err(..) => Err(warp::reject::custom(ActionStatus::Err {
                     reason: "failed to generate key".into(),
--- a/src/databases/mysql.rs
+++ b/src/databases/mysql.rs
@@ -1,4 +1,5 @@
 use std::str::FromStr;
+use std::time::Duration;
 
 use async_trait::async_trait;
 use log::debug;
@@ -94,7 +95,7 @@ impl Database for MysqlDatabase {
                 "SELECT `key`, valid_until FROM `keys`",
                 |(key, valid_until): (String, i64)| AuthKey {
                     key,
-                    valid_until: Some(valid_until as u64),
+                    valid_until: Some(Duration::from_secs(valid_until as u64)),
                 },
             )
             .map_err(|_| database::Error::QueryReturnedNoRows)?;
@@ -187,7 +188,7 @@ impl Database for MysqlDatabase {
         {
             Some((key, valid_until)) => Ok(AuthKey {
                 key,
-                valid_until: Some(valid_until as u64),
+                valid_until: Some(Duration::from_secs(valid_until as u64)),
             }),
             None => Err(database::Error::InvalidQuery),
         }
@@ -197,7 +198,7 @@ impl Database for MysqlDatabase {
         let mut conn = self.pool.get().map_err(|_| database::Error::DatabaseError)?;
 
         let key = auth_key.key.to_string();
-        let valid_until = auth_key.valid_until.unwrap_or(0).to_string();
+        let valid_until = auth_key.valid_until.unwrap_or(Duration::ZERO).as_secs().to_string();
 
         match conn.exec_drop(
             "INSERT INTO `keys` (`key`, valid_until) VALUES (:key, :valid_until)",
--- a/src/databases/sqlite.rs
+++ b/src/databases/sqlite.rs
@@ -7,6 +7,7 @@ use r2d2_sqlite::SqliteConnectionManager;
 
 use crate::databases::database;
 use crate::databases::database::{Database, Error};
+use crate::protocol::clock::clock::SinceUnixEpoch;
 use crate::tracker::key::AuthKey;
 use crate::InfoHash;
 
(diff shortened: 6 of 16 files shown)


mickvandijke:
Hi @da2ce7 ,

I was looking through your PR and was wondering what the use of `StoppedClock` was. Is it just a timestamp? Is it a `WorkingClock`, but with fixed time?

da2ce7:
Hello @WarmBeer ,

> I was looking through your PR and was wondering what the use of `StoppedClock` was. Is it just a timestamp? Is it a `WorkingClock`, but with fixed time?

The stopped clock is an extension of the working clock, implementing the ´Time` trait, but also implementing the ´StoppedTime´ trait. This clock is for testing purposes (or could possibly be used for time stamping), where the time can be mocked.

da2ce7:
Hello @WarmBeer @josecelano ,

I've updated this clock with an minimal implementation of the connection cookie with test that make use of the clock's time mocking features.

I think that this basic implementation could serve as a good basis for further improvement. (i.e, It is just the basic hashing algorithm as suggested by Mick, using the build-in default hasher.).

src/lib.rs
```diff
@@ -16,3 +16,32 @@ pub mod protocol;
 pub mod setup;
 pub mod tracker;
 pub mod udp;
+
+#[macro_use]
+extern crate lazy_static;
+
+pub mod static_time {
```
josecelano, line 23: @da2ce7 I think I would call it `time_log` or `time_events` if the idea is to statically keep track of when some events happen in the app. It happens to be "static" but I think that's an implementation detail. It could be a service.

src/lib.rs
```diff
@@ -16,3 +16,32 @@ pub mod protocol;
...
+    use std::time::SystemTime;
+
+    lazy_static! {
+        pub static ref TIME_AT_APP_START: SystemTime = SystemTime::now();
+    }
+}
+
+pub mod static_keys {
```
josecelano: @da2ce7 for the same reason, I would call it `generic_keys` or something like that. Unless in this case `static` means they do not change. I think it would be better to explain what they are used for instead of their immutability property.

src/lib.rs
josecelano, line 23: I know this is a small mod, but I tend to prefer creating a new independent mod instead of using the generic `lib`. For me, it's very helpful to check the dir tree when I'm trying to understand a new project.

src/protocol/clock.rs
```diff
@@ -0,0 +1,227 @@
...
+            Some(time) => time.as_nanos().checked_div(period.as_nanos()),
+        }
+    }
+    fn now_sub_periods(duration: &Duration, period: &Duration) -> Option<u128> {
+        match Self::now_sub(duration) {
+            None => None,
+            Some(time) => time.as_nanos().checked_div(period.as_nanos()),
+        }
+    }
+}
```
josecelano: @da2ce7 Somehow, I would like to eliminate the prefix `now_` in all methods. But I do not know if it makes sense:

```rust
pub trait Time: Sized {
    fn now() -> SinceUnixEpoch;
}

pub trait TimeNow: Sized {
    fn add(duration: &Duration) -> Option<SinceUnixEpoch> {
        Time::now().checked_add(*duration)
    }
    fn sub(duration: &Duration) -> Option<SinceUnixEpoch> {
        Time::now().checked_sub(*duration)
    }
    fn periods(period: &Duration) -> Option<u128> {
        Time::now().as_nanos().checked_div(period.as_nanos())
    }
    fn add_periods(duration: &Duration, period: &Duration) -> Option<u128> {
        match Self::add(duration) {
            None => None,
            Some(time) => time.as_nanos().checked_div(period.as_nanos()),
        }
    }
    fn sub_periods(duration: &Duration, period: &Duration) -> Option<u128> {
        match Self::sub(duration) {
            None => None,
            Some(time) => time.as_nanos().checked_div(period.as_nanos()),
        }
    }
}
```

`TimeNow` could also be `CurrentTime`.

src/udp/connection_cookie.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        StoppedClock::local_add(&get_period_length()).unwrap();
+
+        let cookie_next = make_test_cookie(None);
+
+        assert_ne!(cookie_now, cookie_next)
+    }
+
+    #[test]
+    fn it_cookies_should_be_valid_for_this_and_next_periods() {
```
josecelano: @da2ce7 typo: `it_cookies_should_be_valid_for_this_and_next_periods` should be `it_should_be_valid_for_this_and_next_periods`

src/udp/connection_cookie.rs
```diff
@@ -0,0 +1,214 @@
...
+mod detail {
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
+    use std::time::Duration;
+
+    use super::*;
+    use crate::protocol::clock::{DefaultClock, Time};
+    use crate::protocol::crypto::keys::seeds::{DefaultSeed, Seed};
+
+    pub(crate) type Period = u128;
```
josecelano: @da2ce7 I think the name is too generic, but I do not come up with a better name. Maybe `FixedPeriod`. I think we need to add something to make it clear period starts at an exact minute. And they start counting at Unix Epoch.

- `FixedPeriod`
- `UnixEpochPeriod`

src/udp/connection_cookie.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub(crate) type Periods = u64;
+
+    pub(crate) type Cookie = u64;
+    pub(crate) type Seconds = u64;
+
+    pub(crate) const PERIODS_IN_LIFETIME: Periods = 60;
+
+    pub(crate) const PERIOD_LENGTH: Seconds = 2;
+
+    pub(crate) fn get_period_now() -> Period {
```
josecelano: @da2ce7 maybe `get_current_period`?

src/udp/connection_cookie.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub(crate) fn get_time_at_period_start(period: &Period) -> SinceUnixEpoch {
+        let periods_sec = (*period).saturating_mul(PERIOD_LENGTH as u128);
+        SinceUnixEpoch::from_secs(u64::try_from(periods_sec).unwrap())
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn get_time_at_period_end(period: &Period) -> SinceUnixEpoch {
+        get_time_at_period_start(&(period + 1))
+    }
+
```
josecelano: @da2ce7 could we extract a `Period` struct?:

```rust
pub struct Period(u128);

impl Period {
    pub fn get_current_period() { ... }
    pub get_start_time(&self) { ... }
    pub get_end_time(&self) { ... }
}
```

src/udp/connection_cookie.rs
```diff
@@ -0,0 +1,214 @@
...
+use aquatic_udp_protocol::ConnectionId;
+
+use crate::protocol::clock::SinceUnixEpoch;
+use crate::udp::ServerError;
+
+pub fn make_connection_cookie(remote_address: &SocketAddr) -> ConnectionId {
+    let period = detail::get_period_now();
+
+    let cookie = detail::make_cookie_internal(remote_address, &period);
```
josecelano: @da2ce7 can we move period generation inside?:

```rust
let cookie = detail::make_cookie_for_current_period(remote_address)
```

On the other hand, maybe we can rename the `detail` mod to something like:

```rust
let cookie = cookie_builder::make_for_current_period(remote_address)
```

josecelano (approved):

src/protocol/clock.rs
da2ce7: @josecelano Thanks, and done!

src/udp/connection_cookie.rs
da2ce7: @josecelano Thanks, and done!

src/udp/connection_cookie.rs
da2ce7: @josecelano Thanks, and done!

src/udp/connection_cookie.rs
da2ce7: @josecelano Thanks, and done!

josecelano (approved):

src/udp/connection_cookie.rs
da2ce7: Done :) I have extracted the period concept into the clock.

src/udp/connection_cookie.rs
da2ce7: Done :) Thanks :+1:

josecelano (approved):

da2ce7:
ACK cab093c026cf253ba6f7208211953a25b3d1c91b

PR #83, opened 2022-09-12
Clock: Time Extent, Maker, and Associated Traits

da2ce7:
`TimeExtent` is a simple structure that contains base increment (duration), and amount (a multiplier for the base the base increment). The resulting product of the base and the multiplier is the total duration of the time extent.
    
`TimeExtentClock` is a helper that generates time extents based upon the increment length and the time of the clock, this clock is specialised according to the `test` predicate with the `DefaultClockExtentMaker` type.

These modules will be used by the later connection cookie implementation.

### DIFF
--- a/src/protocol/clock/timeextent.rs
+++ b/src/protocol/clock/timeextent.rs
@@ -142,23 +142,23 @@ where
 #[derive(Debug)]
 pub struct TimeExtentMaker<const CLOCK_TYPE: usize> {}
 
-pub type WorkingClockTimeExtentMaker = TimeExtentMaker<{ ClockType::WorkingClock as usize }>;
-pub type StoppedClockTimeExtentMaker = TimeExtentMaker<{ ClockType::StoppedClock as usize }>;
+pub type WorkingTimeExtentMaker = TimeExtentMaker<{ ClockType::WorkingClock as usize }>;
+pub type StoppedTimeExtentMaker = TimeExtentMaker<{ ClockType::StoppedClock as usize }>;
 
-impl MakeTimeExtent<WorkingClock> for WorkingClockTimeExtentMaker {}
-impl MakeTimeExtent<StoppedClock> for StoppedClockTimeExtentMaker {}
+impl MakeTimeExtent<WorkingClock> for WorkingTimeExtentMaker {}
+impl MakeTimeExtent<StoppedClock> for StoppedTimeExtentMaker {}
 
 #[cfg(not(test))]
-pub type DefaultClockTimeExtentMaker = WorkingClockTimeExtentMaker;
+pub type DefaultTimeExtentMaker = WorkingTimeExtentMaker;
 
 #[cfg(test)]
-pub type DefaultClockTimeExtentMaker = StoppedClockTimeExtentMaker;
+pub type DefaultTimeExtentMaker = StoppedTimeExtentMaker;
 
 #[cfg(test)]
 mod test {
 
     use crate::protocol::clock::timeextent::{
-        checked_duration_from_nanos, DefaultClockTimeExtentMaker, Extent, MakeTimeExtent, TimeExtent, TimeExtentBase,
+        checked_duration_from_nanos, DefaultTimeExtentMaker, Extent, MakeTimeExtent, TimeExtent, TimeExtentBase,
         TimeExtentProduct,
     };
     use crate::protocol::clock::{DefaultClock, DurationSinceUnixEpoch, StoppedTime};
@@ -389,7 +389,7 @@ mod test {
         #[test]
         fn it_should_return_a_time_extent() {
             assert_eq!(
-                DefaultClockTimeExtentMaker::now(&TIME_EXTENT_VAL.increment).unwrap().unwrap(),
+                DefaultTimeExtentMaker::now(&TIME_EXTENT_VAL.increment).unwrap().unwrap(),
                 TimeExtent {
                     increment: TIME_EXTENT_VAL.increment,
                     amount: 0
@@ -399,21 +399,21 @@ mod test {
             DefaultClock::local_set(&DurationSinceUnixEpoch::from_secs(TIME_EXTENT_VAL.amount * 2));
 
             assert_eq!(
-                DefaultClockTimeExtentMaker::now(&TIME_EXTENT_VAL.increment).unwrap().unwrap(),
+                DefaultTimeExtentMaker::now(&TIME_EXTENT_VAL.increment).unwrap().unwrap(),
                 TIME_EXTENT_VAL
             );
         }
 
         #[test]
         fn it_should_return_none() {
-            assert_eq!(DefaultClockTimeExtentMaker::now(&TimeExtentBase::ZERO), None);
+            assert_eq!(DefaultTimeExtentMaker::now(&TimeExtentBase::ZERO), None);
         }
 
         #[test]
         fn it_should_return_tryfrom_int_error() {
             DefaultClock::local_set(&DurationSinceUnixEpoch::MAX);
             assert_eq!(
-                DefaultClockTimeExtentMaker::now(&TimeExtentBase::from_millis(1))
+                DefaultTimeExtentMaker::now(&TimeExtentBase::from_millis(1))
                     .unwrap()
                     .unwrap_err(),
                 u64::try_from(u128::MAX).unwrap_err()
@@ -429,12 +429,9 @@ mod test {
         #[test]
         fn it_should_return_a_time_extent() {
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_after(
-                    &TIME_EXTENT_VAL.increment,
-                    &Duration::from_secs(TIME_EXTENT_VAL.amount * 2)
-                )
-                .unwrap()
-                .unwrap(),
+                DefaultTimeExtentMaker::now_after(&TIME_EXTENT_VAL.increment, &Duration::from_secs(TIME_EXTENT_VAL.amount * 2))
+                    .unwrap()
+                    .unwrap(),
                 TIME_EXTENT_VAL
             );
         }
@@ -442,22 +439,19 @@ mod test {
         #[test]
         fn it_should_return_none() {
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_after(&TimeExtentBase::ZERO, &Duration::ZERO),
+                DefaultTimeExtentMaker::now_after(&TimeExtentBase::ZERO, &Duration::ZERO),
                 None
             );
 
             DefaultClock::local_set(&DurationSinceUnixEpoch::MAX);
-            assert_eq!(
-                DefaultClockTimeExtentMaker::now_after(&TimeExtentBase::ZERO, &Duration::MAX),
-                None
-            );
+            assert_eq!(DefaultTimeExtentMaker::now_after(&TimeExtentBase::ZERO, &Duration::MAX), None);
         }
 
         #[test]
         fn it_should_return_tryfrom_int_error() {
             DefaultClock::local_set(&DurationSinceUnixEpoch::MAX);
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_after(&TimeExtentBase::from_millis(1), &Duration::ZERO)
+                DefaultTimeExtentMaker::now_after(&TimeExtentBase::from_millis(1), &Duration::ZERO)
                     .unwrap()
                     .unwrap_err(),
                 u64::try_from(u128::MAX).unwrap_err()
@@ -474,7 +468,7 @@ mod test {
             DefaultClock::local_set(&DurationSinceUnixEpoch::MAX);
 
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_before(
+                DefaultTimeExtentMaker::now_before(
                     &TimeExtentBase::from_secs(u32::MAX as u64),
                     &Duration::from_secs(u32::MAX as u64)
                 )
@@ -490,12 +484,12 @@ mod test {
         #[test]
         fn it_should_return_none() {
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_before(&TimeExtentBase::ZERO, &Duration::ZERO),
+                DefaultTimeExtentMaker::now_before(&TimeExtentBase::ZERO, &Duration::ZERO),
                 None
             );
 
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_before(&TimeExtentBase::ZERO, &Duration::MAX),
+                DefaultTimeExtentMaker::now_before(&TimeExtentBase::ZERO, &Duration::MAX),
                 None
             );
         }
@@ -504,7 +498,7 @@ mod test {
         fn it_should_return_tryfrom_int_error() {
             DefaultClock::local_set(&DurationSinceUnixEpoch::MAX);
             assert_eq!(
-                DefaultClockTimeExtentMaker::now_before(&TimeExtentBase::from_millis(1), &Duration::ZERO)
+                DefaultTimeExtentMaker::now_before(&TimeExtentBase::from_millis(1), &Duration::ZERO)
                     .unwrap()
                     .unwrap_err(),
                 u64::try_from(u128::MAX).unwrap_err()
--- a/src/protocol/clock/timeextent.rs
+++ b/src/protocol/clock/timeextent.rs
@@ -1,0 +1,185 @@
+use std::num::{IntErrorKind, TryFromIntError};
+use std::time::Duration;
+
+use super::{ClockType, StoppedClock, TimeNow, WorkingClock};
+
+pub trait Extent: Sized + Default {
+    type Base;
+    type Multiplier;
+    type Product;
+
+    fn new(unit: &Self::Base, count: &Self::Multiplier) -> Self;
+
+    fn add(&self, add: Self::Multiplier) -> Result<Self, IntErrorKind>;
+    fn sub(&self, sub: Self::Multiplier) -> Result<Self, IntErrorKind>;
+
+    fn total(&self) -> Result<Option<Self::Product>, TryFromIntError>;
+    fn total_next(&self) -> Result<Option<Self::Product>, TryFromIntError>;
+}
+
+pub type TimeExtentBase = Duration;
+pub type TimeExtentMultiplier = u64;
+pub type TimeExtentProduct = TimeExtentBase;
+
+#[derive(Debug, Default, Hash, PartialEq, Eq)]
+pub struct TimeExtent {
+    pub increment: TimeExtentBase,
+    pub amount: TimeExtentMultiplier,
+}
+
+impl TimeExtent {
+    pub const fn from_sec(seconds: u64, amount: &TimeExtentMultiplier) -> Self {
+        Self {
+            increment: TimeExtentBase::from_secs(seconds),
+            amount: *amount,
+        }
+    }
+}
+
+impl Extent for TimeExtent {
+    type Base = TimeExtentBase;
+    type Multiplier = TimeExtentMultiplier;
+    type Product = TimeExtentProduct;
+
+    fn new(increment: &Self::Base, amount: &Self::Multiplier) -> Self {
+        Self {
+            increment: *increment,
+            amount: *amount,
+        }
+    }
+
+    fn add(&self, add: Self::Multiplier) -> Result<Self, IntErrorKind> {
+        match self.amount.checked_add(add) {
+            None => Err(IntErrorKind::PosOverflow),
+            Some(amount) => Ok(Self {
+                increment: self.increment,
+                amount,
+            }),
+        }
+    }
+
+    fn sub(&self, sub: Self::Multiplier) -> Result<Self, IntErrorKind> {
+        match self.amount.checked_sub(sub) {
+            None => Err(IntErrorKind::NegOverflow),
+            Some(amount) => Ok(Self {
+                increment: self.increment,
+                amount,
+            }),
+        }
+    }
+
+    fn total(&self) -> Result<Option<Self::Product>, TryFromIntError> {
+        match u32::try_from(self.amount) {
+            Err(error) => Err(error),
+            Ok(amount) => Ok(self.increment.checked_mul(amount)),
+        }
+    }
+
+    fn total_next(&self) -> Result<Option<Self::Product>, TryFromIntError> {
+        match u32::try_from(self.amount) {
+            Err(e) => Err(e),
+            Ok(amount) => match amount.checked_add(1) {
+                None => Ok(None),
+                Some(amount) => match self.increment.checked_mul(amount) {
+                    None => Ok(None),
+                    Some(extent) => Ok(Some(extent)),
+                },
+            },
+        }
+    }
+}
+
+pub trait MakeTimeExtent<Clock>: Sized
+where
+    Clock: TimeNow,
+{
+    fn now(increment: &TimeExtentBase) -> Option<Result<TimeExtent, TryFromIntError>> {
+        Clock::now()
+            .as_nanos()
+            .checked_div((*increment).as_nanos())
+            .map(|amount| match TimeExtentMultiplier::try_from(amount) {
+                Err(error) => Err(error),
+                Ok(amount) => Ok(TimeExtent::new(increment, &amount)),
+            })
+    }
+
+    fn now_add(increment: &TimeExtentBase, add_time: &Duration) -> Option<Result<TimeExtent, TryFromIntError>> {
+        match Clock::add(add_time) {
+            None => None,
+            Some(time) => {
+                time.as_nanos()
+                    .checked_div(increment.as_nanos())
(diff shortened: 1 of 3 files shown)


src/protocol/clock/clockperiod.rs
```diff
@@ -0,0 +1,92 @@
+use std::num::TryFromIntError;
+use std::time::Duration;
+
+use super::clock::{ClockType, StoppedClock, TimeNow, WorkingClock};
+use super::period::{self, Periods};
+
+pub trait ClockPeriods<T>: Sized
```
josecelano: @da2ce7 rename it to `Timer`?

src/protocol/clock/clockperiod.rs
```diff
@@ -0,0 +1,92 @@
...
+where
+    T: TimeNow,
```
josecelano: @da2ce7 shouldn't be `Time`? because you use `T::now()`.

src/protocol/clock/clockperiod.rs
josecelano: Maybe `Timer` is not a good name because it always starts at the same time (Unix Epoch).
Maybe better `UnixEpochTimer`. We use it to calculate intervals of customisable length from Unix Epoch.

https://wikidiff.com/interval/period

src/protocol/clock/period.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+pub trait PeriodsTotals: Periods {
+    fn total_time(&self) -> Result<Option<PeriodTotalTime>, TryFromIntError>;
+    fn total_time_end(&self) -> Result<Option<PeriodTotalTime>, TryFromIntError>;
+}
+
+#[derive(Debug, Default, Hash, PartialEq)]
+pub struct Period {
+    pub length: PeriodLength,
```
josecelano: @da2ce7 rename to `PeriodDuration`?

src/protocol/clock/period.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub count: PeriodCount,
```
josecelano: @da2ce7 what is a `PeriodCount`?

da2ce7:
Rebased after merge of #81

mickvandijke:
Where are the comments >:(

src/protocol/clock/period.rs
```diff
@@ -0,0 +1,108 @@
+use std::num::{IntErrorKind, TryFromIntError};
+use std::time::Duration;
+
+pub type PeriodLength = Duration;
+pub type PeriodCount = u64;
```
mickvandijke: Does `PeriodCount` have to be a `u64`? A `u32` would be a lot more convenient as argument for `checked_mul`, `checked_add`, etc..

src/protocol/clock/period.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub trait Periods: Sized + Default {
+    fn new(length: &PeriodLength, count: &PeriodCount) -> Self;
+
+    fn add_count(&self, add: PeriodCount) -> Result<Self, IntErrorKind>;
+    fn sub_count(&self, sub: PeriodCount) -> Result<Self, IntErrorKind>;
+}
+
+pub trait PeriodsTotals: Periods {
+    fn total_time(&self) -> Result<Option<PeriodTotalTime>, TryFromIntError>;
```
mickvandijke: Nested Results or Options such as `Result<Option<T>, E>` and `Option<Result<T, E>>` are not really nice to work with. We can avoid the parent Result type by changing `Period.count` to `u32`. Else maybe create a custom error enum and return `Result<PeriodTotalTime, PeriodError>` instead. :)

src/protocol/clock/period.rs
```diff
@@ -0,0 +1,108 @@
+use std::num::{IntErrorKind, TryFromIntError};
+use std::time::Duration;
+
+pub type PeriodLength = Duration;
+pub type PeriodCount = u64;
+pub type PeriodTotalTime = Duration;
+
+pub trait Periods: Sized + Default {
```
mickvandijke: Is this trait really necessary? I would prefer directly implementing these functions for `Period` (unless you're thinking of a future use-case of this trait).

da2ce7:
To do: as per discussion rename from "period" to "interval" terminology.

src/protocol/clock/extent.rs
```diff
@@ -0,0 +1,113 @@
...
+        assert_eq!(
+            Extent::from_sec(12, &12).product_current().unwrap().unwrap(),
+            Duration::from_secs(144)
+        );
+        assert_eq!(
+            Extent::from_sec(12, &12).product_next().unwrap().unwrap(),
+            Duration::from_secs(156)
+        );
+    }
+}
```
josecelano: @da2ce7 I would split this test into three tests:

- "the total duration for a default extend should be zero"
- "the total duration for an extend product should equal its size multiplied by its quantity"
- "an extend should calculate the product of the next extend after it"

And maybe I would add two more at the beginning:

- "an extend is a time period divided into a given quantity of same-sized intervals"
- "the default extend should be zero amount of zero-duration intervals"

src/protocol/clock/extent.rs
```diff
@@ -0,0 +1,113 @@
...
+    type IncrementIncrementsProduct;
+
+    fn new(size: &Self::Increment, quantity: &Self::IncrementsCount) -> Self;
+
+    fn add(&self, add: Self::IncrementsCount) -> Result<Self, IntErrorKind>;
+    fn sub(&self, sub: Self::IncrementsCount) -> Result<Self, IntErrorKind>;
+
+    fn product_current(&self) -> Result<Option<Self::IncrementIncrementsProduct>, TryFromIntError>;
+    fn product_next(&self) -> Result<Option<Self::IncrementIncrementsProduct>, TryFromIntError>;
+}
```
josecelano: @da2ce7 extracting a trait when only one struct implements it is sometimes considered an unneeded level of indirection. Although I also think the "Extend" and the "IncrementCounter" are the "thing" and set of operators for the "thing". What if you converted the trait into a struct like this:

```rust
pub struct Extent {
    pub size: ExtentIncrement,
    pub quantity: ExtentIncrementsCount,
}

impl Extent {
    fn new(size: &Self::Increment, quantity: &Self::IncrementsCount) -> Self {
        Self {
            size: *size,
            quantity: *quantity,
    }
}

pub struct ExtendCounter {}

impl ExtendCounter {
    fn add(&self, add: ExtentIncrementsCount) -> Result<Self, IntErrorKind> {
        match self.quantity.checked_add(add) {
            None => Err(IntErrorKind::PosOverflow),
            Some(quantity) => Ok(Self {
                size: self.size,
                quantity,
            }),
        }
    }

    fn sub(&self, sub: ExtentIncrementsCount) -> Result<Self, IntErrorKind> {
        match self.quantity.checked_sub(sub) {
            None => Err(IntErrorKind::NegOverflow),
            Some(quantity) => Ok(Self {
                size: self.size,
                quantity,
            }),
        }
    }

    fn product_current(&self) -> Result<Option<ExtentIncrementIncrementProduct>, TryFromIntError> {
        match u32::try_from(self.quantity) {
            Err(error) => Err(error),
            Ok(quantity) => Ok(self.size.checked_mul(quantity)),
        }
    }

    fn product_next(&self) -> Result<Option<ExtentIncrementIncrementProduct>, TryFromIntError> {
        match u32::try_from(self.quantity) {
            Err(e) => Err(e),
            Ok(quantity) => match quantity.checked_add(1) {
                None => Ok(None),
                Some(quantity) => match self.size.checked_mul(quantity) {
                    None => Ok(None),
                    Some(extent) => Ok(Some(extent)),
                },
            },
        }
    }
}
```

src/protocol/clock/extent.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn add(&self, add: Self::IncrementsCount) -> Result<Self, IntErrorKind> {
+        match self.quantity.checked_add(add) {
+            None => Err(IntErrorKind::PosOverflow),
+            Some(quantity) => Ok(Self {
+                size: self.size,
+                quantity,
+            }),
+        }
+    }
```
josecelano: @da2ce7 I think I would call this method `extend` or `expand` because with `add` it seems you are adding to structs of the same type.

For the `sub` it could be `collapse` or `shorten`.

src/protocol/clock/extentmaker.rs
```diff
@@ -0,0 +1,88 @@
+use std::num::TryFromIntError;
+use std::time::Duration;
+
+use super::extent::{Extent, ExtentIncrement, ExtentIncrementsCount, IncrementCounter};
+use super::{ClockType, StoppedClock, TimeNow, WorkingClock};
+
+pub trait ExtentFromClock<T>: Sized
```
josecelano: @da2ce7 I think it's a good name, but the first time I read `ExtendFromClock` it seemed you were going to create a child trait for another clock trait. In general, "Extent" and "Extend" are very similar for me, maybe because I'm not a native English speaker.

src/protocol/clock/extentmaker.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl ExtentFromClock<WorkingClock> for WorkingClockExtentMaker {}
+
+impl ExtentFromClock<StoppedClock> for StoppedClockExtentMaker {}
+
+#[cfg(not(test))]
+pub type DefaultClockExtentMaker = WorkingClockExtentMaker;
+
+#[cfg(test)]
+pub type DefaultClockExtentMaker = StoppedClockExtentMaker;
```
josecelano: @da2ce7 should not be `DefaultExtendClockMaker`? Because you are making an EntentClock.

josecelano (commented):
I've added more comments.

da2ce7:
> To do: as per discussion rename from "period" to "interval" terminology.

Hello @WarmBeer @josecelano,

I have rethought the period structure, and now I think that I have a naming that is good. I have decided on the name of 'extent', while it is already used in computer science (for a continuous range of blocks on a disk), it is somewhat adoptable, as blocks all have the same size.

By using simple mathematical terms: `Base`, `Multiplier`, and `Product` for Extent Trait Types, and clear labels for the `TimeExtent` specialization: `increment` and `amount`. I hope the code is much more intuitively comprehensible.

I think that this pull request is now ready for review.

src/protocol/clock/clockperiod.rs
da2ce7: Renamed to `Extent`

src/protocol/clock/clockperiod.rs
da2ce7: Renamed to `Clock`

src/protocol/clock/period.rs
da2ce7: If the increment is less than one second, you can easily overflow the u32 size, in this case u64 is better.

src/protocol/clock/period.rs
da2ce7: Hmm. I thought that it is better to just pass through the error to the caller to deal with it there. (if they want to).

src/protocol/clock/period.rs
da2ce7: Now that I have called it `Extent`, I have made it fully generic and may extracted and reused for other purposes in the future.

src/protocol/clock/period.rs
da2ce7: Renamed to `TimeExtentBase` with the variable called `increment`.

src/protocol/clock/period.rs
da2ce7: A count of periods, I suppose. Anyway renamed to `TimeExtentMultiplier`, with the variable: `amount`.

src/protocol/clock/extent.rs
da2ce7: Okay, I agree, some more work improving the tests would be good :)

src/protocol/clock/extent.rs
da2ce7: Since the trait is fully generic now, `Extent`, it could be reused for other situations where you have a count of a unit.

I also think that this trait improves the readability, as it explains the general interface.

src/protocol/clock/extent.rs
da2ce7: Hmm... This is something that I need to consider.

src/protocol/clock/extentmaker.rs
da2ce7: Renamed to `MakeTimeExtent<Clock>`, that I hope is more intuitive.

src/protocol/clock/extentmaker.rs
da2ce7: I've updated the names: `DefaultClockTimeExtentMaker` is uses the default clock for the `TimeExtentMaker` implementation.

src/protocol/clock/timeextent.rs
```diff
@@ -0,0 +1,185 @@
...
+pub struct TimeExtentMaker<const CLOCK_TYPE: usize> {}
+
+pub type WorkingClockTimeExtentMaker = TimeExtentMaker<{ ClockType::WorkingClock as usize }>;
+pub type StoppedClockTimeExtentMaker = TimeExtentMaker<{ ClockType::StoppedClock as usize }>;
+
+impl MakeTimeExtent<WorkingClock> for WorkingClockTimeExtentMaker {}
+impl MakeTimeExtent<StoppedClock> for StoppedClockTimeExtentMaker {}
+
+#[cfg(not(test))]
+pub type DefaultClockTimeExtentMaker = WorkingClockTimeExtentMaker;
```
josecelano: @da2ce7 can we just use `DefaultTimeExtentMaker`? We can hide the implementation detail. Especially because we do not have any other implementation. The name is confusing for me probably because I'm not an English native speaker. It could be:

- The `Default` `ClockTimeExtentMaker`
- or the `TimeExtentMaker` which uses the `DefaultClock`

Following your conventions, it should be `DefaultDeafultClockTimeEntentMaker` because it's the default implementation of the TimeExtentMaker using the DefaultClock.

I would use: 

- `DefaultTimeExtentMaker` (removing information about where the time comes from)
- or `DefaultTimeExtentMakerUsingDefaultClock`

src/protocol/clock/timeextent.rs
```diff
@@ -0,0 +1,185 @@
...
+        );
+
+        DefaultClock::local_set(&DurationSinceUnixEpoch::from_secs(12387687123));
+
+        assert_eq!(
+            DefaultClockTimeExtentMaker::now(&Duration::from_secs(2)).unwrap().unwrap(),
+            TimeExtent::from_sec(2, &6193843561)
+        );
+    }
+}
```
josecelano: @da2ce7 I would try to have only one assert per test (if possible). Some people could not be familiar with some terms like "extent" in a software context, so having a nice set of test could help to understand it:

For the `TimeExtent`:

- An extend is a time period (duration) which is divided into a number of regular fixed increments.
- The default time extent contains zero increments of zero duration.
- A time extent can be extended with more increments.
- A time extent can be shortened by some increments.


For the `DefaultClockTimeExtentMaker`:

- A time extent maker can make the time extent for the period since the Unix Epoch by using the default clock.

src/protocol/clock/timeextent.rs
```diff
@@ -0,0 +1,185 @@
...
+        assert_eq!(
+            TimeExtent::from_sec(12, &12).total_next().unwrap().unwrap(),
+            Duration::from_secs(156)
+        );
+    }
+
+    #[test]
+    fn it_should_make_the_current_extent() {
+        assert_eq!(
+            DefaultClockTimeExtentMaker::now(&Duration::from_secs(2)).unwrap().unwrap(),
```
josecelano: @da2ce7 I think this would be clearer with something like:

```rust
DefaultClockTimeExtentMaker::now(&TimeExtentBase::from_secs(2)).unwrap().unwrap()
```

josecelano (approved):
@da2ce7 I've only added a couple of comments.

src/protocol/clock/extent.rs
da2ce7: I've decided on `increase` and `decrease`.

src/protocol/clock/timeextent.rs
da2ce7: I think that `DefaultTimeExtentMaker` is better. :)

src/protocol/clock/timeextent.rs
da2ce7: Fixed. :)

src/protocol/clock/timeextent.rs
da2ce7: I have regrouped the tests, now I think that all the asserts within a test check the same function. This should be clearer.

src/protocol/clock/timeextent.rs
```diff
@@ -149,37 +157,358 @@ pub type DefaultClockTimeExtentMaker = StoppedClockTimeExtentMaker;
...
+            );
+            assert_eq!(
+                checked_duration_from_nanos(u64::MAX as u128).unwrap(),
+                Duration::from_nanos(u64::MAX)
+            );
+            assert_eq!(
+                checked_duration_from_nanos(TIME_EXTENT_VAL.amount as u128 * NANOS_PER_SEC as u128).unwrap(),
+                Duration::from_secs(TIME_EXTENT_VAL.amount)
+            );
+        }
```
josecelano: @da2ce7 reanme to `it should return duration from nanos where they are in range`?

src/protocol/clock/timeextent.rs
da2ce7: Done :)

src/protocol/clock/timeextent.rs
```diff
@@ -149,37 +157,358 @@ pub type DefaultClockTimeExtentMaker = StoppedClockTimeExtentMaker;
...
+
+        #[test]
+        fn it_should_return_tryfrom_int_error() {
+            assert_eq!(
+                checked_duration_from_nanos(u128::MAX).unwrap_err(),
+                u64::try_from(u128::MAX).unwrap_err()
+            );
+        }
+    }
```
josecelano: @da2ce7 rename to `it should fails returning a duration from nanos when nanos ar out of duration range`?

src/protocol/clock/extent.rs
da2ce7: Mostly Done, Still have grouped some asserts to make things more concise.

src/protocol/clock/timeextent.rs
da2ce7: Done :)

josecelano:
@da2ce7 these are the test for the new features:

```s
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::fn_checked_duration_from_nanos::it_should_return_a_duration
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::fn_checked_duration_from_nanos::it_should_return_tryfrom_int_error
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now::it_should_return_a_time_extent
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now::it_should_return_none
        PASS [   0.286s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now::it_should_return_tryfrom_int_error
        PASS [   0.285s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_after::it_should_return_a_time_extent
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_after::it_should_return_none
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_after::it_should_return_tryfrom_int_error
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_before::it_should_return_a_time_extent
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_before::it_should_return_none
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_decrease::it_should_return_an_negitive_overflow
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::make_time_extent_now_before::it_should_return_tryfrom_int_error
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_decrease::it_should_return_decreased
        PASS [   0.362s] torrust-tracker protocol::clock::timeextent::test::time_extent_from_sec::it_should_make_time_extent
        PASS [   0.370s] torrust-tracker protocol::clock::timeextent::test::time_extent_default::it_should_make_time_extent
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::time_extent_increase::it_should_postive_overflow
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::time_extent_increase::it_should_return_increased
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_total::it_should_return_none
        PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::time_extent_new::it_should_make_time_extent
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_total::it_should_return_total
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_total::it_should_return_tryfrom_int_error
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_total_next::it_should_get_the_time_extent_total_next
        PASS [   0.008s] torrust-tracker protocol::clock::timeextent::test::time_extent_total_next::it_should_return_none
```

When I'm reviewing or writing tests, I always check that reading this output makes sense without looking at the implementation. For me, in the case of tests, namespaces (mod names) are also important. For instance. In this case:

```s
PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::fn_checked_duration_from_nanos::it_should_return_tryfrom_int_error
```

I would like to read:

```s
PASS [   0.009s] torrust-tracker protocol::clock::timeextent::test::checked_duration_from_nanos_function::should_fails_when_nanos_are_out_of_the_duration_range
```

The concrete error is an implementation detail. I should not change the test name is you thrown a different error type. And with the new name I know when the function throws that error.

Same for other tests.

I like reading this test output like a documentation to know what the strcut/function/mod/ does not only to protect from regression bugs.

UPDATE:

And I would use snake case:

```
PASS [   0.009s] torrust-tracker protocol::clock::time_extent::test::checked_duration_from_nanos_function::should_fails_when_nanos_are_out_of_the_duration_range
```

Notice `time_extent` instead of `timeextent`.

da2ce7:
@josecelano reworded the last comments to include you as the co-author. :)

josecelano (approved):
@da2ce7 wonderful!!!

With a little bit of indent and order in the output of the tests, you get this beautiful documentation for the mod:

```s
clock::time_extent::test::
    fn_checked_duration_from_nanos::
        it_should_be_the_same_as_duration_implementation_for_u64_numbers
        it_should_fail_for_numbers_that_are_too_large
        it_should_give_zero_for_zero_input
        it_should_work_for_some_numbers_larger_than_u64

    make_time_extent::
        fn_now::
            it_should_fail_for_zero
            it_should_give_a_time_extent
            it_should_fail_if_amount_exceeds_bounds

        fn_now_after::
            it_should_fail_for_zero
            it_should_fail_if_amount_exceeds_bounds
            it_should_give_a_time_extent

        fn_now_before::
            it_should_fail_for_zero
            it_should_fail_if_amount_exceeds_bounds
            it_should_give_a_time_extent
    
    time_extent::
        fn_increase::
            it_should_increase
            it_should_not_increase_for_zero
            it_should_fail_when_attempting_to_increase_beyond_bounds

        fn_decrease::
            it_should_decrease
            it_should_not_decrease_for_zero
            it_should_fail_when_attempting_to_decrease_beyond_bounds
            
        fn_default::
            it_should_default_initialize_to_zero

        fn_from_sec::
            it_should_make_from_seconds
            it_should_make_empty_for_zero

        fn_new::
            it_should_make_empty_for_zero
            it_should_make_new

        fn_total::
            it_should_be_zero_for_zero
            it_should_fail_when_product_is_too_large
            it_should_fail_when_too_large
            it_should_give_a_total

        fn_total_next::
            it_should_be_zero_for_zero
            it_should_fail_when_product_is_too_large
            it_should_fail_when_too_large
            it_should_give_a_total
```

We should write a script to format the output like that. Like other testing frameworks.

For example: https://github.com/Nautilus-Cyberneering/git-queue/actions/runs/3094430462/jobs/5007796148#step:4:6

src/protocol/clock/timeextent.rs
da2ce7: Done: `it_should_be_the_same_as_duration_implementation_for_u64_numbers` and `it_should_work_for_some_numbers_larger_than_u64` and `it_should_fail_for_numbers_that_are_too_large`

src/protocol/clock/timeextent.rs
da2ce7: Done.

da2ce7:
ACK e785a7fd8e0d0a393b1d312d07178ebef43dda25

PR #84, opened 2022-09-12
Crypto: Ephemeral Instance Random Seeds with Keepers

da2ce7:
This module makes use of [lazy_static](https://docs.rs/lazy_static/latest/lazy_static/), to initialize an instance ephemeral random seed for the server to use on startup.

This seed can be used to derive other instance ephemeral keys. The seed is forgotten upon shutdown of the server.

To help with testing, a `DefaultSeed` helping structure that provides a zeroed seed when testing, and the random seed when not testing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,3 +27,14 @@ pub mod static_time {
         pub static ref TIME_AT_APP_START: SystemTime = SystemTime::now();
     }
 }
+
+pub mod ephemeral_instance_keys {
+    use rand::rngs::ThreadRng;
+    use rand::Rng;
+
+    pub type Seed = [u8; 32];
+
+    lazy_static! {
+        pub static ref RANDOM_SEED: Seed = Rng::gen(&mut ThreadRng::default());
+    }
+}
--- a/src/protocol/crypto.rs
+++ b/src/protocol/crypto.rs
@@ -1,0 +1,98 @@
+pub mod keys {
+
+    pub mod seeds {
+        use self::detail::DEFAULT_SEED;
+        use crate::ephemeral_instance_keys::{Seed, RANDOM_SEED};
+
+        pub trait SeedKeeper {
+            type Seed: Sized + Default + AsMut<[u8]>;
+            fn get_seed() -> &'static Self::Seed;
+        }
+
+        pub struct InstanceSeed;
+        pub struct DefaultSeed;
+
+        impl SeedKeeper for InstanceSeed {
+            type Seed = Seed;
+
+            fn get_seed() -> &'static Self::Seed {
+                &RANDOM_SEED
+            }
+        }
+
+        impl SeedKeeper for DefaultSeed {
+            type Seed = Seed;
+
+            #[allow(clippy::needless_borrow)]
+            fn get_seed() -> &'static Self::Seed {
+                &DEFAULT_SEED
+            }
+        }
+
+        #[cfg(test)]
+        mod tests {
+            use super::detail::ZEROED_TEST_SEED;
+            use super::{DefaultSeed, InstanceSeed, SeedKeeper};
+            use crate::ephemeral_instance_keys::Seed;
+
+            pub struct ZeroedTestSeed;
+
+            impl SeedKeeper for ZeroedTestSeed {
+                type Seed = Seed;
+
+                #[allow(clippy::needless_borrow)]
+                fn get_seed() -> &'static Self::Seed {
+                    &ZEROED_TEST_SEED
+                }
+            }
+
+            #[test]
+            fn the_default_seed_and_the_zeroed_seed_should_be_the_same_when_testing() {
+                assert_eq!(DefaultSeed::get_seed(), ZeroedTestSeed::get_seed())
+            }
+
+            #[test]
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -16,3 +16,24 @@ pub mod protocol;
...
+
+#[macro_use]
+extern crate lazy_static;
+
+pub mod static_keys {
+    pub mod static_seeds {
+        use rand::rngs::ThreadRng;
+        use rand::Rng;
+
+        pub type SeedSize = [u8; 32];
```
josecelano: @da2ce7 why not just `Seed`? I would expect `SeedSize` to be the "size" of the seed.

src/lib.rs
```diff
@@ -16,3 +16,24 @@ pub mod protocol;
 pub mod setup;
 pub mod tracker;
 pub mod udp;
+
+#[macro_use]
+extern crate lazy_static;
+
+pub mod static_keys {
```
josecelano: @da2ce7 I think I commented on the name in another PR. If `static` means they do not change, the name is OK. If, in this context, `static` means they are global, then do not like it. I would prefer something that reveals the intention. Maybe:

- `generic_keys`.
- `global_generic_keys`.

Same for `static_seeds`.

I think you describe them as "generic keys that you can use anywhere in the app".

On the other hand, why do you need the `static_keys` mod if you have the `crypto` mod?

src/protocol/crypto.rs
```diff
@@ -0,0 +1,82 @@
+pub mod keys {
+
+    pub mod seeds {
+        use crate::static_keys::static_seeds::{SeedSize, INSTANCE_SEED};
+
+        pub trait Seed {
+            type Seed: Sized + Default + AsMut<[u8]>;
+            fn seed() -> &'static Self::Seed;
+        }
```
josecelano: @da2ce7 I think I would call the trait `Seeder` and the method `new_seed`:

```rust
pub trait Seeder {
    type Seed: Sized + Default + AsMut<[u8]>;
    fn new_seed() -> &'static Self::Seed;
}
```

The word `Seed` is ambiguous. Itis a trait, a type and a method name.

josecelano: I've just seen it's not a builder. It's a kind of Singleton where you have two instances instead of just one. In that case, maybe it makes more sense:

```rust
pub trait SeedKeeper {
    type Seed: Sized + Default + AsMut<[u8]>;
    fn get_seed() -> &'static Self::Seed;
}

impl SeedKeeper for InstanceSeed {
    type Seed = SeedSize;
    fn get_seed() -> &'static Self::Seed {
        &*INSTANCE_SEED
    }
}

impl SeedKeeper for DefaultSeed {
    type Seed = SeedSize;
    fn get_seed() -> &'static Self::Seed {
        &*self::detail::DEFAULT_SEED
    }
}
```

I was thinking of other alternatives like `SeedChain` or `SeedRing` but they would make sense if you use the same instance to return both seeds.

da2ce7:
Rebased after merge of #81

josecelano (changes requested):
hi @da2ce7, I reviewed this PR and made some comments.

src/lib.rs
da2ce7: Renamed to `SeedData` to be more descriptive.

src/lib.rs
da2ce7: In rust, `static` refers to the fact that the memory location is fixed, not that the data at that memory location is immutable.

Originally I chose to name the module `static_foo` as I wanted to namespace the difference between the normal crate root scope and the static variables.

The `lazy_static!` macro only works in the crate root `lib.rs` or `main.rs` files. I wanted to keep the code here as minimal as possible, hence the reason for the `cypto` mod.

Hmm... `generic_keys` or `global_generic_keys` doesn't seem to be as clear to me; maybe: `ephemeral_instance_keys`, to show that these keys are not saved, and are generated on startup?

josecelano: I think the keys are both:

- Generic: because they are not intended to be used only for a single purpose.
- Ephemeral: because there are re-generated on the startup.

Is not the adjective "instance" supposed to describe that the life cycle is attached to the process (instance)?

I think the conflict is I want to explain what you should use them for, and you want to explain how long they last. They both are valid properties, but I think `ephemeral` is more precise than `generic`.

In summary, `ephemeral_instance_keys` is a good name for me.

src/lib.rs
da2ce7: Done, Renamed to 'Seed'.

src/lib.rs
da2ce7: Done, Renamed to `ephemeral_instance_keys`.

src/protocol/crypto.rs
da2ce7: Done, Renamed to `SeedKeeper` and `get_seed()`

da2ce7:
ACK d09c58f93aa6d3cc0d15bfcb20fd08b6f4abbd79

da2ce7:
ACK 873293a6f871ba6102d75df2379471b43bcf0096

PR #92, opened 2022-09-18
refactor: move clock/clock.rs into clock.rs (clippy)

da2ce7:
Fix the clippy warning that the module has the same name as it's directory. Will recreate the folder when needed by the interval counter, i.e. #83


da2ce7:
ACK e31d10563bdab5f5d30f1d01b41e563121091a02

da2ce7:
ACK 20f099eb0763a2d4625b788af63b2cd808440dd8

PR #96, opened 2022-10-04
Integration tests for UDP Tracker

### LINKED ISSUE #42: Torrust scrape has packed data errors and is not properly formatted
Hello, there seems to be an error with packed data when scraping announce both udp and http/s.

The format should be `d5:filesd20`, but on gbitt and torrust it is random, sometimes it is `d5:filesd50`, `d5:filesd54`, etc. Thats why almost all scrapers returns invalid data for info hash.

Seems like it is using aquatic source for scraping, but aquatic returns correct bytes of 20, where torrust doesn't.

@Power2All @WarmBeer

josecelano:
I'm trying to reproduce the error in issue #42. I've only created the scaffolding for integration tests so far.

If there is no bug I want to use this scaffolding for other integration tests. If we consider them helpful.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,27 +1,32 @@
-use log::info;
+use std::str::FromStr;
+use std::sync::Once;
+
+use log::{info, LevelFilter};
 
 use crate::Configuration;
 
-pub fn setup_logging(cfg: &Configuration) {
-    let log_level = match &cfg.log_level {
-        None => log::LevelFilter::Info,
-        Some(level) => match level.as_str() {
-            "off" => log::LevelFilter::Off,
-            "trace" => log::LevelFilter::Trace,
-            "debug" => log::LevelFilter::Debug,
-            "info" => log::LevelFilter::Info,
-            "warn" => log::LevelFilter::Warn,
-            "error" => log::LevelFilter::Error,
-            _ => {
-                panic!("Unknown log level encountered: '{}'", level.as_str());
-            }
-        },
-    };
+static INIT: Once = Once::new();
 
-    if log_level == log::LevelFilter::Off {
+pub fn setup_logging(cfg: &Configuration) {
+    let level = config_level_or_default(&cfg.log_level);
+
+    if level == log::LevelFilter::Off {
         return;
     }
 
+    INIT.call_once(|| {
+        stdout_config(level);
+    });
+}
+
+fn config_level_or_default(log_level: &Option<String>) -> LevelFilter {
+    match log_level {
+        None => log::LevelFilter::Info,
+        Some(level) => LevelFilter::from_str(level).unwrap(),
+    }
+}
+
+fn stdout_config(level: LevelFilter) {
     if let Err(_err) = fern::Dispatch::new()
         .format(|out, message, record| {
             out.finish(format_args!(
@@ -32,11 +37,12 @@ pub fn setup_logging(cfg: &Configuration) {
                 message
             ))
         })
-        .level(log_level)
+        .level(level)
         .chain(std::io::stdout())
         .apply()
     {
         panic!("Failed to initialize logging.")
     }
+
     info!("logging initialized.");
 }
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -18,6 +18,10 @@ pub fn setup_logging(cfg: &Configuration) {
         },
     };
 
+    if log_level == log::LevelFilter::Off {
+        return;
+    }
+
     if let Err(_err) = fern::Dispatch::new()
         .format(|out, message, record| {
             out.finish(format_args!(
--- a/tests/udp.rs
+++ b/tests/udp.rs
@@ -240,6 +240,19 @@ mod udp_tracker_server {
         assert!(is_connect_response(&response, TransactionId(123)));
     }
 
+    async fn send_connection_request(transaction_id: TransactionId, client: &UdpTrackerClient) -> ConnectionId {
+        let connect_request = ConnectRequest { transaction_id };
+
+        client.send(connect_request.into()).await;
+
+        let response = client.receive().await;
+
+        match response {
+            Response::Connect(connect_response) => connect_response.connection_id,
+            _ => panic!("error connecting to udp server {:?}", response),
+        }
+    }
+
     #[tokio::test]
     async fn should_return_an_announce_response_when_the_client_sends_an_announce_request() {
         let configuration = tracker_configuration();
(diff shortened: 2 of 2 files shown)


josecelano:
hi @da2ce7 @WarmBeer I have added three tests:

- [should_return_a_bad_request_response_when_the_client_sends_an_empty_request](https://github.com/torrust/torrust-tracker/pull/96/files#diff-4b41e272c2b5b49b68ebf4dff4c646aabc672609db1ef4248db9145960ac53d8R210-R223)
- [should_return_a_connect_response_when_the_client_sends_a_connection_request](https://github.com/torrust/torrust-tracker/pull/96/files#diff-4b41e272c2b5b49b68ebf4dff4c646aabc672609db1ef4248db9145960ac53d8R225-R240)
- [should_return_an_announce_response_when_the_client_sends_an_announce_request](https://github.com/torrust/torrust-tracker/pull/96/files#diff-4b41e272c2b5b49b68ebf4dff4c646aabc672609db1ef4248db9145960ac53d8R242-R293)

After adding the third one, I started having random failures for the third one. First, I send the "connect" request and then the "announce" request. For some reason, the client does not receive the "announce" response, but the server is sending it (I see it in the log).

If I comment out the first two tests, It always works, so tests must not be isolated.
I'm using a different client instance with a different local port for each test to be sure clients do not collide.

josecelano:
Sometimes I get this message:

```
2022-10-06T19:36:29.518810036+01:00 [torrust_tracker::logging][INFO] logging initialized.
2022-10-06T19:36:29.518884993+01:00 [torrust_tracker::jobs::udp_tracker][INFO] Starting UDP server on: 127.0.0.1:6969
test udp_tracker_server::should_return_a_bad_request_response_when_the_client_sends_an_empty_request ... ok
thread 'udp_tracker_server::should_return_an_announce_response_when_the_client_sends_an_announce_request' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }', tests/udp.rs:118:43
test udp_tracker_server::should_return_a_connect_response_when_the_client_sends_a_connection_request ... ok
test udp_tracker_server::should_return_an_announce_response_when_the_client_sends_an_announce_request ... FAILED

failures:

failures:
    udp_tracker_server::should_return_an_announce_response_when_the_client_sends_an_announce_request

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

And sometimes, the tests stop at the third test (there is no timeout for the "receive" data function).

image

If I stop the test execution with "CTRL+C" the socket is not close because there is still a thread (I suppose it is the udp server).

mickvandijke:
Hey @josecelano ,

When you run the tests synchronously Eg:

```rust
#[tokio::test]
async fn test() {
    should_return_a_bad_request_response_when_the_client_sends_an_empty_request().await;
    should_return_a_connect_response_when_the_client_sends_a_connection_request().await;
    should_return_an_announce_response_when_the_client_sends_an_announce_request().await;
}
```

They pass 100% of the time. 

I think when you run the tests at the same time, each of the tests will try and boot up a new udp tracker while all the tests are ongoing. This probably leads to packet loss.

So to fix it, we either have to run the tests synchronously or provide each test with its own UdpServer (on a unique port).

josecelano:
hi @WarmBeer @da2ce7, I've also added the test for the scrape request. I think it's ready for review. I only check basic behaviour. Only the happy path, and I only check the response type. Particular cases are (or should) be covered by unit tests. If you think that other tests could be helpful, let me know.

Maybe adding more tests to cover the full specification would be nice. That would be useful to run smoke tests against live UDP servers. But maybe it makes more sense to use a real client for that and make it an independent project.

src/logging.rs
```diff
@@ -18,6 +18,10 @@ pub fn setup_logging(cfg: &Configuration) {
         },
     };
 
+    if log_level == log::LevelFilter::Off {
```
da2ce7: @WarmBeer Why do you introduce this change here? It seems unrelated.

tests/udp.rs
```diff
@@ -0,0 +1,317 @@
...
+    use tokio::task::JoinHandle;
+    use torrust_tracker::jobs::udp_tracker;
+    use torrust_tracker::tracker::statistics::StatsTracker;
+    use torrust_tracker::tracker::tracker::TorrentTracker;
+    use torrust_tracker::udp::MAX_PACKET_SIZE;
+    use torrust_tracker::{logging, static_time, Configuration};
+
+    fn tracker_configuration() -> Arc<Configuration> {
+        let mut config = Configuration::default();
+        config.log_level = Some("off".to_owned()); // "off" is necessary when running multiple trackers
```
da2ce7: What happens, is it not possible for the logs to be tracker-local?

da2ce7 (approved):

src/logging.rs
josecelano: hi @da2ce7 We're using an independent UdpServer for each test, so we isolate tests.

```rust
    impl UdpServer {
       // ..

        pub async fn start(&mut self, configuration: Arc<Configuration>) {
            if !self.started.load(Ordering::Relaxed) {
                // ...

                // Initialize logging
                logging::setup_logging(&configuration);

                // ...
            }
        }
    }
```

The problem is each UdpServer instance tries to setup logging again with:

```rust
    if let Err(_err) = fern::Dispatch::new()
        .format(|out, message, record| {
            out.finish(format_args!(
                "{} [{}][{}] {}",
                chrono::Local::now().format("%+"),
                record.target(),
                record.level(),
                message
            ))
        })
        .level(log_level)
        .chain(std::io::stdout())
        .apply()
    {
        panic!("Failed to initialize logging.")
    }
```

I do not know how that code works, but I suppose is a kind of "pipe" to redirect the log to the stdout, and probably that can be done only once. If that is the case, we have these options:

1. What @WarmBeer did. We can disable log for testing. I do not like this solution because, very often, logging is helpful during testing. I use it a lot.
2. We can setup logging in a different way for testing, for example, using a different log file for each server. I do not like this option. It generates too much garbage.
3. We can use a different `setup_logging` for tests where we ensure we only call once the "pipe" generation code.
4. We can change `setup_logging` function in the production code to avoid calling that part twice, regardless the log level you want.

I would go for option 4.

image

tests/udp.rs
josecelano: Yes, I think it's possible. Read my [comment avobe](https://github.com/torrust/torrust-tracker/pull/96#discussion_r1001527325).

src/logging.rs
josecelano: hi @da2ce7 @WarmBeer, I've changed the `setup_logging` function to avoid panicking if it's called more than once. This way, we can enable logging even for testing (using more than one UdpServe). I still use the "off" level in testing because I prefer to keep the test output clean. You can enable it whenever you needed it.

da2ce7 (approved):
ACK 5dcea43da6fe6e7590ff69f4ac04200d2394acde

PR #99, opened 2022-10-17
format: re-apply rust-format

da2ce7:


da2ce7:
ACK 5e2bef3ce9c2c7924de2fcc93d65debf39c784de

josecelano:
hi @da2ce7 how are you formatting the code?

It seems the [format check](https://github.com/torrust/torrust-tracker/actions/runs/3214029904/jobs/5254120837#step:5:2) in the [test_build_release workflow](https://github.com/torrust/torrust-tracker/blob/develop/.github/workflows/test_build_release.yml#L20) does work.

I'm using VSCode, which I suppose runs something like:

```s
cargo clippy --no-deps
```

And you can fix with:

```
cargo clippy --no-deps --fix
```

The code in the PR is not well formatted because I've been using Clippy.

I assumed we were using Clippy because it shows more warnings (not only linting errors). But in practice, we were not forcing any format since the linting check is not working.

If we agree, we should change the workflow, although we have some warnings for Clippy. I would add a new issue for changing the formatter and fixing the warnings.

da2ce7:
@josecelano I think that the issue is because the difference in version of the rust-format command.  I've got the nightly version installed; that I think is more strict than the stable version.

Issue #68, opened 2022-08-14
Compilation issues for --target `armv7-unknown-linux-gnueabihf`


mickvandijke:
Torrust won't compile to armv7-unknown-linux-gnueabihf (Raspberry Pi etc..).

mickvandijke:
After some painful hours, I've figured out a way to cross compile to armv7-unknown-linux-gnueabihf from an Ubuntu host.

I mostly ran into errors with openssl and sqlite3, so I had to add and update some dependencies in the Cargo.toml to fix these. `openssl` had to be included with the `vendored` feature:
```toml
openssl = { version = "0.10.41", features = ["vendored"] }
```

And `rusqlite` had to include the feature `bundled` through the `r2d2_sqlite` crate:
```toml
r2d2_sqlite = { version = "0.21.0", features = ["bundled"] }
```

These are the commands I had to use for the compiler to succeed:
```
sudo apt-get install binutils-arm-linux-gnueabihf gcc-arm-linux-gnueabihf musl-tools

sudo apt-get install gcc-multilib

sudo apt-get install clang

sudo apt-get install sudo apt-get install libsqlite3-dev openssl libssl-dev pkg-config

sudo apt-get install -y libc6-armhf-cross libc6-dev-armhf-cross

sudo ln -s /usr/bin/arm-linux-gnueabihf-gcc /usr/bin/arm-linux-musleabihf-gcc
```

If you are coss compiling to armv7, you also have to add the following to `torrust-tracker/.cargo/config`:
```
[target.armv7-unknown-linux-gnueabihf]
linker = "arm-linux-gnueabihf-gcc"
```
If this file and folder don’t exist, you’ll have to create them yourself.

mickvandijke:
https://github.com/torrust/torrust-tracker/pull/69

josecelano:
hi @WarmBeer in the docker configuration I'm working on [here](https://github.com/torrust/torrust-tracker/pull/123), I'm using a docker image that probably does that internally. It works with `--target x86_64-unknown-linux-musl` but it does not work with `--target armv7-unknown-linux-gnueabihf`.

I think we could add a GitHub workflow to test cross-compilations like [this](https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml).

[...]

PR #100, opened 2022-10-18
Use nightly toolchain for formatting

josecelano:
@da2ce7 started using the nightly build for rust formatting in this [PR-99](https://github.com/torrust/torrust-tracker/pull/99)

And the job 'format' in the workflow [did not work](https://github.com/torrust/torrust-tracker/actions/runs/3273393423/jobs/5385541084#step:5:9), showing these warnings:

```
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```

So we needed the nightly channel anyway. I've changed the workflow to use it.

### DIFF
--- a/.github/workflows/test_build_release.yml
+++ b/.github/workflows/test_build_release.yml
@@ -13,11 +13,15 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
-          components: rustfmt
+          toolchain: nightly
+          override: true
+          components: rustfmt, clippy
       - uses: Swatinem/rust-cache@v1
       - name: Check Rust Formatting
-        run: cargo fmt --check
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --check
 
   test:
     needs: format


da2ce7:
@josecelano before enabling clippy in the CI, we should fix all the errors...

josecelano:
> @josecelano before enabling clippy in the CI, we should fix all the errors...

[Yes](https://github.com/torrust/torrust-tracker/pull/99#issuecomment-1282061661). I was only testing that the command works if we add it in the future.

josecelano:
ACK f2eaf9584cf8d396660f325594b835cbd70dedfa

da2ce7:
@josecelano why change to use the cargo action?

josecelano:
> @josecelano why change to use the cargo action?

@da2ce7 I think the action will create an "annotation" (GitHub UI) when something fails. The cargo command would only show the error in the logs. That was the only reason.

Anyway, you can easily enable [cross-compilation](https://github.com/actions-rs/cargo#cross-compilation). Although we do not need it in this case.

@WarmBeer was working on this cross-compilation [issue](https://github.com/torrust/torrust-tracker/issues/68). Maybe we could add a workflow to compile for different targets in the future.

da2ce7 (approved):
ACK f2eaf9584cf8d396660f325594b835cbd70dedfa

PR #103, opened 2022-10-21
Refactor `StatsTracker`

josecelano:
This is an experimental refactor of the [StatsTracker](https://github.com/torrust/torrust-tracker/blob/develop/src/tracker/statistics.rs).

It's still a WIP.

Tracker statistics are optional, and we had [problems testing that behaviour](https://github.com/torrust/torrust-tracker/pull/102).

I decided to try to make that code more testable. My idea is:

- Convert the StatsTracker into a StatsService that only creates the communication channel and the other services (EventSender, EventListener, ...)
- A EventSender to send the events.
- A EventListener that updates the statistics.
- A StatisticsRepository that contains the stats.

The high-level service (StatsService) only creates the StatisticsRepository used by the EventListener and the TorrusTracker. And the communication channel (with the sender and listener).

TODO:

- [X] Extract EventSender
- [X] Extract EventListener
- [x] Extract StatisticsRepository
- [x] Fix StatsEventSenderMock
- [x] Add tests for EnventListener

Once we have the EventListener, we can add tests to check if it updates the stats correctly, which is something we did not test on [PR-112](https://github.com/torrust/torrust-tracker/pull/102).

While working on the refactor, I realized the `StatsEventSenderMock` does not work correctly. It only works when the event is sent, but it does not fail if no event is sent. The obvious solution would be to store the event sent in the Mock, but the method is not mutable. 

```rust
    struct StatsEventSenderMock {
        expected_event: Option<TrackerStatisticsEvent>,
    }

    impl StatsEventSenderMock {
        fn new() -> Self {
            Self { expected_event: None }
        }

        fn should_throw_event(&mut self, expected_event: TrackerStatisticsEvent) {
            self.expected_event = Some(expected_event);
        }
    }

    #[async_trait]
    impl TrackerStatisticsEventSender for StatsEventSenderMock {
        async fn send_event(&self, event: TrackerStatisticsEvent) -> Option<Result<(), SendError<TrackerStatisticsEvent>>> {
            if self.expected_event.is_some() {
                assert_eq!(event, *self.expected_event.as_ref().unwrap());
            }
            None
        }
    }

    #[tokio::test]
    async fn it_should_send_the_upd4_connect_event_when_a_client_tries_to_connect_using_a_ip4_socket_address() {
        let tracker_stats_service = Box::new(TrackerStatsServiceMock::new());
        let mut stats_event_sender = Box::new(StatsEventSenderMock::new());

        let client_socket_address = sample_ipv4_socket_address();
        stats_event_sender.should_throw_event(TrackerStatisticsEvent::Udp4Connect);

        let torrent_tracker =
            Arc::new(TorrentTracker::new(default_tracker_config(), tracker_stats_service, Some(stats_event_sender)).unwrap());
        handle_connect(client_socket_address, &sample_connect_request(), torrent_tracker)
            .await
            .unwrap();
    }

```

I did not want to add a mocking library only for this case. That's why I did it manually. And I do not want to make that function mutable only for the test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,3 +59,6 @@ async-trait = "0.1"
 
 aquatic_udp_protocol = "0.2"
 uuid = { version = "1", features = ["v4"] }
+
+[dev-dependencies]
+mockall = "0.11.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,4 +61,4 @@ aquatic_udp_protocol = "0.2"
 uuid = { version = "1", features = ["v4"] }
 
 [dev-dependencies]
-mockall = "0.11.3"
+mockall = "0.11"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -480,6 +480,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "difflib"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6184e33543162437515c2e2b48714794e37845ec9851711914eec9d308f6ebe8"
+
+[[package]]
 name = "digest"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -509,6 +515,12 @@ name = "dlv-list"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0688c2a7f92e427f44895cd63841bff7b29f8d7a1648b9e7e07a4a365b2e1257"
+
+[[package]]
+name = "downcast"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1435fa1053d8b2fbbe9be7e97eca7f33d37b28409959813daefc1446a14247f1"
 
 [[package]]
 name = "either"
@@ -571,6 +583,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "float-cmp"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "98de4bbd547a563b716d8dfa9aad1cb19bfab00f4fa09a6a4ed21dbcf44ce9c4"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -599,6 +620,12 @@ checksum = "a9c384f161156f5260c24a097c56119f9be8c798586aecc13afbcbe7b7e26bf8"
 dependencies = [
  "percent-encoding",
 ]
+
+[[package]]
+name = "fragile"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6c2141d6d6c8512188a7891b4b01590a45f6dac67afb4f255c4124dbb86d4eaa"
 
 [[package]]
 name = "frunk"
@@ -1009,6 +1036,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b0fd2260e829bddf4cb6ea802289de2f86d6a7a690192fbe91b3f46e0f2c8473"
+dependencies = [
+ "either",
+]
+
+[[package]]
 name = "itoa"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


josecelano:
hi @da2ce7 @WarmBeer, I've fixed the tests that were using my incomplete custom mock. In [this commit](https://github.com/torrust/torrust-tracker/pull/103/commits/8874032074c12aecb1f250e054caac4d8c9c63f4).

I did not find a way to mock a trait function which is not mutable. This was my mock:

```rust
    struct StatsEventSenderMock {
        expected_event: Option<TrackerStatisticsEvent>,
    }

    impl StatsEventSenderMock {
        fn new() -> Self {
            Self { expected_event: None }
        }

        fn should_throw_event(&mut self, expected_event: TrackerStatisticsEvent) {
            self.expected_event = Some(expected_event);
        }
    }

    #[async_trait]
    impl TrackerStatisticsEventSender for StatsEventSenderMock {
        async fn send_event(&self, event: TrackerStatisticsEvent) -> Option<Result<(), SendError<TrackerStatisticsEvent>>> {
            if self.expected_event.is_some() {
                assert_eq!(event, *self.expected_event.as_ref().unwrap());
            }
            None
        }
    }
```

I needed to know if the method was called. If the method `send_event` is not called, the test does not fail. 

The only way to do it (or the only way I found) is by adding a new field to the struct to store the state. But I could not do that without changing the function signature to:

```rust
async fn send_event(&mut self, event: TrackerStatisticsEvent) -> Option<Result<(), SendError<TrackerStatisticsEvent>>>;
```

I tried with a [mocking library](https://docs.rs/mockall/latest/mockall/), and it works:

```rust
let mut stats_event_sender_mock = MockTrackerStatisticsEventSender::new();
            stats_event_sender_mock
                .expect_send_event()
                .with(eq(TrackerStatisticsEvent::Udp4Connect))
                .times(1)
                .returning(|_| Box::pin(future::ready(Some(Ok(())))));
            let stats_event_sender = Box::new(stats_event_sender_mock);
```

I think sooner or later, we will need it for more complex mocks.

I wonder how they can check that the method is called without changing its signature. I'm curious.

The other tricky thing was the returning value: `.returning(|_| Box::pin(future::ready(Some(Ok(())))))`

Since the function is an async function, you need to return a "future". But I do not know why they need the "pin".

We can apply this commit in the `develop` branch even if we do not merge this PR. Although I think all the changes I've made so far are useful.

josecelano:
hi @da2ce7 @WarmBeer, I've finished the experimental refactor. I think it is worth merging it for two reasons:

1. I fixed some tests with the `MockTrackerStatisticsEventSender` mock.
2. I've increased test coverage for the `statistics` mod.

There are a couple of things I do not like:

-  I think I can remove the trait `TrackerStatisticsEventSender`. Now I use `mockall` and I think I can mock the struct without adding the trait, which is only used in tests. I prefer to avoid adding traits if that's the only reason and we can mock the struct.
- I have isolated the asynchronous part that is hard to test. The infinite loop (`event_listener`) and launching the thread for the listener (`run_event_listener`). Those things are not tested, but I do not know how to do it without changing the production code (by introducing things that would be used only by tests).

josecelano:
> hi @da2ce7 @WarmBeer, I've finished the experimental refactor. I think it is worth merging it for two reasons:
> 
> 1. I fixed some tests with the `MockTrackerStatisticsEventSender` mock.
> 2. I've increased test coverage for the `statistics` mod.
> 
> There are a couple of things I do not like:
> 
> * I think I can remove the trait `TrackerStatisticsEventSender`. Now I use `mockall` and I think I can mock the struct without adding the trait, which is only used in tests. I prefer to avoid adding traits if that's the only reason and we can mock the struct.
> * I have isolated the asynchronous part that is hard to test. The infinite loop (`event_listener`) and launching the thread for the listener (`run_event_listener`). Those things are not tested, but I do not know how to do it without changing the production code (by introducing things that would be used only by tests).

I forgot to mention that this refactor makes it easier to decouple events from statistics in the future. Statistics would be only one of the consumers of those events. But it's not a reason to merge this PR since we have not decided how to handle events and their relationship with logging.

Cargo.toml
```diff
@@ -59,3 +59,6 @@ async-trait = "0.1"
 
 aquatic_udp_protocol = "0.2"
 uuid = { version = "1", features = ["v4"] }
+
+[dev-dependencies]
+mockall = "0.11.3"
```
da2ce7: Maybe we can just specify the major version, 0.11, here?

da2ce7 (approved):
Looks Good to me. Nice work! :)

Cargo.toml
josecelano: Done.

da2ce7:
ACK 6f77dfeee64037cd9011617057d9f1e5c3e2fd63

PR #105, opened 2022-10-24
Setting Overhaul

da2ce7:
* Move internal vocabulary from  'Configuration' with 'Settings'.
* Make use of a 'config' folder. Will generate new config files when needed.
* New Config is stored in json. And parsed without the need of the config crate.
* Automatically finds and moves old configuration to new format, (keeping a backup).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,8 @@
     "[rust]": {
         "editor.formatOnSave": true
     },
-    "rust-analyzer.checkOnSave.command": "clippy",
-    "rust-analyzer.checkOnSave.allTargets": true,
-    "rust-analyzer.checkOnSave.extraArgs": ["--","-W","clippy::pedantic"],
+    "rust-analyzer.checkOnSave": true,
+    "rust-analyzer.check.allTargets": true,
+    "rust-analyzer.check.command": "clippy",
+    "rust-analyzer.check.extraArgs": ["--","-W","clippy::pedantic"],
 }
--- a/src/databases/sqlite.rs
+++ b/src/databases/sqlite.rs
@@ -1,4 +1,5 @@
 use std::panic::Location;
+use std::path::Path;
 use std::str::FromStr;
 
 use async_trait::async_trait;
@@ -7,11 +8,35 @@ use r2d2_sqlite::SqliteConnectionManager;
 
 use super::driver::Driver;
 use crate::databases::{Database, Error};
+use crate::errors::settings::DatabaseSettingsError;
 use crate::protocol::clock::DurationSinceUnixEpoch;
 use crate::protocol::info_hash::InfoHash;
+use crate::settings::DatabaseSettings;
 use crate::tracker::auth;
 
 const DRIVER: Driver = Driver::Sqlite3;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct Sqlite3DatabaseSettings {
+    pub database_file_path: Box<Path>,
+}
+impl TryFrom<&DatabaseSettings> for Sqlite3DatabaseSettings {
+    type Error = DatabaseSettingsError;
+
+    fn try_from(value: &DatabaseSettings) -> Result<Self, Self::Error> {
+        match value.get_driver()? {
+            Driver::Sqlite3 => Ok(Sqlite3DatabaseSettings {
+                database_file_path: value.get_slq_lite_3_file_path()?,
+            }),
+            driver => Err(DatabaseSettingsError::WrongDriver {
+                field: "driver".to_string(),
+                expected: Driver::Sqlite3,
+                actual: driver,
+                data: value.to_owned(),
+            }),
+        }
+    }
+}
 
 pub struct Sqlite {
     pool: Pool<SqliteConnectionManager>,
--- a/src/databases/sqlite.rs
+++ b/src/databases/sqlite.rs
@@ -7,35 +7,17 @@ use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 
 use super::driver::Driver;
+use super::settings;
 use crate::databases::{Database, Error};
-use crate::errors::settings::DatabaseSettingsError;
 use crate::protocol::clock::DurationSinceUnixEpoch;
 use crate::protocol::info_hash::InfoHash;
-use crate::settings::DatabaseSettings;
 use crate::tracker::auth;
 
 const DRIVER: Driver = Driver::Sqlite3;
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub struct Sqlite3DatabaseSettings {
+pub struct Settings {
     pub database_file_path: Box<Path>,
-}
-impl TryFrom<&DatabaseSettings> for Sqlite3DatabaseSettings {
-    type Error = DatabaseSettingsError;
-
-    fn try_from(value: &DatabaseSettings) -> Result<Self, Self::Error> {
-        match value.get_driver()? {
-            Driver::Sqlite3 => Ok(Sqlite3DatabaseSettings {
-                database_file_path: value.get_slq_lite_3_file_path()?,
-            }),
-            driver => Err(DatabaseSettingsError::WrongDriver {
-                field: "driver".to_string(),
-                expected: Driver::Sqlite3,
-                actual: driver,
-                data: value.to_owned(),
-            }),
-        }
-    }
 }
 
 pub struct Sqlite {
@@ -47,8 +29,9 @@ impl Database for Sqlite {
     /// # Errors
     ///
     /// Will return `r2d2::Error` if `db_path` is not able to create `SqLite` database.
-    fn new(db_path: &str) -> Result<Sqlite, Error> {
-        let cm = SqliteConnectionManager::file(db_path);
+    fn new(settings: &settings::Settings) -> Result<Sqlite, Error> {
+        let sqlite_settings = settings.get_sqlite_settings()?;
+        let cm = SqliteConnectionManager::file(sqlite_settings.database_file_path);
         Pool::new(cm).map_or_else(|err| Err((err, Driver::Sqlite3).into()), |pool| Ok(Sqlite { pool }))
     }
 
--- a/src/errors/settings.rs
+++ b/src/errors/settings.rs
@@ -1,0 +1,243 @@
+use thiserror::Error;
+
+use super::FilePathError;
+use crate::databases;
+use crate::settings::{
+    CommonSettings, DatabaseSettings, GlobalSettings, ServiceNoSecrets, ServiceProtocol, TlsSettings, TrackerSettings,
+};
+
+#[derive(Error, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum SettingsError {
+    #[error("Bad Namespace: \".{field}\" {message}")]
+    NamespaceError { message: String, field: String },
+
+    // Todo: Expand this for Semantic Versioning 2.0.0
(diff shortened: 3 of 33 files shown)


src/logging.rs
```diff
@@ -3,30 +3,30 @@ use std::sync::Once;
...
+pub fn setup_logging(settings: &Settings) {
+    let level = settings_level_or_default(&settings.log_level);
 
     if level == log::LevelFilter::Off {
         return;
     }
 
     INIT.call_once(|| {
-        stdout_config(level);
+        stdout_settings(level);
```
josecelano: hi @da2ce7, in this case, "config" is a verb. The function `stdout_config` sets up the logging. I was tempted to rename it to `redirect_logging_to_stdout` or something like that.

josecelano (approved):

josecelano:
hi @da2ce7, it looks good to me. In the future, something that could be useful is to allow overwriting default settings from env vars instead of from a `config/local.toml` file. 

I've seen that the crate we are using allows it.

I think the `local.toml` file is useful for development env and installations that are done manually. If you are using docker, it could be nice to overwrite the settings using env vars. Anyway, you can also mount the configuration as I did here:

https://github.com/josecelano/torrust-tracker/blob/issue-11-docker-support/bin/docker/run.sh#L6

And with K8S, you also have [ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/). These are the ways that the container can get the configuration:

- Inside a container command and args.
- Environment variables for a container.
- Add a file in read-only volume, for the application to read.
- Write code to run inside the Pod that uses the Kubernetes API to read a ConfigMap.

There is also another thing to consider. Some settings are secrets. From the application point of view, I suppose it does not matter. If we are using docker, we can decide that the whole configuration is passed as a mounted read-only file or as env vars. I do not know if one could be considered safer than the other.

josecelano:
ACK 698f1819f2661871462182cf9f20cca17f1e5e9a

da2ce7:
Needs Rebase.

.vscode/settings.json
```diff
@@ -1,14 +1,51 @@
 {
     "cSpell.words": [
```
josecelano: hi @da2ce7 I would move this list to a `.cspell.json` file in the root folder. I think the VSCode detects it anyway and we can keep the list with any IDE.

Besides, if we use Megalinter in the future it can parse the independent file.

src/databases/sqlite.rs
```diff
@@ -1,23 +1,50 @@
...
 }
 
 impl SqliteDatabase {
-    pub fn new(db_path: &str) -> Result<SqliteDatabase, r2d2::Error> {
-        let cm = SqliteConnectionManager::file(db_path);
+    pub fn new(settings: &Sqlite3DatabaseSettings) -> Result<SqliteDatabase, r2d2::Error> {
+        let cm = SqliteConnectionManager::file(settings.database_file_path.as_path());
         let pool = Pool::new(cm).expect("Failed to create r2d2 SQLite connection pool.");
         Ok(SqliteDatabase { pool })
     }
```
josecelano: hi @da2ce7, I always have doubts about this kind of change. I have a "rule" which is: "do not pass to a function more than what it needs". In this case, the constructor only needs the path for the DB file. In the first version, SqliteDatabase did not know anything about the `Settings`, and now they are coupled. If you change the settings you also have to change this function.

Normally you would do that because you think in the future, the `SqliteDatabase` might need other values.

Anyway, I think it's a nice generalization since we have more than one database driver. I would expect all drivers to have a similar constructor regardless of their specific parameters.

On the other hand, would it make sense to add a wrapper for the DB path like this:

```rust
pub struct Sqlite3DatabaseSettings {
    pub database_file_path: Sqlite3DatabaseFilePath,
}

pub struct Sqlite3DatabaseFilePath {
    pub file_path: PathBuf,
}
```

In general, I have another rule: "do not add a wrapper to a primitive type (or other types) if you do not have any rule (invariant) to enforce".

In this case, we do not have it unless we want to check that the file exists and contains SQLite data, which is, I think, a bad idea to do in the constructor. But maybe we can check that the path is for a file and not a directory.

But sometimes, I think these custom types are a good way to easily find where you are using them.

Anyway, I would probably do this:

```rust
pub fn new(db_file_path: &PathBuf) -> Result<SqliteDatabase, r2d2::Error> {
    let cm = SqliteConnectionManager::file(db_file_path.as_path());
    let pool = Pool::new(cm).expect("Failed to create r2d2 SQLite connection pool.");
        Ok(SqliteDatabase { pool })
    }
```

By the way, maybe we should use `Path` instead of `PathBuf` here. But I do not know very weel the difference. I think here we could use a `Path` because we do not need to change it.

src/errors/settings.rs
```diff
@@ -0,0 +1,243 @@
...
+                message: _,
+                data: _,
+            } => field,
+        }
+        .to_owned()
+    }
+}
+
+#[derive(Error, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum CommonSettingsError {
```
josecelano, line 92: hi @da2ce7, what about `GenericSettingsErrors`?

src/jobs/torrent_cleanup.rs
```diff
@@ -4,15 +4,13 @@ use chrono::Utc;
 use log::info;
 use tokio::task::JoinHandle;
 
-use crate::tracker::tracker::TorrentTracker;
-use crate::Configuration;
+use crate::tracker::core::TorrentTracker;
 
-pub fn start_job(config: &Configuration, tracker: Arc<TorrentTracker>) -> JoinHandle<()> {
+pub fn start_job(interval: u64, tracker: Arc<TorrentTracker>) -> JoinHandle<()> {
```
josecelano: hi @da2ce7, this is an example where you decided to pass only what the function needs. I would like to know why you did it here.

src/jobs/tracker_api.rs
```diff
@@ -3,19 +3,15 @@ use std::sync::Arc;
...
+use crate::tracker::core::TorrentTracker;
 
-pub fn start_job(config: &Configuration, tracker: Arc<TorrentTracker>) -> JoinHandle<()> {
-    let bind_addr = config
-        .http_api
-        .bind_address
-        .parse::<std::net::SocketAddr>()
-        .expect("Tracker API bind_address invalid.");
-    info!("Starting Torrust API server on: {}", bind_addr);
+pub fn start_job(settings: &ApiServiceSettings, tracker: Arc<TorrentTracker>) -> JoinHandle<()> {
```
josecelano: hi @da2ce7, Another example where you reduce the data you pass to the function.

src/lib.rs
```diff
@@ -1,23 +1,27 @@
...
+pub mod config_const {
+    pub const CONFIG_FOLDER: &str = "config";
+    pub const CONFIG_BACKUP_FOLDER: &str = "config.backup";
+    pub const CONFIG_ERROR_FOLDER: &str = "config.error";
+    pub const CONFIG_DEFAULT: &str = "default";
+    pub const CONFIG_LOCAL: &str = "local";
+    pub const CONFIG_OVERRIDE: &str = "override";
+    pub const CONFIG_OLD: &str = "../config";
+}
+
```
josecelano, line 26: hi @da2ce7 why do not you put this in the `config mod`?.

src/settings/manager.rs
```diff
@@ -0,0 +1,552 @@
...
+
+        assert!(!temp.exists());
+
+        SettingsManager::make_folder(&temp).unwrap();
+
+        assert!(temp.is_dir());
+    }
+
+    #[test]
+    fn it_should_write_and_read_and_write_default_config() {
```
josecelano: @da2ce7 Do you mean "it should migrate config file from toml to json"? 

And there are no assertions.

src/settings/mod.rs
```diff
@@ -0,0 +1,1150 @@
...
+
+        if let Some(val) = old_settings.on_reverse_proxy {
+            self.global_settings.on_reverse_proxy = Some(val);
+        }
+        self
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Debug, Clone, Hash)]
+pub struct CommonSettings {
```
josecelano, line 615: @da2ce7 what's the difference between "global" and "common"?

src/settings/mod.rs
```diff
@@ -0,0 +1,1150 @@
...
+            .map_err(|source| TlsSettingsError::BadKeyFilePath {
+                field: "key_file_path".to_string(),
+                source,
+            })
+    }
+}
+
+#[derive(Serialize, Deserialize, PartialEq, PartialOrd, Ord, Eq, Debug, Copy, Clone, Hash, Display)]
+#[serde(rename_all = "snake_case")]
+pub enum LogFilterLevel {
```
josecelano, line 1129: @da2ce7 move to logging?

src/tracker/mod.rs
```diff
@@ -1,6 +1,78 @@
...
+    use std::net::IpAddr;
+    use std::sync::Arc;
+
+    use super::mode::TrackerMode;
+    use crate::databases::database::{self, Database};
+    use crate::settings::{
+        CommonSettings, CommonSettingsBuilder, DatabaseSettingsBuilder, GlobalSettings, GlobalSettingsBuilder, LogFilterLevel,
+    };
+
+    pub struct TrackerArgs {
```
josecelano: @da2ce7 I think this struct is useful now, but in the long term, I think we should extract some responsibilities from the Tracker, for example persistence. I think we could have:

- One handler or controller fer each request type (announce, scrape, ...)
- An app service (announce is the same from HTTP and udp). It could be also a command or query with its handler. For example: AnnounceCommand with AnnounceCommandHandler.
- A TorrentService to add/update/remove torrents. We can trigger events at this level.
- The TorrentService can use a TorrentRepository (DB: SQLite or MySQL) and a TorrentCache (the current in-memory structure).
But it's just a vague idea I have. I wanted to draft something, but I have not had time. And I also wanted to get to know the repo better before spending time proposing this kind of change.

tests/udp.rs
```diff
@@ -149,8 +171,8 @@ mod udp_tracker_server {
     }
 
     /// Creates a new UdpTrackerClient connected to a Udp Tracker server
-    async fn new_connected_udp_tracker_client(remote_address: &str) -> UdpTrackerClient {
-        let udp_client = new_connected_udp_client(remote_address).await;
+    async fn new_connected_udp_tracker_client(remote_socket: &SocketAddr) -> UdpTrackerClient {
```
josecelano: @da2ce7 I like this change. We should try to normalize this variable name. We are using:

- remote_address
- remote_socket
- remote_ip
- remote_client
- client?

I think `client_socket` is better because it's more specific. But I have not changed to avoid introducing more versions. 

And I've just realized in this case, it's the `server_socket` which makes my assumptions more valid that we should be more specific.

josecelano (approved):

josecelano:
hi @da2ce7, it looks good to me! I've just added some minor comments. I agree on all refactors, but I'm not sure why you want to change the config file from `toml` to `json`. I do not see any reason to do it but do not see any reason not to do it, either.

josecelano (approved):
It looks to me, although I still do not know the reason for using JSON instead of TOML. is it only because you want to remove the dependency?

da2ce7:
Considering to Rewrite this entire thing all-again.  This time making full use of https://github.com/SergioBenitez/Figment

da2ce7:
Partial Rebase. Completed Database Settings.

da2ce7:
Closing, needs to be re-imagined, suggest using https://github.com/SergioBenitez/Figment as config framework library.

PR #115, opened 2022-11-23
Fix Clippy Pedantic Warnings and Structure Rename

da2ce7:
Massive Rename of Structs and Fixing of Many Pedantic Clippy Warnings.


src/tracker/torrent.rs
```diff
@@ -93,7 +93,7 @@ impl Default for Entry {
 }
 
 #[derive(Debug)]
-pub struct Stats {
+pub struct Swarm {
```
josecelano: hi @da2ce7, just to confirm  you have used the term "Swarm" because in BitTorrent that means all the nodes that have the torrent.

https://en.wikipedia.org/wiki/Glossary_of_BitTorrent_terms#Swarm

josecelano (approved):

src/tracker/torrent.rs
da2ce7: If this was a torrent client, I believe that should be something like "connections", as this is the index, the "swam" is all the peers that have this torrent. - So I believe it is the correct word.

da2ce7: I just call it `Swam` instead of `SwamSize` or something, as we may add more info related to the Swam to this struct in the future.

josecelano: OK, I thought that was the reason. Maybe we can call it `SwamStats` now until we have more fields and then rename it. What do you think? I always try to reveal current intentions, not future behaviour, with names.

da2ce7:
ACK 728fe8a17e090b56dbde3284c22f3956b641edbd

da2ce7:
ACK 0f075e4daee476db840a4dbf98e3639a84ecd1bc

PR #117, opened 2022-11-24
Fix API bug: Torrent cannot be whitelisted twice

### LINKED ISSUE #74: Tracker API should return an OK status when whitelisting an already whitelisted info hash

josecelano:
- [x] Reproduce the error with an e2e test.
- [x] Fix it.
- [x] Rename test to `should_allow_to_whitelist_a_torrent`
- [x] Refactor: extract function `is_info_hash_whitelisted` as described [here](https://github.com/torrust/torrust-tracker/pull/118#issuecomment-1326847691).

### DIFF
--- a/src/jobs/tracker_api.rs
+++ b/src/jobs/tracker_api.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use log::info;
-use tokio::sync::oneshot::{self, Receiver};
+use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 
 use crate::api::server;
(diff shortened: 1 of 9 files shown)


josecelano:
ACK 2d621c5e64f657e9be1f99af65ba8b160b9b68d0

da2ce7 (changes requested):
You seem to be just checking if there is any error, then assuming that it isn't already inserted.

I suggest that you consider: https://github.com/torrust/torrust-tracker/pull/118

josecelano:
hi @da2ce7 @WarmBeer, I changed the implementation to send a message from the API server to the initiator when the API is ready, as we discussed yesterday. You can see the changes [here](https://github.com/torrust/torrust-tracker/pull/117/commits/32a6d79ea3c53401a14a44dede63e039330675e1).

I do not like it yet. I think we should send the message after starting the server.

I think we can do what @da2ce7 suggested. Something like:

```rust
tokio::spawn(async move {
    let (_addr, api_server) = serve(server).bind_with_graceful_shutdown(socket_addr, async move {
        tokio::signal::ctrl_c().await.expect("Failed to listen to shutdown signal.");
    });
});

// Send a message to the initiator to notify the API is ready to accept requests
if messenger_to_initiator.send(ApiReady()).is_err() {
    panic!("the receiver dropped");
}
```

@da2ce7 is that what you suggested? Using a new thread to launch the server and send the message afterwards.

I'm not sure if that's a significant change. I suppose the problem was not the time it takes the server to be ready but the fact that tests run in parallel.

josecelano:
ACK 15aa8313fde6aeb82daa563bc9bc4ca902402808

src/jobs/tracker_api.rs
```diff
@@ -1,30 +1,40 @@
...
 
+    let (tx, rx) = oneshot::channel::<ApiServerJobStarted>();
+
+    // Run the API server
     let join_handle = tokio::spawn(async move {
-        server::start(bind_addr, tracker, tx).await;
+        if tx.send(ApiServerJobStarted()).is_err() {
+            panic!("the start job dropped");
+        }
+        server::start(bind_addr, tracker).await;
```
da2ce7: What reason to send `ApiServerJobStarted()` before `server::start` has finished starting?

Maybe something like this is better:

```Rust
    let join_handle = tokio::spawn(async move {
        let handel = server::start(bind_addr, tracker);

        if tx.send(ApiServerJobStarted()).is_err() {
            panic!("the start job dropped");
        }

        handel.await
    });
```
?

josecelano:
ACK 5274b2c067aeea9d1bf344930fe7f1bdea794627

da2ce7 (approved):
ACK 5274b2c067aeea9d1bf344930fe7f1bdea794627

Issue #122, opened 2022-12-02
[proposal] Create an installer with a flag (argument)


josecelano:
When you run the tracker the first time, you see this output:

```
No config file found.
Creating config file..
thread 'main' panicked at 'Please edit the config.TOML in the root folder and restart the tracker.', src/main.rs:22:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Running the tracker without the setting file has a side effect. It creates a setting file with the default configuration.

```s
$ cat config.toml
log_level = "info"
mode = "public"
db_driver = "Sqlite3"
db_path = "data.db"
announce_interval = 120
min_announce_interval = 120
max_peer_timeout = 900
on_reverse_proxy = false
external_ip = "0.0.0.0"
tracker_usage_statistics = true
persistent_torrent_completed_stat = false
inactive_peer_cleanup_interval = 600
remove_peerless_torrents = true

[[udp_trackers]]
enabled = false
bind_address = "0.0.0.0:6969"

[[http_trackers]]
enabled = false
bind_address = "0.0.0.0:6969"
ssl_enabled = false
ssl_cert_path = ""
ssl_key_path = ""

[http_api]
enabled = true
bind_address = "127.0.0.1:1212"

[http_api.access_tokens]
admin = "MyAccessToken"
```

I suggest showing a message like this:

```
No config file found: "config.toml".
You can create the default settings file running: cargo run --bin install
```

That "install" binary could be extended in the feature to be an installation assistant if you run it in interactive mode.

But the main reason for this change is I'm working on setting up docker configuration to automate deployments for the Turrust tracker and running a process that returns an error that could lead to errors while building docker images or others tasks.

On the other hand, I always prefer to be explicit about side effects. Besides, showing an error the first time you run a tracker could make people think the application is not good.

If you do not like that approach, the second option could be to ask the user if they want to create that file, but I do not like that option because it forces the app to be always interactive. We should have added an option to avoid that question in this case and return an error message with an error return value (not a panic message).

da2ce7:
I would like to have tracker-configuration file to be specified by some schema, and configured over a management API.

The  exception should be the core configuration options of the tracker. Settings that guide the access to the trackers management API (such as the management socket, or the public key of the administrator). These should be passed-in either by launch arguments, or picked up from environmental variables.

Once the tracker has been launched, then all configuration should happen over it's management API.

For example, The torrust index client could then be extended with an interface for management of the Tracker.

[...]

PR #123, opened 2022-12-06
Full docker support

### LINKED ISSUE #11: Docker Support

josecelano:
This is a re-work of [PR-55](https://github.com/torrust/torrust-tracker/pull/55) adding more features.

UPDATE ON: 2022/12/16

### TASKS

- [x] Docker image
- [X] Publish the docker image manually (https://hub.docker.com/repository/docker/josecelano/torrust-tracker)
- [x] Cache for dependencies in Dockerfile is not working. It compiles all the dependencies every time you build the image.
- [x] Workflow to publish new docker images when new semver tags are created.
- [x] Fix broken test after moving database to `storage/database` dir.
- [x] Deploy sample app using SQLite with [docker on Azure](https://docs.docker.com/cloud/aci-integration/).
- [x] Docker compose configuration using MySQL instead of SQLite for development
- [x] Docker compose configuration using MySQL instead of SQLite for production. Moved to another [repo](https://github.com/josecelano/awesome-torrust-tracker-compose).
- [x] Deploy sample app using MySQL with [docker compose on Azure](https://docs.docker.com/cloud/aci-integration/). Moved to another [repo](https://github.com/josecelano/awesome-torrust-tracker-compose).

### SUBTASKS (deploy with docker)

This option is a deployment to ACI using only one container with SQLite and no HTTPS.

- [x] Create a certificate with Let's Encrypt and use Rust (no Nginx). DISCARDED (I'll do it in the docker-compose version)
- [x] Enable HTTPS for the API.
- [x] Error duplciate port 6969.

### SUBTASKS (deploy with docker-compose)

- [x] Create a certificate with Let's Encrypt and use Nginx. DISCARDED for this PR.
- [x] Auto-renew certificate with `certbot`. DISCARDED for this PR.
- [x] Use MySQL instead of SQLite

### REQUIREMENTS

- Multistage builds with slim/alpine images for production
- Do not execute the container with the `root` user. 
- Try to cache cargo dependencies with [cargo-chef](https://github.com/LukeMathWalker/cargo-chef)
- Build static rust binaries for x86_64 linux environments (without other system library dependencies)
- Consider using [certificates for localhost](https://letsencrypt.org/docs/certificates-for-localhost/) too?
- Use Nginx with certbot on Production for HTTPS.
- Auto-renew certificate with cerbot or Nginx docker image with auto-renew.
- Use Github Actions cache for docker cache: https://docs.docker.com/build/building/cache/backends/gha/
- User semver for docker image tags.

### OPTIONAL

- [x] `.devcontainer` configuration for GitHub Codespaces. DISCARDED for this PR.
- [x] docker compose configuration for [docker env environments](https://docs.docker.com/desktop/dev-environments/set-up/). DISCARDED for this PR.

### LINKS

- https://github.com/docker/awesome-compose
- https://hub.docker.com/repository/docker/josecelano/torrust-tracker
- https://docs.docker.com/cloud/aci-integration/
- https://docs.docker.com/desktop/dev-environments/set-up/
- https://github.com/clux/muslrust
- https://github.com/LukeMathWalker/cargo-chef
- https://letsencrypt.org/docs/certificates-for-localhost/
- https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf
- https://github.com/docker/awesome-compose/tree/master/react-rust-postgres
- https://kerkour.com/rust-small-docker-image


josecelano:
I can't manage to use the build cache for the dependencies. I've created an [issue](https://github.com/LukeMathWalker/cargo-chef/issues/172) in the [cargo-chef repo](https://github.com/LukeMathWalker/cargo-chef) to see if someone can help me to find the problem.

josecelano:
UPDATE:

- The [docker image](https://hub.docker.com/repository/docker/josecelano/torrust-tracker) is being [published automatically](https://hub.docker.com/repository/docker/josecelano/torrust-tracker) following semver conventions.
- I had to move the persistent state to a folder (`./storage/config/config.toml` and `./storage/database/data.db`) because Azure Container Instances do not allow to mount a single file. Anyway, it makes things easier in general. We do not need to pre-generate those files before running the container.
- Now I'm trying to fix a permissions problem. "Azure file share volume mount" [requires the Linux container run as root](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-volume-azure-files#limitations). I do not know why. But I had to change the public docker images to run as "root" because the docker integration with ACI does not allow overwriting the USER.
- One test is broken in the GitHub runner because it can't create a new folder for the database (`./storage/database/`). It's working on my local machine because I have it. I will create a new issue to review all test using the default configuration. We should **NOT** use the default location for the database.

josecelano:
UPDATE:

The permission problem was fixed using the `root` user to run the container. I'm planning to build docker images only for Azure Container Instances with a tag prefix like `v1.0.0-aci`.

I managed to deploy the container but the UDP port is closed. The API TCP port is working fine.

I'm running the container with:

```s
docker run \
    --publish 80:80/udp \
    --publish 1212:1212 \
    --volume torrustracker/test-volume:/app/storage \
    registry.hub.docker.com/josecelano/torrust-tracker:0.4.0
```

The volume has the proper configuration file (./storage/config/config.toml).

I've opened an issue on the docker forum: https://forums.docker.com/t/unable-to-open-a-udp-port-on-port-80/132661. You could not expose more than one port some years ago, but now you can also do it with the Azure dashboard assistant. I do not see any error running the container so I guess it's a firewall/forwarding issue.

josecelano:
UPDATE:

### Problem 1: closed port

I've spent the whole day trying to find out why the UDP tracker does not work.

I was using [this service](https://www.yougetsignal.com/tools/open-ports/) to check if the port is open. The service says the port is closed, but in fact, I can connect using [netcat](https://manpages.debian.org/stretch/netcat-traditional/nc.1.en.html):

```s
$ nc -u 20.246.181.9 6969
ss
bad request
```

In the end, there was no problem exposing the UDP port on the Azure container.

### Problem 2: qBittorrent can't connect

After confirming that the port was responding, I thought the client could be the problem.

I'm using BitTorrent v4.4.1.

After trying different combinations on the server and client sides, this one worked:

Tracker URL on the client: `udp://20.246.181.9:6969`

Tracker config on the server:

```toml
log_level = "debug"
mode = "public"
db_driver = "Sqlite3"
db_path = "./storage/database/data.db"
announce_interval = 120
min_announce_interval = 120
max_peer_timeout = 900
on_reverse_proxy = false
external_ip = "0.0.0.0"
persistent_torrent_completed_stat = true
inactive_peer_cleanup_interval = 600
remove_peerless_torrents = false

[[udp_trackers]]
enabled = true
bind_address = "0.0.0.0:6969"

[[http_trackers]]
enabled = true
bind_address = "0.0.0.0:6969"
ssl_enabled = false
ssl_cert_path = ""
ssl_key_path = ""

[http_api]
enabled = true
bind_address = "0.0.0.0:1212"

[http_api.access_tokens]
admin = "MyAccessToken"
```

I do not know why it's working now because I tried that configuration since the beginning. I remember that @Power2All reported a [problem with qBitTorrent](https://github.com/torrust/torrust-tracker/issues/42#issuecomment-1284478448) before.

### Problem 3: UDP debug messages do not appear in logs

There is a problem that made me think the UDP tracker was not receiving the request. With debug enabled It does not show these two messages:

```s
debug!("{:?}", payload);
```

It happens when I run the tracker with docker on both my local machine and the Azure container. But It does not happen when you run the app directly on your machine with `cargo run`.

I can see messages from the parent thread (main docker process) and the first child:

```s
$ docker logs heuristic-stonebraker
2022-12-09T14:34:15.173329600+00:00 [torrust_tracker::logging][INFO] logging initialized.
2022-12-09T14:34:15.193516700+00:00 [torrust_tracker::jobs::udp_tracker][INFO] Starting UDP 197789626+00:00 [torrust_tracker::jobs::torrent_cleanup][INFO] Cleaning up torrents..
```

But the primary process uses `tokio::spawn` to launch a child thread that creates the UDP server. And then, it calls `udp_start()` function which launches two threads with `tokio::select!`; one of them, the UDP packet handler, is the one that prints the message.

### Problem 4: every time I restart the Azure container, the IP changes

I have to find a way to assign a static IP for the time being. Maybe [this](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-egress-ip-address) could be the solution.

Later, when I start using Nginx and docker-compose I want to use a domain instead.

josecelano:
UPDATE:

Regarding problem 3, I've found the problem. In the `Config.toml` file we have this section:

```
[profile.release]
debug = 1
opt-level = 3
lto = "fat"
strip = true
```

The `debug!` macro is stripped in the release build. So even if you enable "debug" mode in production is not going to show the debug messages.

josecelano:
UPDATE:

- Tests were fixed with https://github.com/torrust/torrust-tracker/pull/126
- I have [removed](https://github.com/torrust/torrust-tracker/pull/123/commits/8333929d0e2876425c822b99a4e60be3269b07b2) the `strip = true` option from the `Cargo.toml` file. That's why the [docker cache layer for cargo dependencies did not work](https://github.com/LukeMathWalker/cargo-chef/issues/149). @WarmBeer I suppose that is not a problem. The binary is 83M without the strip and 9,8M with the strip. We can revert this change when it's fixed in cargo-chef. I suppose the binary size is not a problem right now, is not it?

mickvandijke:
> UPDATE:
> 
> * Tests were fixed with [Use an independent SQLite DB for each test #126](https://github.com/torrust/torrust-tracker/pull/126)
> * I have [removed](https://github.com/torrust/torrust-tracker/pull/123/commits/8333929d0e2876425c822b99a4e60be3269b07b2) the `strip = true` option from the `Cargo.toml` file. That's why the [docker cache layer for cargo dependencies did not work](https://github.com/LukeMathWalker/cargo-chef/issues/149). @WarmBeer I suppose that is not a problem. The binary is 83M without the strip and 9,8M with the strip. We can revert this change when it's fixed in cargo-chef. I suppose the binary size is not a problem right now, is not it?

I don't think the binary size is a problem at this stage. So you can disable the strip flag if that works better.

josecelano:
hi @da2ce7 @WarmBeer, I've enabled the HTTPS for the API too. You can see it in this [commit](https://github.com/torrust/torrust-tracker/pull/123/commits/46971a208eeffe31318f4dd517c6742b3494f174).

I think it's helpful if you want to deploy the app using only the application and a single docker container, but you also want to use HTTPS for HTTP trackers and the API.

josecelano:
hi @da2ce7 @WarmBeer, It seems I have to create [a lot of resources on Azure](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-application-gateway) if I want to use a static IP for the tracker.

The tracker is running on an Azure Container Instance. You can run one instance with only docker and a few steps I've added to the README. It's straightforward to do, but it's also constrained:

### For installation

- You have to do a lot of extra steps if you want a static IP. It makes sense to use Kubernetes to [Azure Container Apps](https://learn.microsoft.com/en-us/azure/container-apps/overview).

### For upgrades

- You need to stop the docker container and start a new one. If you have the container group, the virtual private network and the application gateway keeping the IP is not a problem, but I think you have to replace the docker images with your script. Maybe it also makes sense to use Kubernetes or Azure Container Apps.

### Conclusion

The goal for this PR was:

- Set up the basic docker configuration to enable the deployment using docker.
- Set up a demo production environment to be sure the tracker is easily deployable to any infrastructure, especially using docker.

We could limit the scope of this PR to only a simple demo installation (including HTTPs certificates and MySQL) but without proposing a complete production solution that includes:

- 100% automatic setup with one single command (with K8S, Terraform, etcetera ...)
- 100% automatic upgrades: upgrade docker images when a new docker image is available or running migration scripts, ...
- Scripts to handle backups (storage folder).
- Inject configuration into the container automatically.
- Rules for autoscaling.
- ...

Maybe I can finish this PR and other similar PRs for the index (backend and frontend) with the basic docker setup. If we want to run a complete 100% autonomous tracker and index as a demo, we could do it later using K8S or a virtual machine using terraform, docker-compose and some custom scripts.

josecelano:
I'm trying to publish the port 6969 for both TCP and udp and I get this error:

```s
$ docker run \
    --publish 6969:6969/udp \
    --publish 6969:6969/tcp \
    --publish 1212:1212/tcp \
    --volume torrustracker/test-volume:/app/storage \
    registry.hub.docker.com/josecelano/torrust-tracker:0.7.0
[+] Running 0/1
 ⠿ Group sweet-gates  Error                                                                                                                  4.2s
containerinstance.ContainerGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="DuplicateContainerPorts" Message="Duplicate ports '6969' found in container group 'sweet-gates' container 'sweet-gates'."
```

It's working with docker on my machine, so It must be an Azure constrain.

josecelano:
UPDATE:

Added docker-compose configuration for development. You can run it with `docker compose up`.

josecelano:
UDPATE:

hi @WarmBeer @da2ce7, I've been able to deploy the Tracker to Azure Container Instances with docker compose, but only with one service (the tracker). When I try to add other services like Nginx, Certbot, and Mysql, the instance gets stuck in "creating" or "updating" state. I've created a new repo with all the configurations I've been testing: https://github.com/josecelano/awesome-torrust-tracker-compose.

I think the reason is Azure Container Instances are very small, you only have 4 CPUs for all the containers. They even recommend using [Azure Container Apps](https://azure.microsoft.com/en-us/products/container-apps/#overview) when you have more complex services. I prefer K8S than Azure Container Apps because you could use it with the other big cloud providers.

I'm going to add to this PR the docker-compose configuration to run all those services locally. It works locally when I use Nginx and MySQL. I think we can keep a sample config in this repo with Nginx (reverse proxy) and MySQL, but without certbot.

I can add another example to the [new repo](https://github.com/josecelano/awesome-torrust-tracker-compose) for deploying the app with docker-compose to another provider. For example, [Digial Ocean](https://api-platform.com/docs/deployment/docker-compose/) with a droplet or K8S.

josecelano:
It seems you can create remote contexts with your docker nodes (virtual machines with docker):

https://danielwachtel.com/devops/deploying-multiple-dockerized-apps-digitalocean-docker-compose-contexts

```s
docker context create remote --docker "host=ssh://deployer@yourdomain.com"
docker context ls

# output:
default * .... unix://var/run/docker.sock     .... swarm
remote    .... ssh://deployer@yourdomain.com
```

I think this is better than ACI (for our requirements) because

- It's generic. You can install it on any VM and provider.
- You can change the instance size. 
- You can deploy more than one app. For example, different configurations.

The example sets up the Nginx and Certbot on the docker host (VM) instead of using docker-compose. I think I would keep that configuration in the docker-compose, even if we need an extra Nxing instance on the host as load balancer (or gateway).

My idea was even more straightforward: just to run "docker-compose" manually on the VM. The advantage of the previous solution is you can deploy or update remotely, for example, from a CI workflows.

josecelano:
hi @WarmBeer @da2ce7 

I tried to deploy the tracker to [Digital Ocean App Platfom](https://www.digitalocean.com/products/app-platform), but it's not possible for two reasons:

1. You can only inject env variables into the container. You can't mount a volume, so the only way to make it work is to build your docker image with the `config.toml` file inside.
2. When the application starts, it creates the config file and fails. If it did not fail at least, we could run it with the default configuration.

It would be very easy to deploy if we could inject the configuration using env vars (and use MySQL instead of SQLite).

@da2ce7 since you are working on the [setting overhaul](https://github.com/torrust/torrust-tracker/pull/105), I think you should consider that env vars for configuration is still a common way to inject configuration variables. With Kubernetes, you can [mount a file with secrets](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-pod-that-has-access-to-the-secret-data-through-a-volume). It seems that's a safer way to do it. In that case, we should be able to generate a K8S secret file from the tracker configuration. See https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-config-file/. I think we should make it clear which config values are secrets and which are not. In general, I've always seen people using the same list, for example, the common `.env` file in the root folder, but K8S treat them in a different way. Anyway, maybe that's something to consider only by the people who are deploying the app and we should treat all settings like secrets.

Power2All:
@josecelano I would suggest you check out the Docker deployment on my Axum project. It doesn't need any "volumes" and such, and might be what you're looking for. It uses also ENVIRONMENT variables to be used to build for instance the configuration file, so that it can be deployed anywhere.

josecelano:
> @josecelano I would suggest you check out the Docker deployment on my Axum project. It doesn't need any "volumes" and such, and might be what you're looking for. It uses also ENVIRONMENT variables to be used to build for instance the configuration file, so that it can be deployed anywhere.

hey @Power2All thank you! I've already done it before starting to work on the PR. I think you did a great job. You are generating the `config.toml` file from env vars on the fly when the app starts. I think that's a good solution, but I was looking for a solution that works in both the development and production environments using the same Dockerfile. My idea is to publish "official" multistage docker images you can use for general purposes. Your idea is perfect to deploy the app to production in its current state.

On the other hand, the config.toml file allows you to define the configuration for multiple udp_trackers and http_trackers which is not supported by your solution right now.

Anyway, today @WarmBeer @da2ce7 and I were discussing how to handle the application configuration. It was a long discussion, but to make it short, we agree that that should be definitively a feature and not a hack for deployments. I mean, we should change the application to accept env vars for the configuration because that's a common practice in a lot of hosting providers.

One of our ideas, the want we agreed on at the end, was we could add a new env var (for example 
 `TURRUST_TRACKER_CONFIG`), which could contain the whole configuration at the beginning in toml format and in the future in json format. That way, we can run the tracker in a stateless mode, without requiring any volume, and without providing the config file. We could also do what you did, and add a second way to provide configuration to the application by using env vars.

We discussed a lot of ideas. Maybe we should open a [discussion](https://github.com/torrust/torrust-tracker/discussions) for it. @da2ce7 is working on a refactoring of the [configuration](https://github.com/torrust/torrust-tracker/pull/105) which is related to this. And I also proposed a [new feature](https://github.com/torrust/torrust-tracker/issues/122) to generate the configuration file.

Power2All:
> On the other hand, the config.toml file allows you to define the configuration for multiple udp_trackers and http_trackers which is not supported by your solution right now.

Correct, I was planning to fix that up as well, but had no time to look into that.
Most people won't go that far, or you could grab a config file from somewhere, or using the "secrets" for instance. Been too busy with my new job that is eating a ton of my time.

josecelano:
I've extracted part of this PR into a new [PR](https://github.com/torrust/torrust-tracker/pull/128). It's the HTTPS support for the API.

josecelano:
I've created a new [PR](https://github.com/torrust/torrust-tracker/pull/129) to replace this one.

PR #124, opened 2022-12-08
located errors: database

da2ce7:
Related Issue: #176 

Depends #177 

This is the first in a series of work to add locations to error messages. I have created the [located_error.rs](https://github.com/da2ce7/torrust-tracker/blob/database-error-overhall/src/located_error.rs) class that gathers location from the error call, and make use of it for the database errors.

### DIFF
--- a/tests/api/asserts.rs
+++ b/tests/api/asserts.rs
@@ -118,8 +118,11 @@ pub async fn assert_failed_to_reload_keys(response: Response) {
 async fn assert_unhandled_rejection(response: Response, reason: &str) {
     assert_eq!(response.status(), 500);
     assert_eq!(response.headers().get("content-type").unwrap(), "text/plain; charset=utf-8");
-    assert_eq!(
-        response.text().await.unwrap(),
-        format!("Unhandled rejection: Err {{ reason: \"{reason}\" }}")
+
+    let reason_text = format!("Unhandled rejection: Err {{ reason: \"{reason}");
+    let response_text = response.text().await.unwrap();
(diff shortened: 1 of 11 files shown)


da2ce7:
This pull request is Ready to Review.

tests/api/asserts.rs
```diff
@@ -37,9 +37,20 @@ pub async fn assert_auth_key_utf8(response: Response) -> AuthKey {
...
+    let response_status = response.status().clone();
+    let response_headers = response.headers().get("content-type").cloned().unwrap();
+    let response_text = response.text().await.unwrap();
+
+    let details = format!(
+        r#"
+   status: ´{response_status}´
+  headers: ´{response_headers:?}´
+     text: ´"{response_text}"´"#
+    );
```
josecelano, line 49: hey @da2ce7 , this is an interesting way to have a kind of "object literal" or "anonymous struct". It's related to https://github.com/torrust/torrust-tracker/pull/172

I would not include it in this PR unless we change all the asserts to follow the same pattern. Maybe the final solution could be a mix of this solution and what I proposed on that [PR](https://github.com/torrust/torrust-tracker/pull/172). In this solution, I still miss the error pointing to the origin line when the condition was not fulfilled.

josecelano (approved):

josecelano:
I think this is a very helpful improvement.

ack 9a3ba0668efd7543e45265518cb35e622d6e5a63

da2ce7:
closing in favor of #178

PR #140, opened 2022-12-23
Publish docker image only when secrets are available

josecelano:
See: https://github.com/torrust/torrust-tracker/discussions/132#discussioncomment-4475603

https://hub.docker.com/ does not have scoped tokens. They only have personal tokens. If you want to deploy the docker image with a workflow, you have to use your personal token.

To avoid sharing your personal tokens with all the maintainers, we agreed on using forks. I've changed the workflow. now it is only executed if the secret "DOCKER_HUB_USERNAME" is set in the "dockerhub-torrust" environment.

We can disable the workflow in this repo just by deleting the env vars:

- DOCKER_HUB_USERNAME
- DOCKER_HUB_REPOSITORY_NAME
- DOCKER_HUB_ACCESS_TOKEN

I only check for `DOCKER_HUB_USERNAME`.

We can use the same approach for publishing the crate on https://crates.io

### DIFF
--- a/.github/workflows/publish_docker_image.yml
+++ b/.github/workflows/publish_docker_image.yml
@@ -15,7 +15,21 @@ env:
   TORRUST_TRACKER_RUN_AS_USER: appuser
 
 jobs:
+  check-secret:
+    runs-on: ubuntu-latest
+    environment: dockerhub-torrust
+    outputs:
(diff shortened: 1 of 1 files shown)


da2ce7 (approved):
ACK e4b2a8eb7a24c763c19207a64d273d634abf3626

PR #147, opened 2022-12-23
Workflow to publish crate on crates.io

### LINKED ISSUE #146: Publish it on crates.io
### Discussed in https://github.com/torrust/torrust-tracker/discussions/132

<div type='discussions-op-text'>

<sup>Originally posted by **josecelano** August 10, 2022</sup>
Wouldn't it be useful to publish it on https://crates.io/?

I would like to write more documentation for the package. For example, for API endpoints like this:

image

If we do that, we can automatically have online documentation on https://docs.rs/.

If for some reason, we do not want to publish it yet, we could deploy the documentation to GitHub Pages.

</div>

josecelano:
Workflow to publish the crate on [crates.io](https://crates.io/).

It only works if the secret `CRATES_TOKEN` exists in the `crates-io-torrust` environment.

Since crates.io does not support scoped tokens, we can publish the crate using a fork where the crate owners can set up their crates.io tokens without sharing them with other maintainers.

### DIFF
--- a/.github/workflows/publish_crate.yml
+++ b/.github/workflows/publish_crate.yml
@@ -1,0 +1,54 @@
+name: Publish crate
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  check-secret:
+    runs-on: ubuntu-latest
+    environment: crates-io-torrust
+    outputs:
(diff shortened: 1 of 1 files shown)


josecelano:
ACK 203d356da0706b91084789207460e5ed934c32a4

Issue #144, opened 2022-12-23
API V2: new API with the correct response codes


josecelano:
Parent issue: https://github.com/torrust/torrust-tracker/issues/141
It depends on: https://github.com/torrust/torrust-tracker/issues/143

We start building a new API using the `/v2` prefix. For example: `/v2/api/stats`. We can use the same URLs but change the response codes.

Change the response codes to the right one: 200, 201, 404, etcetera. Right now, it only uses 200 and 500.

### Potential changes

I think we could make other changes:

- Refactor: create a custom Reject for all the errors. Like the one I have added: `struct Unauthorized`.
- Refactor: move routes to their own mod `routes.rs`.
- The `authenticate` function should also return a 401 response instead of `token not valid` when the token is present but not valid.
- The `GET /api/torrent/:info_hash` endpoint should return a 404 when the torrent does not exist and the token is valid.
- The endpoint `GET /api/torrent/:info_hash` should return a 404 when the torrent does not exist, and the token is invalid. See [Insecure Direct Object References (IDOR)](https://portswigger.net/web-security/access-control/idor)). We should not expose that the tracker has a given torrent because the tracker could be private.
- The `DELETE /api/whitelist/:info_hash` endpoint should return: a "204 No Content" when it deletes the torrent and a "404 Not Found" when the torrent does not exist or the token is invalid.
- The `POST /api/whitelist/:info_hash` endpoint should return a "201 Created" response with the location of the new resource `GET /api/whitelist/:info_hash`. In this case we would need to create that new endpoint. It could be empty. It could be used only to know if a torrent is whitelisted. Alternatively, we could consider the whitelist a unique resource and use a `PUT /api/whitelist/:info_hash` to add a new torrent to the list.
- The `POST /api/key/:seconds_valid` endpoint should return a "201 Created".
- The `DELETE /api/key/:key` should return "204 No Content".
- The `GET /api/whitelist/reload` endpoint should be a `POST`, `PUT` or `PATCH`. 

### Notes

- We can keep the old API in parallel until we release the new major version.

### Long-term changes

These are changes not included in this issue, but they should be considered in future versions.

- The `GET /api/whitelist/reload` endpoint is a little weird. It's like a remote command. I think I should have used` POST /api/commands` with the command you want to execute in the request body or `POST /api/reload-whitelist`.
- Same for `GET /api/keys/reload`
[...]

PR #152, opened 2023-01-02
Refactor API: replace Warp with Axum

### LINKED ISSUE #143: API overhaul:  reimplement the current API using Axum
Parent issue: https://github.com/torrust/torrust-tracker/issues/141
It depends on: https://github.com/torrust/torrust-tracker/issues/142

- [ ]  Reimplement the current API using [Axum](https://github.com/tokio-rs/axum)

josecelano:
**UPDATED**: 12/01/2022

The tracker API uses the web framework [Warp](https://github.com/seanmonstar/warp).
This PR replaces Warp with [Axum](https://github.com/tokio-rs/axum).

### Tasks

- [x] Basic Axum configuration.
- [x] Authentication using token in GET params.
- [x] Enable SSL.
- [x] Add tests for the current API. Error cases are not tested. I think all cases are when there is a database error, and it returns an `ActionStatus::Err` error with a message.
- [x] Reimplement endpoints. See subtasks.
- [x] Improve asserts by checking the response body for OK (`200`) responses.
- [x] Use the same URL prefix as the current API `/api/xxx`.
- [x] Remove duplicate definitions of routes in Axum server (`start` and `start_tls`)
- [x] Add tests for the current API. Error cases when GET params cannot be parsed are not tested. `info_hash` and `seconds_valid` Path params, and `offset` and `limit` GET params.
- [x] Fix a bug with the auth `token`. The token must be the first param in the `query` otherwise, it does not work. In the end, it was not a bug. It only happens with `curl`.
- [x] Add test for the current API. When you try to remove a torrent from the whitelist twice, you get a 500 response with this body
: `Unhandled rejection: Err { reason: "failed to remove torrent from whitelist" }`
- [x] Add some tests for malicious user input.
- [x] Remove the Warp implementation.

### Endpoints subtasks

Stats:

- [x] `GET /api/stats`

Torrents:

- [x] `GET /api/torrents?offset=:u32&limit=:u32`
- [x] `GET /api/torrent/:info_hash`

Whitelisted torrents:

- [x] `POST   /api/whitelist/:info_hash`
- [x] `DELETE /api/whitelist/:info_hash`

Whitelist commands:

- [x] `GET /api/whitelist/reload`

Keys:

- [x] `POST   /api/key/:seconds_valid`
- [x] `DELETE /api/key/:key`

Key commands
- [x] `GET /api/keys/reload`


src/setup.rs
```diff
@@ -52,6 +56,19 @@ pub async fn setup(config: &Configuration, tracker: Arc<tracker::Tracker>) -> Ve
         jobs.push(tracker_api::start_job(&config.http_api, tracker.clone()).await);
     }
 
+    // Start HTTP APIs server (multiple API versions)
+    if config.http_api.enabled {
+        // Temporarily running the new API in the 1313 port
+        let bind_address = config.http_api.bind_address.clone();
+        let mut bind_socket: SocketAddr = bind_address.parse().unwrap();
```
da2ce7: prefer expects over unwrap?

josecelano: I've changed it, but we have a lot of "unwraps". For example like this:

```
InfoHash::from_str(&hash).unwrap();
```

I think it would not be very pleasant to write a message for all of them:

```
InfoHash::from_str(&hash).expect("string should be a valid infohash");
```

I have to think about it, but I would say it makes sense when the case is unique.

Maybe the problem is we are duplicating code, and we should add a function like this:

```rust
fn parse_socket_addr_or_fail(bind_address: &str) -> SocketAddr {
    bind_address.parse().expect("the bind address string should contain a valid socket address like this 127.0.1.1:8080")
}
```

josecelano:
hi @da2ce7 @WarmBeer It seems [Axum does not support SSL directly](https://github.com/tokio-rs/axum/issues/677). I can use the [axum-server](https://github.com/programatik29/axum-server), which is the one used in this [fork](https://github.com/Power2All/torrust-axum).

josecelano:
> hi @da2ce7 @WarmBeer It seems [Axum does not support SSL directly](https://github.com/tokio-rs/axum/issues/677). I can use the [axum-server](https://github.com/programatik29/axum-server), which is the one used in this [fork](https://github.com/Power2All/torrust-axum).

I did it using [axum-server](https://github.com/torrust/torrust-tracker/pull/152/commits/fe4303c0af047add8b413e8cb8b4f980e58b7e8c).

josecelano:
I'm working on the `GET /api/torrent/:info_hash` endpoint. I'm getting an error in the serialization of an `u128` field. 

```
#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
pub struct Peer {
    pub peer_id: Id,
    pub peer_addr: String,
    #[deprecated(since = "2.0.0", note = "please use `updated_milliseconds_ago` instead")]
    pub updated: u128,
    pub updated_milliseconds_ago: u128,
    pub uploaded: i64,
    pub downloaded: i64,
    pub left: i64,
    pub event: String,
}
```

I have commented on this issue https://github.com/serde-rs/json/issues/502

I'm going to find out why it's working in the current API (if it's)

josecelano:
The happy path for the endpoint `GET /api/torrent/:info_hash` is done. I've realized I do not have a test for the case where the torrent does not exist. I will add the test for the current API and implement it in the new Axum API.

josecelano:
> I'm working on the `GET /api/torrent/:info_hash` endpoint. I'm getting an error in the serialization of an `u128` field.
> 
> ```
> #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
> pub struct Peer {
>     pub peer_id: Id,
>     pub peer_addr: String,
>     #[deprecated(since = "2.0.0", note = "please use `updated_milliseconds_ago` instead")]
>     pub updated: u128,
>     pub updated_milliseconds_ago: u128,
>     pub uploaded: i64,
>     pub downloaded: i64,
>     pub left: i64,
>     pub event: String,
> }
> ```
> 
> I have commented on this issue [serde-rs/json#502](https://github.com/serde-rs/json/issues/502)
> 
> I'm going to find out why it's working in the current API (if it's)

It works if I do not use `serde` directly:

```
pub async fn get_torrent(State(tracker): State<Arc<Tracker>>, Path(info_hash): Path<String>) -> Response {
    let optional_torrent_info = get_torrent_info(tracker.clone(), &InfoHash::from_str(&info_hash).unwrap()).await;

    match optional_torrent_info {
        Some(info) => Json(Torrent::from(info)).into_response(),
        None => Json(json!("torrent not known")).into_response(),
    }
}
```

Probably for the same reason it's working on the old Warp implementation.

josecelano:
hi @da2ce7 @WarmBeer, this PR is almost finished. I will add more tests to be sure I keep the current behaviour. See pending tasks in the PR description.

The API returns a 500 error if you try to remove a torrent from the whitelist and the torrent is not whitelisted. [We fixed the same problem when you try to add the same torrent twice](https://github.com/torrust/torrust-tracker/issues/74). I will change it even if we break backwards compatibility because we already changed it for the other case. I would expect the same behaviour in both cases. And I do not think users rely on that error response.

da2ce7:
@josecelano Really Great Work!  I am going into the jungle for the next week and bit.  Unfortunately I do not have time to review. Cam.

josecelano:
hi, @WarmBeer I think I've found a bug in both the current API implementation and the new one with Axum. It seems the `token` must be the first `query` param. Curiously enough, I've reproduced the exact error in the new implementation:

```
curl -v http://127.0.0.1:1212/api/torrents?limit=1&token=MyAccessToken -> 500 unauthorized
curl -v http://127.0.0.1:1212/api/torrents?token=MyAccessToken&limit=1 -> 200 OK

curl -v http://127.0.0.1:1313/api/torrents?limit=1&token=MyAccessToken -> 500 unauthorized
curl -v http://127.0.0.1:1313/api/torrents?token=MyAccessToken&limit=1 -> 200 OK
```

I suppose that is not the expected behavior.

josecelano:
> hi, @WarmBeer I think I've found a bug in both the current API implementation and the new one with Axum. It seems the `token` must be the first `query` param. Curiously enough, I've reproduced the exact error in the new implementation:
> 
> ```
> curl -v http://127.0.0.1:1212/api/torrents?limit=1&token=MyAccessToken -> 500 unauthorized
> curl -v http://127.0.0.1:1212/api/torrents?token=MyAccessToken&limit=1 -> 200 OK
> 
> curl -v http://127.0.0.1:1313/api/torrents?limit=1&token=MyAccessToken -> 500 unauthorized
> curl -v http://127.0.0.1:1313/api/torrents?token=MyAccessToken&limit=1 -> 200 OK
> ```
> 
> I suppose that is not the expected behavior.

It seems the problem is only with `curl`. I've added a new test, and it does not fail.

josecelano:
hi @da2ce7 @WarmBeer this is ready to review. I would like to refactor a little bit more, but I think the PR is getting too big. We can merge it as it is and I will create more issues.

I have added some `todos` and `code-reviews` in the code.

I also want to make some improvements, like reorganising the API mod structure. Right now, the folder structure is like this:

```s
src/apis/
├── handlers.rs
├── middlewares
│   ├── auth.rs
│   └── mod.rs
├── mod.rs
├── resources
│   ├── auth_key.rs
│   ├── mod.rs
│   ├── peer.rs
│   ├── stats.rs
│   └── torrent.rs
├── responses.rs
├── routes.rs
└── server.rs
```

It's an implementation-detail or technical-detail organization. I would like to use something more like this:

```s
└── src
    ├── apis
    │   ├── contexts
    │   │   ├── auth_key
    │   │   │   ├── handlers.rs
    │   │   │   ├── mod.rs
    │   │   │   ├── resources.rs
    │   │   │   ├── responses.rs
    │   │   │   └── routes.rs
    │   │   ├── mod.rs
    │   │   ├── stats
    │   │   │   ├── handlers.rs
    │   │   │   ├── mod.rs
    │   │   │   ├── resources.rs
    │   │   │   ├── responses.rs
    │   │   │   └── routes.rs
    │   │   ├── torrent
    │   │   │   ├── handlers.rs
    │   │   │   ├── mod.rs
    │   │   │   ├── resources.rs
    │   │   │   ├── responses.rs
    │   │   │   └── routes.rs
    │   │   └── whitelist
    │   │       ├── handlers.rs
    │   │       ├── mod.rs
    │   │       ├── resources.rs
    │   │       ├── responses.rs
    │   │       └── routes.rs
    │   ├── middlewares
    │   │   ├── auth.rs
    │   │   └── mod.rs
    │   ├── mod.rs
    │   ├── routes.rs
    │   └── server.rs
    └── mod.rs
```

I like the folder structure to be more domain-oriented, and it's also more scalable. If you agree I can add a new issue for that and we can implement it before [implementing the new API version](https://github.com/torrust/torrust-tracker/issues/144).

josecelano:
ACK 0c3ca8798c143828d64b2bbbb13bba7c44fc931

PR #174, opened 2023-02-02
Isolate configuration and primitives in their own crates

### LINKED ISSUE #173: Isolate configuration and primitives in their own crates
This has the benefit that we can import both into other projects. 

For example, I'm working on a [test-helpers crate](https://github.com/torrust/torrust-tracker/pull/170). And from this crate I want to use the tracker `Configuration`. 

In the current state, this is impossible. Because if we want to use the `Configuration` struct, we would have to import it from the `torrust-tracker` crate. So we would have to add the `torrust-tracker` crate to the dependencies of the `test-helpers` crate and that would mean that we have a cyclic dependency (`test-helpers` <-> `torrust-tracker`). Not good!

To avoid this and similar issues, we must isolate code in their own crates as much as possible. This way we can easily selectively import code without running into a cyclic disaster.

mickvandijke:
https://github.com/torrust/torrust-tracker/issues/173


josecelano:
@WarmBeer I think we could use this folder structure:

```text
.
├── packages
│   └── test-helpers
└── src
```
Instead of:

```text
.
├── src
└── test-helpers
```

So that we can add more dependencies and make it more explicit that it's a dependency.

PR #177, opened 2023-02-05
dev: pass through errors to api handlers

da2ce7:
Allow the messages from underlying errors to be passed through to the api handlers.


josecelano (approved):

josecelano:
Hey @da2ce7, this makes the API far more user-friendly. I think it breaks the current API contract since we do not have error codes or a specific error message field in a structured response, but I guess nobody is relying on parsing the exact text response. I suppose the only API client is the [Torrust Index Backed](https://github.com/torrust/torrust-index-backend)

josecelano:
ack 9e30d019fbbf48762e6b1a2cd509e15a19338395

da2ce7:
closing in favor of #178

PR #180, opened 2023-02-09
Axum HTTP tracker: basic scaffolding

josecelano:
This is the first of a series of PR associated with the issue https://github.com/torrust/torrust-tracker/issues/160

We are changing the HTTP tracker to use Axum instead of Warp. This PR allows:

- Running full Warp implementation in production while enabling new implementation.
- Running both implementations for testing while implementing the new Axum version.

It also allows parsing the URL query params for the announce query. For the time being, only the mandatory params.

The announce request URL has two parameters that contain binary data (20 bytes).

Sample Announce URL:

http://0.0.0.0:7070/announce?info_hash=%3B%24U%04%CF%5F%11%BB%DB%E1%20%1C%EAjk%F4Z%EE%1B%C0&peer_id=-qB00000000000000001&port=17548

The "infohash" and "peer id" cannot be extracted using the generic Axum `Query` extractor because it cannot parse binary data. As you can read [here](https://docs.rs/axum/latest/axum/extract/struct.Query.html).

I could have used the [RawQuery extractor](https://docs.rs/axum/latest/axum/extract/struct.RawQuery.html), but I thought this way would be much cleaner because you can pass the `AnnounceParams` into the handler like this:

```rust
pub async fn announce_handler(
    State(_tracker): State<Arc<Tracker>>,
    ExtractAnnounceParams(_announce_params): ExtractAnnounceParams,
) -> Json<Ok> {
    todo!()
}
```

And you can also handle the errors in the extractor if you have to return a "bad request" response. Alternatively, I could have handled parsing and error handling inside the handler.


da2ce7:
ACK 0dc305023772ef5d65d1bf1ae6c41daff44f797e

da2ce7:
ACK 995397eae4a4b96820f52bcd203b2e0bd67be745

da2ce7 (approved):
Looks Good

PR #183, opened 2023-02-10
Remove duplicate code from `announce` request handlers

josecelano:
Before implementing the [new](https://github.com/torrust/torrust-tracker/pull/182) `announce` handler for the Axum HTTP tracker, I needed to clean up the current handlers (UDP and HTTP). The code was duplicated, and I did not want to add a third copy.

I planned to do it after migrating to Axum, but I changed my mind. I think the migration to Axum is going to be easier after this refactor.


josecelano:
@da2ce7 @WarmBeer, this is ready to review. It's only a refactor.

PR #197, opened 2023-02-27
Refactor: `auth::Key`

### LINKED ISSUE #171: Refactor proposal: rename auth key structs
Module: [auth.rs](https://github.com/torrust/torrust-tracker/blob/develop/src/tracker/auth.rs)

We have two structs related to the authentication keys used by the HTTP tracker `Key` and `KeyId`:

```rust
pub struct Key {
    pub key: String,
}

pub struct KeyId(String);
```

There is a pending refactor to use the `KeyId` in the `Key` struct like this:

```rust
pub struct Key {
    pub key: KeyId,
}

pub struct KeyId(String);
```

I introduced the `KeyId` struct because sometimes you need to create the ID, but you do not care about the expiration date. For example, in the handler to delete an auth key:

```rust
pub async fn delete_auth_key_handler(
    State(tracker): State<Arc<Tracker>>,
    Path(seconds_valid_or_key): Path<KeyIdParam>,
) -> Response {
    match KeyId::from_str(&seconds_valid_or_key.0) {
        Err(_) => invalid_auth_key_param_response(&seconds_valid_or_key.0),
        Ok(key_id) => match tracker.remove_auth_key(&key_id.to_string()).await {
            Ok(_) => ok_response(),
            Err(_) => failed_to_delete_key_response(),
        },
    }
}
```

The handler receives the key ID, which is the ID that identifies the key.

I think there are two concepts:

- The token or key which is used to authenticate a client. It's just the secret.
- And the whole key, which is the secret or token with the extra metadata. In our case, only one extra field, which is the expiration date.

What could better names for those two structs @da2ce7 @WarmBeer?

I think we could use something like:

```rust
pub struct RegisteredKey {
    pub key: KeyId,
}

pub struct Key(String);
```

or:

```rust
pub struct ExpiringKey {
    pub key: KeyId,
}

pub struct Key(String);
```

josecelano:
The struct `KeyId` was extracted to wrap the primitive type, but it was not being used in the `auth::Key` struct yet.

`Key` before this PR:

```rust
pub struct Key {
    pub key: String,
}
```

After this PR:

```rust
pub struct ExpiringKey {
    pub id: KeyId,
}
``` 

- [x] Use the new struct `auth::KeyId`in the `auth::Key` struct.
- [x] Rename `auth::Key` to `auth::ExpiringKey`.

### DIFF
--- a/src/apis/resources/auth_key.rs
+++ b/src/apis/resources/auth_key.rs
@@ -3,18 +3,18 @@ use std::convert::From;
 use serde::{Deserialize, Serialize};
 
 use crate::protocol::clock::DurationSinceUnixEpoch;
-use crate::tracker::auth;
+use crate::tracker::auth::{self, KeyId};
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
 pub struct AuthKey {
-    pub key: String,
+    pub key: String, // todo: rename to `id`
     pub valid_until: Option<u64>,
 }
 
 impl From<AuthKey> for auth::Key {
     fn from(auth_key_resource: AuthKey) -> Self {
         auth::Key {
-            key: auth_key_resource.key,
+            id: auth_key_resource.key.parse::<KeyId>().unwrap(),
             valid_until: auth_key_resource
                 .valid_until
                 .map(|valid_until| DurationSinceUnixEpoch::new(valid_until, 0)),
@@ -25,7 +25,7 @@ impl From<AuthKey> for auth::Key {
 impl From<auth::Key> for AuthKey {
     fn from(auth_key: auth::Key) -> Self {
         AuthKey {
-            key: auth_key.key,
+            key: auth_key.id.to_string(),
             valid_until: auth_key.valid_until.map(|valid_until| valid_until.as_secs()),
         }
     }
@@ -37,7 +37,7 @@ mod tests {
 
     use super::AuthKey;
     use crate::protocol::clock::{Current, TimeNow};
-    use crate::tracker::auth;
+    use crate::tracker::auth::{self, KeyId};
 
     #[test]
     fn it_should_be_convertible_into_an_auth_key() {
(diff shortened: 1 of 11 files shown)


josecelano:
Hi @da2ce7 @WarmBeer, I want to decouple authentication (the Key is valid) from authorization  (the peer can get the whitelisted torrent) in the new HTTP Axum implementation.

The `authenticated_request` function has both responsibilities (or authentication inside the authorization):

```rust
pub async fn authenticate_request(&self, info_hash: &InfoHash, key: &Option<KeyId>) -> Result<(), Error> {
    // no authentication needed in public mode
    if self.is_public() {
        return Ok(());
    }

    // check if auth_key is set and valid
    if self.is_private() {
        match key {
            Some(key) => {
                if let Err(e) = self.verify_auth_key(key).await {
                    return Err(Error::PeerKeyNotValid {
                        key_id: key.clone(),
                        source: (Arc::new(e) as Arc<dyn std::error::Error + Send + Sync>).into(),
                    });
                }
            }
            None => {
                return Err(Error::PeerNotAuthenticated {
                    location: Location::caller(),
                });
            }
        }
    }

    // check if info_hash is whitelisted
    if self.is_whitelisted() && !self.is_info_hash_whitelisted(info_hash).await {
        return Err(Error::TorrentNotWhitelisted {
            info_hash: *info_hash,
            location: Location::caller(),
        });
    }

    Ok(())
}
```

That's fine, but I want to use directly the `verify_auth_key` in the HTTP announce handlers to check if the peer is authenticated when the tracker is running in `private` mode. I can do it since the function is public, but I wanted to finish this refactor in order to use a `KeyId` which does not require you to set a zero expiration time. In this context, the key expiration time does not make any sense.

```rust
pub async fn verify_auth_key(&self, key_id: &KeyId) -> Result<(), auth::Error> {
    match self.keys.read().await.get(key_id) {
        None => Err(auth::Error::UnableToReadKey {
            location: Location::caller(),
            key_id: Box::new(key_id.clone()),
        }),
        Some(key) => auth::verify(key),
    }
}
```

This was a pending refactor to use the `KeyId`in the `Key` and rename the `Key` struct.

The refactor is ready to review.

PR #199, opened 2023-02-28
Axum HTTP tracker: `scrape` request in `private` mode

### LINKED ISSUE #196: Axum HTTP tracker: `scrape` request in `private` mode

josecelano:
`scrape` request implementation for tracker in `private` mode.

### DIFF
--- a/tests/http_tracker.rs
+++ b/tests/http_tracker.rs
@@ -2606,8 +2606,7 @@ mod axum_http_tracker_server {
             use crate::http::responses::scrape::{File, ResponseBuilder};
             use crate::http::server::start_private_http_tracker;
 
-            //#[tokio::test]
-            #[allow(dead_code)]
+            #[tokio::test]
             async fn should_return_the_zeroed_file_when_the_client_is_not_authenticated() {
                 let http_tracker = start_private_http_tracker(Version::Axum).await;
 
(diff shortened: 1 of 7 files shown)


tests/http_tracker.rs
```diff
@@ -2677,10 +2705,10 @@ mod axum_http_tracker_server {
                 assert_scrape_response(response, &expected_scrape_response).await;
             }
 
-            //#[tokio::test]
-            #[allow(dead_code)]
+            #[tokio::test]
             async fn should_return_the_zeroed_file_when_the_authentication_key_provided_by_the_client_is_invalid() {
                 // There is not authentication error
+                // code-review: should this really be this way?
```
da2ce7, line 2711: Shouldn't we return a 401?

josecelano: I don't know. I think all responses have status 200 and errors are just an special bencoded dictionary. But in this case this is not even an error in the current Warp implementation. @WarmBeer do you know it?. Where can we find the specification for private trackers?

josecelano: @WarmBeer I think the Axum fork returns an error when the tracker runs in `private` mode, and the `key` is missing, invalid or unknown.

- For `announce`: https://github.com/Power2All/torrust-axum/blob/master/src/http_service.rs#L97-L125
- For `scrape`: https://github.com/Power2All/torrust-axum/blob/master/src/http_service.rs#L240-L268

I can't find any official documentation:

- Bep 48. Tracker Protocol Extension: Scrape: https://www.bittorrent.org/beps/bep_0048.html
- BEP 33. DHT Scrapes: https://www.bittorrent.org/beps/bep_0033.html
- https://wiki.theory.org/BitTorrentSpecification#Unofficial_extensions_to_scrape
- https://wiki.vuze.com/w/Scrape

For me, it makes sense to change this behaviour. I will merge the PR because it's the actual Warp implementation behavior but let me know if we should change it.

PR #214, opened 2023-03-05
Compiling time test with one test binary

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/discussions/188

Hi @da2ce7 @WarmBeer, I've merged the integrations test binaries into one binary to see if the build time decreases, but it takes even longer for some reason.

Maybe I need to do something differently, or maybe it's because we have more tests now.

build-time-with-one-test-binary


josecelano:
I have included this change in this [PR](https://github.com/torrust/torrust-tracker/pull/214). Even if the time seems not to improve, I think test organization is better.

PR #215, opened 2023-03-05
Refactor `ExpiringKey`

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/171

The `ExpiringKey::valid_until` (duration) does not need to be optional anymore since we introduced the `KeyId` struct.

Besides, after renaming the `Key` struct to `ExpiringKey`, there is no longer a conflict with the `KeyId`, so we can use the name `Key` for the `KeyId` again.

Before:

```rust
pub struct ExpiringKey {
    pub id: KeyId,
}
```

After:

```rust
pub struct ExpiringKey {
    pub key: Key,
    pub valid_until: DurationSinceUnixEpoch,
}
```

After this refactoring, you can read the code like this:

"An `ExpiringKey` is a `Key` valid until a date expressed in a duration from the Unix epoch (timestamp)".

### DIFF
--- a/src/apis/handlers.rs
+++ b/src/apis/handlers.rs
@@ -101,15 +101,15 @@ pub async fn generate_auth_key_handler(State(tracker): State<Arc<Tracker>>, Path
 }
 
 #[derive(Deserialize)]
-pub struct KeyIdParam(String);
+pub struct KeyParam(String);
 
 pub async fn delete_auth_key_handler(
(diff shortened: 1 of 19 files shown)


josecelano:
Hi @da2ce7 @WarmBeer, this is ready to review. It's just a minor refactoring.

da2ce7 (approved):
Looks good to me :)

PR #217, opened 2023-03-06
Axum HTTP tracker: add unit tests for custom Axum extractors

### LINKED ISSUE #216: Axum HTTP tracker: add unit tests for custom Axum extractors

josecelano:


josecelano:
Hi @da2ce7 @WarmBeer, this is ready to review. It's just some unit tests for the custom Axum extractors I've written in the new Axum HTTP tracker implementation. I plan to open some more PRs like this. I did not add too many unit tests to the Axum implementation because we had the integration tests and the design was still a proof of concept (since I'm a newbie with Axum and I did not know which solution would be the best).

Besides, as I mentioned before, I prefer more unit tests and fewer integration tests.

PR #226, opened 2023-03-09
Rebase of PR 170

### LINKED ISSUE #156: Fix random test failure due to port collision

josecelano:
Hi @WarmBeer, I've rebased your [PR](https://github.com/torrust/torrust-tracker/pull/170), fixing all merge conflicts.

Do you want to rewrite the history of your changes or can we merge it as it's is?

Now you can rewrite your commits if you want to, with the following:

```s
git reset develop
git add -A 
git commit -m "isolated configuration, test-helpers and primitives crates"
git push --force
```

As you mentioned [here](https://github.com/torrust/torrust-tracker/pull/170#issuecomment-1461803276).


mickvandijke:
Thank you @josecelano !

it can be merged as is :)

josecelano:
ACK 4ad981574ac319ccd9fcd86c425caf0a0ae82168

Issue #351, opened 2023-08-01
(Docker) thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CreatedNewConfigHalt


Slluxx:
Following the documentation and trying this command

```
mkdir -p ./storage/database \
  && mkdir -p ./storage/ssl_certificates \
  && export TORRUST_TRACKER_USER_UID=1000 \
  && docker run -it \
    --user="$TORRUST_TRACKER_USER_UID" \
    --publish 6969:6969/udp \
    --publish 7070:7070/tcp \
    --publish 1212:1212/tcp \
    --volume "$(pwd)/storage":"/app/storage" \
    torrust/tracker:3.0.0-alpha.3
```

results in:

```
Loading configuration from config file ./config.toml
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CreatedNewConfigHalt { location: Location { file: "packages/configuration/src/lib.rs", line: 520, col: 27 }, path: "./config.toml" }', src/bootstrap/app.rs:83:52
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

josecelano:
Hi @Slluxx, I'm sorry, the documentation is ~~unclear~~ wrong in this [section](https://docs.rs/torrust-tracker/3.0.0-alpha.3/torrust_tracker/#run-with-docker).

There are two ways to inject configuration into the application: a `config.toml` file in the root folder or an environment variable `TORRUST_TRACKER_CONFIG` with the same content as the file.

When you start the application, it fails if you do not provide one of those two options, and it creates a default config file in the root folder. Since the docker command only mounts the `storage` folder, the newly generated config file is not persisted.

If you only want to use docker without cloning the repo, you must inject the configuration. You can do it with:

```s
TORRUST_TRACKER_USER_UID=$UID
TORRUST_TRACKER_CONFIG=$(wget -qO- "https://raw.githubusercontent.com/torrust/torrust-tracker/develop/config.toml.local")

mkdir -p ./storage/database \
  && mkdir -p ./storage/config \
  && mkdir -p ./storage/ssl_certificates \
  && touch ./storage/database/data.db \
  && echo ";" | sqlite3 ./storage/database/data.db \
  && docker run -it \
    --user="$TORRUST_TRACKER_USER_UID" \
    --publish 6969:6969/udp \
    --publish 7070:7070/tcp \
    --publish 1212:1212/tcp \
[...]

Issue #317, opened 2023-05-01
Reduce release binary size


nyacat:
Current release profile come with debug info and lto=fat, which is quite unnecessary for release build.
this cause `cargo build --release` generated 60M+ binary file, and is not ideal for docker image or small footprint devices.

```
[profile.release]
debug = 1
opt-level = 3
lto = "fat"
```

result:
```
root@debian-dev:~/torrust-tracker# ls -lah target/release/torrust-tracker
-rwxr-xr-x 2 root root 64M May  1 23:55 target/release/torrust-tracker
root@debian-dev:~/torrust-tracker# file target/release/torrust-tracker
target/release/torrust-tracker: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=somebuild-sha1-hash, for GNU/Linux 3.2.0, with debug_info, not stripped
```

when change the release profile to
```
[profile.release]
strip = true
opt-level = 3
lto = true
codegen-units = 1
panic = "abort"
```

result:
```
root@debian-dev:~/torrust-tracker# ls -lah target/release/torrust-tracker
-rwxr-xr-x 2 root root 11M May  1 23:58 target/release/torrust-tracker
root@debian-dev:~/torrust-tracker# file target/release/torrust-tracker
target/release/torrust-tracker: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=somebuild-sha1-hash, for GNU/Linux 3.2.0, stripped
```

The details are in [link](https://github.com/johnthagen/min-sized-rust)

josecelano:
Thank you, @nyacat! Regarding the `strip=true` option, I removed it some months ago because of this https://github.com/torrust/torrust-tracker/commit/3098ed2c59c420167797b3a6f697c697d440c0f2. I think that's the flag that makes the binary bigger. If that's the case, it's a trade-off between image size and compilation time.

Maybe we could dynamically add it only when we are building docker images for releases, so we have faster builds for each commit and smaller docker images for releases.

Regarding the other options. I don't know how they affect the final binary size. I'll check them, if they do not have any other undesirable side effects, we can change them.

josecelano:
I can confirm that `strip = true` is the option that reduces the size the most:

With `strip = true`:

```s
[...]

PR #363, opened 2023-08-06
Overhaul Containers

### LINKED ISSUE #389: Review Docker Publishing Workflows
Continuation of #363 

Review and rework docker publishing workflows.

da2ce7:
Depends on #391 

Ready for Review. :)

Big Update for Containers.

- [x] Rework Dockerfile
- [x] Rework Container Workflow
- [x] Rework Docker Compose File
- [x] Rework Environmental Variables

Todo:
- [x] Review Documentation.
- [x] Rebase and Make Good Commit History

Moved to a new issue:
- [x] Review Docker Publishing Workflows. closes #389


josecelano:
Hi @da2ce7, I have a couple of comments for now. In general, I like the new structure with the two types of base builds (release and debug).

I did not know:

- cargo binstall
- cargo nextest archive

It seems you are using `cargo nextest archive` to move the builds from one stage to another.

### Rename target

Could we rename `precook_release` to `cook_dependencies`?

These two stages are used to cache cargo dependencies:

```
## Cook (release)
FROM chef AS precook_release
WORKDIR /app/src
COPY --from=make_recipe /app/recipe.json /app/recipe.json
RUN cargo chef cook --tests --benches --examples --workspace --all-targets --all-features --recipe-path /app/recipe.json --release
RUN cargo nextest archive --tests --benches --examples --workspace --all-targets --all-features --archive-file /app/temp.tar.zst --release  ; rm /app/temp.tar.zst

## Cook (debug)
FROM chef AS precook_debug
WORKDIR /app/src
COPY --from=make_recipe /app/recipe.json /app/recipe.json
RUN cargo chef cook --tests --benches --examples --workspace --all-targets --all-features --recipe-path /app/recipe.json
RUN cargo nextest archive --tests --benches --examples --workspace --all-targets --all-features --archive-file /app/temp.tar.zst ; rm /app/temp.tar.zst
```

### Final image size

I was using the slim image because it's smaller:

```s
$ docker images | grep tracker
torrust-tracker                         latest          78409bb883c5   16 minutes ago   949MB
```

```s
$ docker images | grep tracker
torrust-tracker                         latest          1bec1e0efef0   10 seconds ago   30.2MB
```

Smaller images are better because:

- They have fewer dependencies, so they should be safer for production.
- They can be moved faster from docker registries to the production servers, which means faster and cheaper deployments.

See: https://kerkour.com/rust-small-docker-image

### Reorder layers to take advantage of the docker cache

For example, you put the line `COPY --from=test_release /app/bin /app/bin` at the beginning of the stage. If the application changes, you have to build the other layers, like the command to create the user account.

If `USER_ID` or `USER_NAME` change more frequently than the app, that's OK, but I think that's not the case in the local dev env and in the GitHub runners.

In general, the more stable things should be the first layers. In this case, it does not matter because I suppose the other commands after copying the app are fast.

```Dockerfile
# Release Torrust-Tracker
FROM rust:slim as release
WORKDIR /app
COPY --from=test_release /app/bin /app/bin

ARG USER_ID=1000
ARG USER_NAME=appuser
ARG UDP_PORT=6969
ARG HTTP_PORT=7070
ARG API_PORT=1212

ENV USER_ID=${USER_ID}
ENV USER_NAME=${USER_NAME}
ENV UDP_PORT=${UDP_PORT}
ENV HTTP_PORT=${HTTP_PORT}
ENV API_PORT=${API_PORT}
ENV TZ=Etc/UTC

EXPOSE ${UDP_PORT}/udp
EXPOSE ${HTTP_PORT}/tcp
EXPOSE ${API_PORT}/tcp

RUN printf "\n in release mode \n \n run 'exec /app/bin/torrust-tracker' (release build) to start tracker \n \n" > /etc/motd
RUN echo '[ ! -z "$TERM" -a -r /etc/motd ] && cat /etc/motd' >> /etc/bash.bashrc

RUN adduser \
  --disabled-password \
  --gecos "" \
  --home "/nonexistent" \
  --shell "/sbin/nologin" \
  --no-create-home \
  --uid "${USER_ID}" \
  "${USER_NAME}"

RUN chown -R "${USER_NAME}":"${USER_NAME}" /app
USER "${USER_NAME}":"${USER_NAME}"

RUN env
```

### Remove duplicate lines

`release`, `debug` and `development` stages are pretty similar. Maybe we could extract a common base image.

```Dockerfile
FROM rust:slim as base
WORKDIR /app

ARG USER_ID=1000
ARG USER_NAME=appuser
ARG UDP_PORT=6969
ARG HTTP_PORT=7070
ARG API_PORT=1212
ARG BUILD_MODE=release

ENV USER_ID=${USER_ID}
ENV USER_NAME=${USER_NAME}
ENV UDP_PORT=${UDP_PORT}
ENV HTTP_PORT=${HTTP_PORT}
ENV API_PORT=${API_PORT}
ENV TZ=Etc/UTC

EXPOSE ${UDP_PORT}/udp
EXPOSE ${HTTP_PORT}/tcp
EXPOSE ${API_PORT}/tcp

RUN printf "\n in $BUILD_MODE mode \n \n run 'exec /app/bin/torrust-tracker' (debug build) to start tracker \n \n" > /etc/motd
RUN echo '[ ! -z "$TERM" -a -r /etc/motd ] && cat /etc/motd' >> /etc/bash.bashrc

RUN env
```

### Docker compose configuration

The `development` stage only runs a `bash` process, so when you run `docker compose up`, the tracker is but executed by default. You have to log in, compile and run it.

I think we are not using docker compose for any specific purpose on this project now. I only wanted to provide a basic configuration to easily run the tracker with MySQL. The new behavior (running bash instead of the tracker) could be even better if you want to use docker compose for development, but by default, I would run the tracker. You can overwrite the entrypoint if you want to use the docker compose configuration for development.

### Make the binary smaller

In the past, I had to remove the `strip` option from the cargo configuration because the cargo chef cache did not work well with that option. I mean this line:

```Dockerfile
RUN strip /app/target/x86_64-unknown-linux-musl/release/torrust-tracker
```

[It seems it's not needed anymore](https://github.com/LukeMathWalker/cargo-chef/issues/149). If that's the case, we should include again the `strip option` in the `cargo.toml` file.

More info:

- https://github.com/torrust/torrust-tracker/issues/317
- https://github.com/LukeMathWalker/cargo-chef/issues/149

### Docker compose configuration

In the current Dockerfile we are [compiling explicitly to a given architecture](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget):

```Dockerfile
RUN cargo build --release --target x86_64-unknown-linux-musl --bin torrust-tracker
```

I suppose if we do not specify any, the default is the host architecture. That's fine because the current configuration does not allow building for different architectures. That's something we have to consider; maybe we should start to use different tags (prefix or suffix) for docker tags. But I need to learn about this topic. It seems that the GitHub Action we are using to build the docker images already supports [multi-platform builds](https://docs.docker.com/build/ci/github-actions/multi-platform/). So maybe the changes in this PR make building easier for different platforms.

See https://github.com/torrust/torrust-tracker/issues/351#issuecomment-1660822346

To keep track, I think I started using the ` --target x86_64-unknown-linux-musl` option because I was using the alpine image.

Links

- https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images
- https://docs.docker.com/build/ci/github-actions/multi-platform/

### Layer for testing is not working

You told me that part is not working yet:

```Dockerfile
RUN cargo nextest run --workspace-remap /app/src/ --cargo-metadata /app/src/target/nextest/cargo-metadata.json --binaries-metadata /app/src/target/nextest/binaries-metadata.json
```

What was the problem? You can't bind IPv6 addresses to the servers inside docker?

### Conclusion

I think the structure is better and it has several improvements and hopefully, it will make it easier to produce multi-platform images. My main concern is only the image size.

da2ce7:
@josecelano I am starting to become happy with our Dockerfile now. :)

da2ce7:
@josecelano Starting to Look More Complete. Need to continue reworking documentation.

da2ce7:
@josecelano I believe that this Pull Request is Ready to Review. :)

.github/workflows/container.yaml
```diff
@@ -1,84 +1,102 @@
...
   push:
+    tags-ignore:
+      - "!v*"
+  pull_request:
     branches:
-      - "main"
       - "develop"
-    tags:
-      - "v*"
+      - "main"
```
josecelano: Hi @da2ce7 I think with this configuration we publish a docker image for each PR if I'm not wrong. See: https://github.com/docker/metadata-action. Maybe we should only test the docker build for PRs.

.github/workflows/container.yaml
```diff
@@ -1,84 +1,102 @@
-name: Publish docker image
+name: Container
```
josecelano, line 1: Hi @da2ce7, I think the workflow name is too generic. We are testing docker builds, and for some cases, we publish the docker image on DockerHub. But I can't find a better name :-).

Containerfile
```diff
@@ -0,0 +1,138 @@
...
+RUN mkdir -p /app/lib/; cp -l $(realpath $(ldd /app/bin/torrust-tracker | grep "libz\.so\.1" | awk '{print $3}')) /app/lib/libz.so.1
+RUN chown -R root:root /app; chmod -R u=rw,go=r,a+X /app; chmod -R a+x /app/bin
+
+
+## Runtime
+FROM gcr.io/distroless/cc:debug as runtime
+RUN ["/busybox/cp", "-sp", "/busybox/sh","/busybox/cat","/busybox/ls","/busybox/env", "/bin/"]
+COPY --from=tester --chmod=0555 /tmp/su-exec/su-exec /bin/su-exec
+
+ARG TORRUST_TRACKER_PATH_CONFIG="/etc/torrust/tracker/config.toml"
```
josecelano: Hi @da2ce7, should not this be `/etc/torrust/tracker/tracker.toml`?

Containerfile
josecelano, line 1: Hi @da2ce7 are we using the `debug` stages? It seems we are not. I would remove it for now, if we are not using it. It would make the file smaller and it would build faster.

On the other hand, we are testing the "debug" build even if we are not using it.

josecelano (approved):
It looks good to me. I've only made minor comments.

.github/workflows/container.yaml
da2ce7: we do not publish for pull requests for the moment

Containerfile
da2ce7: done. :)

Containerfile
da2ce7: we are not building the debug stage for the moment, I wish to add it to the continuous integration in the future using a matrix.

.github/workflows/container.yaml
da2ce7: for now it is the best name I have found.

da2ce7:
ACK 5da5a25e7416408022aec56379c0e62be132d9c1

PR #366, opened 2023-08-08
deps: localy provided bencode implmentation

da2ce7:


josecelano:
Hi @da2ce7, I want to fix [this bug](https://github.com/torrust/torrust-index-backend/issues/256). I need to calculate the infohash from the torrent file. We are using serde_bencode, but it does not work to calculate the infohash because we cannot include the custom fields. I would like to use a different bencoding crate to calculate the infohash:

Something like this:

```rust
use bendy::decoding::{Error as BencodeError, FromBencode, Object};
use sha1::{Digest, Sha1};
use std::fs;
use std::path::Path;

fn calculate_infohash<P: AsRef<Path>>(path: P) -> Result<[u8; 20], BencodeError> {
    // Read and decode the torrent file
    let data = fs::read(path)?;
    let torrent = Object::decode(&data)?;

    // Find the "info" dictionary
    let info = torrent
        .lookup("info")
        .ok_or(BencodeError::Unspecified)?
        .to_owned();

    // Encode the "info" dictionary into bytes
    let info_bytes = info.encode()?;

    // Calculate the SHA-1 hash
    let mut hasher = Sha1::new();
    hasher.update(&info_bytes);
    let result = hasher.finalize();

    // Convert the hash to a 20-byte array and return
    Ok(result.into())
}

fn main() {
    let path = "path_to_torrent_file.torrent";
    match calculate_infohash(path) {
        Ok(hash) => {
            println!("Infohash: {:?}", hash);
        }
        Err(e) => {
            eprintln!("Error: {}", e);
        }
    }
}
```

I need to use a bencoding crate:

- https://crates.io/crates/bip_bencode
- https://crates.io/crates/bendy
- ...

Since we are already using `bip_bencode` I will probably use it:

```rust
use bip_bencode::{Bencode, BencodeRef, BencodeConvert};
use sha1::{Digest, Sha1};
use std::fs;
use std::io::{self, Read};
use std::path::Path;

fn calculate_infohash<P: AsRef<Path>>(path: P) -> io::Result<[u8; 20]> {
    // Read the torrent file
    let mut data = Vec::new();
    let mut file = fs::File::open(path)?;
    file.read_to_end(&mut data)?;

    // Decode the torrent file
    let torrent = BencodeRef::decode(&data).unwrap();

    // Extract the "info" dictionary
    let info = match torrent.dict().unwrap().lookup(b"info") {
        Some(info) => info,
        None => return Err(io::Error::new(io::ErrorKind::Other, "Missing info dictionary")),
    };

    // Re-encode the "info" dictionary into bytes
    let info_bytes = info.encode();

    // Calculate the SHA-1 hash
    let mut hasher = Sha1::new();
    hasher.update(&info_bytes);
    let result = hasher.finalize();

    Ok(result.into())
}

fn main() {
    let path = "path_to_torrent_file.torrent";
    match calculate_infohash(path) {
        Ok(hash) => {
            println!("Infohash: {:?}", hash);
        }
        Err(e) => {
            eprintln!("Error: {}", e);
        }
    }
}
```

I do not like to parse the torrent bytes twice (with serde_bencode to extract the structs and with bip_encode to extract the info dict and calculate the infohash), but I want to fix the bug asap. After fixing the bug, we can consider using only the bip_encode and construct structs manually instead of using serde_bencode.

da2ce7:
ACK a4ac6829a6a4cd35968231d14f09b91c8be47d6e

Issue #151, opened 2022-12-29
A database should be optional


mickvandijke:
The tracker only needs a database if it wants to persist torrent statistics and private tracker keys between reboots. In all other cases a database should not be necessary.

josecelano:
Hi @WarmBeer, the tracker can persist:

- Authentication keys for the HTTP tracker
- Torrents that are whitelisted
- Number of completed peers (seeders) for each torrent. That's what you call "torrent statistics", don't you?

Right now, the database is mandatory, but you can switch off "statistics persistence" with the core config option: `persistent_torrent_completed_stat`, right?

Could you elaborate on how you think we should implement this issue?

I suppose you want to have an extra config option to totally disable the database or maybe one independent flag for each feature?:

- `persistent_whitelist`
- `persistent_http_auth_keys`
- `persistent_torrent_completed_stat`

mickvandijke:
Hi @josecelano ,

> Hi @WarmBeer, the tracker can persist:
> 
> * Authentication keys for the HTTP tracker
> * Torrents that are whitelisted
> * Number of completed peers (seeders) for each torrent. That's what you call "torrent statistics", don't you?
> 
> Right now, the database is mandatory, but you can switch off "statistics persistence" with the core config option: `persistent_torrent_completed_stat`, right?

Correct.

> Could you elaborate on how you think we should implement this issue?
> 
> I suppose you want to have an extra config option to totally disable the database or maybe one independent flag for each feature?:
> 
> * `persistent_whitelist`
> * `persistent_http_auth_keys`
> * `persistent_torrent_completed_stat`

Yes I think an independent flag for each feature would be nice. We could add a ‘[persistency]’ section to the config options with these three flags. And I suppose that the database url config option could also move to this section.

josecelano:
Making the database optional would allow running the tracker without depending on the file system. We could:

- Not use the database file (`data.db`) required for SQLite.
- Not use the config file (`config.toml`). Use the env var instead. This is already possible.

With no file system dependency, we could run more straightforward docker commands to run a tracker using the docker. Instead of running:

```s
export TORRUST_TRACKER_USER_UID=1000 \
  && docker run -it \
    --user="$TORRUST_TRACKER_USER_UID" \
    --publish 6969:6969/udp \
    --publish 7070:7070/tcp \
    --publish 1212:1212/tcp \
    --volume "$(pwd)/storage":"/app/storage" \
[...]

PR #367, opened 2023-08-08
Improve docker UX

### LINKED ISSUE #361: Improve docker UX
Relates to: https://github.com/torrust/torrust-tracker/issues/351

- [x] Improve the specific documentation for docker: https://github.com/torrust/torrust-tracker/tree/develop/docker
- [x] Write a command to generate the default storage files before running the docker image. The installation process is [here](https://github.com/torrust/torrust-tracker/blob/develop/bin/install.sh), but we need to write a gist or something like that so you do not need to clone the repo.
- [x] We should make the same change we did for the [backend](https://github.com/torrust/torrust-index-backend/blob/develop/src/bootstrap/config.rs). The backend allows changing the default config file location with an env var: `TORRUST_IDX_BACK_CONFIG_PATH`. This way could create the config file in the `storage` folder without injecting the env var (although env va also works).
- [x] Improve the error message when the config file is not found. It should say something like: "Can load the configuration file from ./config.toml"

josecelano:
Some minor improvements to make using docker and the app configuration a smoother experience.

### DIFF
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -512,10 +512,11 @@ impl Configuration {
         if Path::new(path).exists() {
             config = config_builder.add_source(File::with_name(path)).build()?;
         } else {
-            warn!("No config file found.");
-            warn!("Creating config file..");
+            warn!("No config file found. Creating config file ...");
+
             let config = Configuration::default();
             config.save_to_file(path)?;
+
             return Err(Error::CreatedNewConfigHalt {
                 location: Location::caller(),
                 path: path.to_string(),
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -228,14 +228,11 @@
 //!```
 use std::collections::{HashMap, HashSet};
 use std::net::IpAddr;
-use std::panic::Location;
-use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::{env, fs};
 
 use config::{Config, ConfigError, File, FileFormat};
-use log::warn;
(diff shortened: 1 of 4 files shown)


josecelano:
There is a breaking change. I have removed the error when the default configuration is created:

```s
Loading configuration from config file ./config.toml
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CreatedNewConfigHalt { location: Location { file: "packages/configuration/src/lib.rs", line: 520, col: 27 }, path: "./config.toml" }', src/bootstrap/app.rs:83:52
```

Now the tracker continues running with the default configuration file:

```s
Loading configuration from configuration file: `./config.toml` ...
Missing configuration file.
Creating a default configuration file: `./config.toml` ...
Please review the config file: `./config.toml` and restart the tracker if needed.
2023-08-08T12:16:24.548218890+01:00 [torrust_tracker::bootstrap::logging][INFO] logging initialized.
2023-08-08T12:16:24.548805484+01:00 [torrust_tracker::bootstrap::jobs::tracker_apis][INFO] Starting Torrust APIs server on: http://127.0.0.1:1212
2023-08-08T12:16:24.548882463+01:00 [torrust_tracker::bootstrap::jobs::tracker_apis][INFO] Torrust APIs server started
```

If we implement [the issue to make the database optional](https://github.com/torrust/torrust-tracker/issues/151) and we make the no-database option the default one, we could run the tracker with default configuration (only memory) with a single docker command.


Alternatively, we could add a new env var `TORRUST_TRACKER_USE_DEFAULT_CONFIG`. If the env var is present, we do not even write the config file with the default config, we simply use the default config only in memory. What do you think @da2ce7?

Issue #326, opened 2023-05-16
Add middleware to compress responses


josecelano:
The HTTP tracker and the API do not send a compressed response.

[tower_http::compression](https://docs.rs/tower-http/0.4.0/tower_http/compression/index.html) middleware could be used.

### Links

- https://tokio.rs/blog/2021-07-announcing-axum#middleware

alexohneander:
@josecelano Do we want to use this as default or do we want to enable/disable this via a config line?

da2ce7:
Hello @alexohneander HTTP since the big-bang has supported compression, I sugest that we just blanket enable it for everything; it is a cheap and effective optimization.

PR #385, opened 2023-08-26
#326: Add tower-http compression as middleware

alexohneander:
**Modified:**
 I have added tower-http compression as middleware on both HTTP routers(API, Tracker HTTP).

All tests were run and were okay.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ uuid = { version = "1", features = ["v4"] }
 axum = "0.6.20"
 axum-server = { version = "0.5", features = ["tls-rustls"] }
 axum-client-ip = "0.4.1"
+tower-http = { version= "0.4.3", features = ["full"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -41,6 +41,7 @@ uuid = { version = "1", features = ["v4"] }
 axum = "0.6.20"
 axum-server = { version = "0.5", features = ["tls-rustls"] }
 axum-client-ip = "0.4.1"
+tower-http = { version= "0.4.3", features = ["full"] }
```
da2ce7, line 44: It would be nice not to include the `full` tower-http feature set, this will increase build times.

da2ce7: Here are the different feature-sets to select from: https://github.com/tower-rs/tower-http/blob/master/tower-http/Cargo.toml#L55

alexohneander: @da2ce7 You are absolutely right.
I have adjusted it. I will create a new pull-request.

PR #399, opened 2023-09-06
ci: use coverage upload token

da2ce7:


josecelano (approved):
Hi @da2ce7, It looks good. I've seen you have created a repository token. Just a reminder for other contributors: in general, we prefer a more restricted scope for secrets, like creating a new environment for codecov and linking the secret to that environment. In this case, I suppose the actions for the token are very limited and with a repository secret we can generate the coverage report everywhere.

da2ce7:
ACK 5bea92d5c62ed4bd38d1033f5ed9d4393cf4a8b6

PR #409, opened 2023-09-10
dev: use releases branch prefix

### LINKED ISSUE #407: Review Deployment Processes
GitHub no longer supports environments from tags. - This means we need to rework our deployment process using the new namespace convention: `release/**/*`.

It is possible to create the tag, after the deployment environment has executed; and in the ruleset enforce this.

It appears that the new semantics versioning release branches follow the following convention: e.g. `refs/heads/releases/v1.2.3-beta.1`

After making the deployment; then we can make a tag with the same version. However this will not trigger any workflow.

GitHub no-longer supporting tags as deployment environments is surprising; and somewhat of an unintuitive choice.

da2ce7:
closes #407

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -5,8 +5,7 @@ on:
     branches:
       - "develop"
       - "main"
-    tags:
-      - "v*"
+      - "releases/**/*"
   pull_request:
     branches:
       - "develop"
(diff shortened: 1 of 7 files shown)


da2ce7:
ACK 6665a059bb1668d64513b6bc61ae7c5184b0bde7

PR #438, opened 2023-09-15
ci: fix for bug in upstream lib

da2ce7:
testing unit coverage step triggers bug in time crate on nightly: https://github.com/time-rs/time/issues/618


da2ce7:
ACK 27a7f92128dbd08f1962a2300d6b5a9935a25282

Issue #70, opened 2022-08-16
Add compiled binaries for different targets to our releases


mickvandijke:

onedr0p:
This would be great!

Issue #325, opened 2023-05-12
Torrust-Tracker won't compile on Windows


Power2All:
Just tried to compile this on Windows CLion with Rust.
These modules are not compatible with Windows:

* ring
* libz-sys
* libsqlite3-sys
* openssl-sys

You might want to look into replacing these, or looking to remove them.
Unless of course, you won't want to support multi-platforms.

Power2All:
Found the problem why it won't compile on Windows.
You use OpenSSL, which breaks.
Would advice to switch to Rustls, which does compile correctly on all platforms.

josecelano:
> Found the problem why it won't compile on Windows. You use OpenSSL, which breaks. Would advice to switch to Rustls, which does compile correctly on all platforms.

Thank you @Power2All. I have to spend some time to understand the problem. Rigth now I only know:

- We changed [here](https://github.com/torrust/torrust-tracker/issues/68) the `openssl` dependency to `vendored`.
- We have had some problems because of this because in fact, we need other system dependencies like make. See https://github.com/torrust/torrust-tracker/issues/319.

I think we should:

1. Decide which platforms we want to support.
2. Make sure the app compiles for all of them.
3. Add a workflow to check cross-compilation as I mentioned [here](https://github.com/torrust/torrust-tracker/issues/68#issuecomment-1340931904).

I guess we want to support Windows unless there is a major performance issue or something like that. I think there is no active contributor using Windows for development right now.

cc @da2ce7 @WarmBeer

Power2All:
> We changed https://github.com/torrust/torrust-tracker/issues/68 the openssl dependency to vendored.

I tried this too, but didn't work out as expected, as it's force to use MSVC, which you definitely don't want to use for compiling.

I've fixed all the issues on my end by skipping most of these -sys libraries it tries to use.
OpenSSL seems to be the only issue, it does this in combination with ring library, hence the issue.

> I think there is no active contributor using Windows for development right now.

Before most of the changes, I was the one using Windows, with CLion as IDE.
Since I've been too busy with other projects, I wasn't paying attention on the changes on Torrust-Tracker (I've noticed RwLock is a performance hit, like a lot, compared when using Crossbeam Channel for mutex read and write actions, which I will be trying to work out anyway).

GGLinnk:
I investigated one of the 5 current test problems.

It seems that reqwest behaves differently under windows because of the way ports are managed.
[...]

Issue #319, opened 2023-05-02
Compilation error: failed to run custom build command for `openssl-sys v0.9.85`


josecelano:
I get an error compiling on a new machine with a newer Ubuntu version. We are using the "[vendored](https://docs.rs/openssl/latest/openssl/#vendored)" feature of the `openssl` crate.

_If the vendored Cargo feature is enabled, the openssl-src crate will be used to compile and statically link to a copy of OpenSSL. The build process requires a C compiler, perl (and perl-core), and make. The OpenSSL version will generally track the newest OpenSSL release, and changes to the version are not considered breaking changes._

```toml
openssl = { version = "0.10", features = ["vendored"] }
```

openssl = `0.10.50`

Ubuntu:

```s
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 23.04
Release:	23.04
Codename:	lunar
```

Output:

```s
   Compiling tokio v1.27.0
[...]

Issue #242, opened 2023-03-13
torrust-tracker fails to compile


Ryahn:
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.5 LTS
Release:	20.04
Codename:	focal

rustc 1.68.0 (2c8cc3432 2023-03-06)
binary: rustc
commit-hash: 2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74
commit-date: 2023-03-06
host: x86_64-unknown-linux-gnu
release: 1.68.0
LLVM version: 15.0.6

cargo 1.68.0 (115f34552 2023-02-26)

```process didn't exit successfully:...(signal: 9, SIGKILL: kill)```

```root@srv75133557:/opt/torrust/torrust-tracker# cargo build --release
  ...
   Compiling torrust-tracker v2.3.0 (/opt/torrust/torrust-tracker)
error: could not compile `torrust-tracker`

Caused by:
[...]

PR #473, opened 2023-10-06
feat: Add Release Pipeline #70

alexohneander:
built a first draft of how we can attach binaries to our release.
As soon as we push a tag, a release will be created and the built binaries will be attached to it.
You can see how it looks like in my repo:
https://github.com/alexohneander/torrust-tracker/releases/tag/v0.0.16


josecelano:
Hi @alexohneander thank you. It looks good, but we have to make some changes. @da2ce7 has defined a [release process](https://github.com/torrust/torrust-tracker/blob/develop/docs/release_process.md). I think we should only publish a new release when we push to the releases branches like in the [deployment](https://github.com/torrust/torrust-tracker/blob/develop/.github/workflows/deployment.yaml) workflow.

```yaml
on:
  push:
    branches:
      - "releases/**/*"
```

alexohneander:
@josecelano Should be one line, I'll see if that works.

alexohneander:
@josecelano We need a Tag to create a Release.
```
Error: ⚠️ GitHub Releases requires a tag
```

It becomes a little difficult. Of course, we could create a tag from the branch. 
But I honestly don't find that very nice.

da2ce7:
@alexohneander @josecelano This pull request needs a muli-iterative review process. We should merge the workflow untested, then follow the release process:
- #483 

To check if the process works as intended.

josecelano:
> @josecelano We need a Tag to create a Release.
> 
> ```
> Error: ⚠️ GitHub Releases requires a tag
> ```
> 
> It becomes a little difficult. Of course, we could create a tag from the branch. But I honestly don't find that very nice.

hi @alexohneander, sorry for the delay, but I was on a trip and later ill.

The process defined by @da2ce7 already contains [a step to create a tag.](https://github.com/torrust/torrust-tracker/blob/develop/docs/release_process.md#6-create-release-tag). In fact, the next step is to publish the release, but it's supposed to be done manually, right @da2ce7 ?

I think your first version was right, I was confused because we do not want to create a release for tags without a release branch. If we follow strictly the release process that would not happen but as far as I know the rule set defined by @da2ce7 does not protect us from creating new releases from a tag without an associated release branch.

Sorry, but we were in the middle of refactoring the release process and I have not had time to review all the work done by @da2ce7 .

I think we could use the push tag event with this pattern `v*.*.*` instead of the one you had originally `*`. We could do the following, you can change the event to:

```yml
on:
  push:
    tags:
      - "v*.*.*"
```

Then I will merge the PR as @da2ce7 suggested [here](https://github.com/torrust/torrust-tracker/pull/473#issuecomment-1758107157) and we can [follow the release process](https://github.com/torrust/torrust-tracker/issues/483) to check if it works as expected.

josecelano (approved):

josecelano:
Sorry @alexohneander I know It can be pretty annoying, but I can't merge if commits are not signed.  We set that rule. we want to promote signing commit. Could you be so kind to sign them with a rebase? Otherwise, I can do it and put you as the author. What do you think about this policy? It seems most people don't care too much about it? We do and we have created this site https://secure-git.guide/ to promote it.

alexohneander:
@josecelano Sure I can make this, then I can directly make a squash commit too. What I think of the rule. I think it is unfamiliar for many and since you have to configure that extra, it is a step that you like to forget.

josecelano:
> @josecelano Sure I can make this, then I can directly make a squash commit too. What I think of the rule. I think it is unfamiliar for many and since you have to configure that extra, it is a step that you like to forget.

I have this config in my .gitcongin

```toml
[user]
        name = Jose Celano
        ....
        signingkey = ..XXX....AX

[commit]
	gpgsign = true
```

So you always automatically sign all commits. If you want you can even save the password so you are not asked for it every session although I prefer to use the GPG agent.

alexohneander:
@josecelano I had a little problem with my E-Mail. Sorry for that.
Should be fine now.

josecelano:
> @josecelano I had a little problem with my E-Mail. Sorry for that. Should be fine now.

Hi @alexohneander,

Yesterday, we (@da2ce7, @mario-nt, @WarmBeer, @grmbyrn and I) were discussing this PR. They are all member of the Nautilus Team and collaborating with the Torrust Org.

The meeting summary: https://github.com/nautilus-cyberneering/.github/discussions/62

I explained the PR and we have some questions/concerns. First of all, I have to say we all appreciate your contribution but we want to know more about the reason why you want to generate the pre-compiled binaries. I will tell my assumptions and you can correct me if I'm wrong.

- It would be easier for people just to download the binaries and run them instead of compiling, which requires setting up a lot of dependencies.
- When you build on a GitHub workflow, the binary will be only compatible with the same target the build was executed on, which means the same system the GitHub runner is using if I'm not wrong. So the binaries will work on a limited number on targets.

### Concerns

On the other hand, the application needs some system dependencies and we have had problems with compilation:

- [ ] https://github.com/torrust/torrust-tracker/issues/325
- [ ] https://github.com/torrust/torrust-tracker/issues/319
- [ ] https://github.com/torrust/torrust-tracker/issues/242
- [ ] https://github.com/torrust/torrust-tracker/issues/54
- [ ] https://github.com/torrust/torrust-tracker/issues/37
- [ ] https://github.com/torrust/torrust-tracker/issues/40
- [ ] https://github.com/torrust/torrust-tracker/issues/325
- [ ] https://github.com/torrust/torrust-tracker/issues/317

Besides, we added support for docker to try to make it easier and remove all these compatibility problems.

@da2ce7 thinks it could not be the best way to distribute the application, but we did not have time during the meeting yesterday to go into the details.

We would like to have more feedback about your use case and your opinion about this topic so we can discuss it again next week. Sorry for not having given you better feedback since the beginning. But for the last two weeks, we have been attending a conference and also not 100% focussed on the project.

I have also done a test that could be useful to discuss this further. I have compiled the application on my laptop (although I have all the needed dependencies) and created a repo to use the rebuild binaries. This is the repo:

https://github.com/josecelano/torrust-tracker-execution-from-binaries

It works because the build platform and the target platform are the same. I only wanted to ensure whether it works in a clean system (same OS as mine).

### Links

- https://www.docker.com/blog/cross-compiling-rust-code-for-multiple-architectures/
- https://users.rust-lang.org/t/how-crossplatform-is-rust-really/6331/9
- https://rust-lang.github.io/rustup/cross-compilation.html

alexohneander:
@josecelano there was an issue with it, I went ahead and fixed it. But if we don't need it that way, that's okay.

josecelano:
> @josecelano there was an issue with it, I went ahead and fixed it. But if we don't need it that way, that's okay.

Right now, we have two ways to install the tracker:

- Clone the repo, install the system dependencies (Rust and others), build the app and setup configuration.
- User docker and setup configuration.

If users download the binaries, they only would need one step less in the first option. They would not need to install Rust. Since we have some system dependencies and the app setup is not that easy. It seems that having the binaries in that case would make things much more manageable. I do not know if that's the reason why @da2ce7 told us it's not so helpful. I'm waiting for his feedback. I do not see any problem with adding the binaries to the release as long as we:

- Generate documentation for the installation process using the pre-build binaries
- Do not generate frustration for users. They might think you only need to download the binaries and run them.

If you do not mind I would keep the PR open to get more feedback.

josecelano:
Relates to: https://github.com/cargo-bins/cargo-binstall

da2ce7:
@alexohneander thankyou for your work, I have rebased your pull request.

I also think that supporting `cargo binstall` would be something that is quire interesting.

PR #512, opened 2023-11-21
Implement a health check for the tracker container

### LINKED ISSUE #508: Implement health check for the tracker application
Implement a HEALTHCHECK for the whole application. [We are doing the same for the index](https://github.com/torrust/torrust-index/issues/254). 

The goal is to add a line similar to this:

```dockerfile
HEALTHCHECK --interval=5s --timeout=5s --start-period=3s --retries=3 \  
  CMD    /usr/bin/health_check --http http://localhost:${API_PORT}/health_check \
      && /usr/bin/health_check --http http://localhost:${HTTP_PORT}/health_check \
      && /usr/bin/health_check --udp udp://localhost:${UDP_PORT} \
      || exit 1
````

So when you run the tracker in a container, it can show its status (healthy or not).

For the HTTP tracker and Tracker API, we can add `health_check` endpoints like in the [Index](https://github.com/torrust/torrust-index/issues/254).

For the UDP Tracker, maybe we can try to connect.

### PR DESCRIPTION
**Pull request by josecelano**
We need to check the three services provided by the container:

- [x] API
- [x] HTTP Tracker (1 or more)
- [x] UDP Tracker (1 or more)

And we also need to:

- [x] Check them only when they are enabled in the configuration.

### Implementation

- [x] High-level health-checker API (in the future management API). It only checks enabled services. It makes a request to the service healthcheck endpoints.
- [x] Healthcheck API endpoint
- [x] Healthcheck HTTP Tracker endpoint
- [x] Healthcheck UDP Tracker request (using the `connect` request endpoint).

With the default configuration, you can use the endpoint: http://localhost:1313/health_check

It will return an OK Json response:

```json
{
  "status": "Ok", 
  "message": ""
}
```

or an Error response:

```json
{ 
  "status": "Error",
  "message": "API is not healthy. Health check endpoint: http://127.0.0.1:1212/health_check"
}
```

**NOTICE**: health checks are not executed when services use port 0 in the configuration. Service launchers must be changed to return the bound port so the health checker handler can connect.


josecelano:
Hi @da2ce7 @mario-nt @WarmBeer this is ready for review.

For the time being, it only actually checks the internal services (API, UDP and HTTP tracker) if the port defined in the configuration is not 0. It's a big effort to get the real port back after launching those services. We have some pending issues for that. We can improve this PR after refactoring those services. I think this solution is enough at least to check that containers are "ready".

If you think it's a good solution I can add the missing tests at least for the part that it's easily testable.

josecelano:
ACK 5e0a686023862f22e24ea9ffea6530f5d9b83efb

PR #531, opened 2023-12-03
dev: refactor: rename tracker mod to core

### LINKED ISSUE #255: Rename `tracker` mod to `core`
I propose renaming the root module `tracker` to `core`.

Mod: https://github.com/torrust/torrust-tracker/tree/develop/src/tracker

```s
$ tree src/core/
src/core/
├── auth.rs
├── databases
│   ├── driver.rs
│   ├── error.rs
│   ├── mod.rs
│   ├── mysql.rs
│   └── sqlite.rs
├── error.rs
├── mod.rs
├── peer.rs
├── services
│   ├── mod.rs
│   ├── statistics
│   │   ├── mod.rs
│   │   └── setup.rs
│   └── torrent.rs
├── statistics.rs
├── torrent.rs
└── tracker.rs
```

That will allow having these namespaces:

- `core::tracker::Tracker`
- `core::peer::Peer`
- Etcetera.

In my opinion, that's better than the:

- `tracker::Tracker`
- `tracker::peer::Peer`

The mod contains more things than just the tracker, and that `Tracker` could even disappear in the future. I think what we have is the domain logic for BitTorrent trackers, regardless of which delivery mechanism you use (the way end-user interact with that domain logic).

I'm writing the [documentation for the crate](https://github.com/torrust/torrust-tracker/issues/253), and it makes sense to give it a name to the "core" tracker logic.

If we move the mod to a package in the future, the name would be `torrust-tracker-tracker` if we follow current conventions. I think `torrust-tracker-core` would be a better name.

I'm open to more options. I think `core` may be too generic.

Other alternatives:

- `torrust-tracker-domain`
- `torrust-bittorrent-tracker`

We must consider that we also have another package called `shared` with logic used by the `core`, `http`, `udp`, ...

cc @da2ce7 @WarmBeer

da2ce7:
closes #255


josecelano:
ACK 986ab64316fc2df977b58115bc8ba218ca0a896a

PR #538, opened 2023-12-14
Move API healthcheck endpoint from `/health_check` to `/api/health_check`

### LINKED ISSUE #537: Move API healthcheck endpoint from `/health_check` to `/api/v1/health_check`
I'm working on setting up a live demo on https://index.torrust-demo.com/.

See https://github.com/torrust/torrust-compose/issues/2.

I'm using Nginx as a reverse proxy. The HTTP tracker has these direct URLs:

- API:          http://tracker.torrust-demo.com:1212/health_check
- HTTP Tracker: http://tracker.torrust-demo.com:7070/health_check
- UDP Tracker:  udp://tracker.torrust-demo.com:6969

But you can load them via the reverse proxy with https:

- API:          http://tracker.torrust-demo.com/api/health_check -> The API healthcheck endpoint.
- HTTP Tracker: http://tracker.torrust-demo.com/health_check     -> This is the HTTPO tracker healthcheck endpoint.

The API healthcheck endpoint is not accessible from the proxy because I'm using the URL segment `api` to redirect requests for the Tracker API.

HTTP Nginx config for the tracker:

```
server
{
	listen 80;
	listen [::]:80;

	root /var/www/html;
	index index.html index.htm index.nginx-debian.html;

	server_name tracker.torrust-demo.com;

	location /api/
	{
		proxy_pass http://tracker:1212/api/;
	}

	location /
	{
		proxy_pass http://tracker:7070;
	}

	location ~ /.well-known/acme-challenge
	{
		allow all;
		root /var/www/html;
	}
}
```

The reason I used the endpoint `/health_check` instead of `/api/v1/health_check` is because:

- I wanted a generic API healthcheck regardless of the API version.
- I wanted that endpoint to be public, not requiring a token.

I could add a new redirection in the Nginx configuration, something like:

```
	location /health_check
	{
		proxy_pass http://tracker:1212/health_check;
	}
```

However, the HTTP Tracker health check endpoint would not be accessible.

Maybe I can use `/api/health_check` and apply the auth middleware only to `/api/v1/*`.

What do you think @da2ce7 @WarmBeer?

josecelano:
Relates to: https://github.com/torrust/torrust-compose/issues/2

From: `GET /health_check`
To:   `GET /api/health_check`

To avoid collission with HTTP Tracker health check enpoint when the API and the HTTP Tracker are using the same domain+port. For example:

- API:          https://tracker.com:443/api/
- HTTP tracker: https://tracker.com:443/

Old health check endpoints:

- API:          https://tracker.com:443/health_check
- HTTP tracker: https://tracker.com:443/health_check

New API health check endpoint:

- API:          https://tracker.com:443/api/health_check
- HTTP tracker: https://tracker.com:443/health_check

### DIFF
--- a/src/bin/http_health_check.rs
+++ b/src/bin/http_health_check.rs
@@ -11,7 +11,7 @@ async fn main() {
     let args: Vec<String> = env::args().collect();
     if args.len() != 2 {
         eprintln!("Usage:   cargo run --bin http_health_check <HEALTH_URL>");
-        eprintln!("Example: cargo run --bin http_health_check http://127.0.0.1:1212/health_check");
+        eprintln!("Example: cargo run --bin http_health_check http://127.0.0.1:1212/api/health_check");
         std::process::exit(1);
     }
 
--- a/src/servers/apis/routes.rs
+++ b/src/servers/apis/routes.rs
@@ -29,6 +29,6 @@ pub fn router(tracker: Arc<Tracker>) -> Router {
             tracker.config.clone(),
             v1::middlewares::auth::auth,
         ))
-        .route("/health_check", get(health_check_handler))
+        .route(&format!("{api_url_prefix}/health_check"), get(health_check_handler))
         .layer(CompressionLayer::new())
 }
--- a/src/servers/apis/v1/context/health_check/mod.rs
+++ b/src/servers/apis/v1/context/health_check/mod.rs
@@ -8,14 +8,14 @@
 //!
 //! # Health Check
 //!
-//! `GET /health_check`
+//! `GET /api/health_check`
 //!
 //! Returns the API status.
 //!
 //! **Example request**
 //!
 //! ```bash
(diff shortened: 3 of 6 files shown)


josecelano:
ACK: 985633e17c0ac341bca72b0fe78c30a64bf97394

da2ce7:
@josecelano 

The original plan was to have it under the v1 prefix, but this pull request doesn't include the version. Hmm?

josecelano:
> @josecelano
> 
> The original plan was to have it under the v1 prefix, but this pull request doesn't include the version. Hmm?

Hi @da2ce7 I changed my mind. In the end, I did what I mentioned at the end of the issue:

image

https://github.com/torrust/torrust-tracker/issues/537#issue-2037507723

- `/api/heal_check` is a public endpoint.
- For `/api/v1/*` you need the token.

That means we are not versioning the `health_check` endpoint for now because we only use it for container health checks.

Issue #401, opened 2023-09-06
Overhaul Configuration and Settings for Torrust


da2ce7:
Our current configuration system is a mess and not sustainable.

I suggest that we move to the [Figment](https://github.com/SergioBenitez/Figment) configuration library and conceptually rework our entire approach to configuration for the tracker.

josecelano:
Related to: https://github.com/torrust/torrust-tracker/discussions/127

### Subtasks (UPDATED 2024-05)

- [x] https://github.com/torrust/torrust-tracker/issues/850
- [x] https://github.com/torrust/torrust-tracker/issues/852
- [x] https://github.com/torrust/torrust-tracker/issues/856
- [x] https://github.com/torrust/torrust-tracker/issues/855
- [x] https://github.com/torrust/torrust-tracker/issues/851
- [x] https://github.com/torrust/torrust-tracker/issues/878

### More subtasks (UPDATED 2024-07)

After an internal discussion, we decided to make more changes:

- [x] https://github.com/torrust/torrust-tracker/issues/932
- [x] https://github.com/torrust/torrust-tracker/issues/939
- [x] https://github.com/torrust/torrust-tracker/issues/935
- [x] https://github.com/torrust/torrust-tracker/issues/936
- [x] https://github.com/torrust/torrust-tracker/issues/937
- [x] https://github.com/torrust/torrust-tracker/issues/948
- [x] https://github.com/torrust/torrust-tracker/issues/950
- [x] https://github.com/torrust/torrust-tracker/issues/958
- [x] https://github.com/torrust/torrust-tracker/issues/938

josecelano:
Some ideas related to configuration:

- https://github.com/torrust/torrust-tracker/discussions/134
- Configuration should be passed with a file or env vars. It can be passed as a whole file. Regarding env vars there is one for the whole file contents and others used to overwrite some values.
- We are allowing user to overwrite some options after loading the configuration file.
- It would be nice to validate the configuration before running the services when possible, in order to fail as soon as possible. We are doing that for one case. That implies to parse invalid configuration from a config file and create a secondary internal structure that is valid.
- It would be nice to allow users to define only the config options they want to change from the default configuration instead. See https://github.com/greatest-ape/aquatic/pull/191/files/c5843eedce6ccbc8f69e3b3f48ade84f705e70ba#r1527138893. This could be a problem if default values change. The user could not realise it's using different values.

josecelano:
[...]

PR #557, opened 2023-12-28
Config Overhaul

da2ce7:
- [x] #561
- [x] #563
- [x] #623
- [x] #644 
- [x] dev: improve announce ip logic test: (c7b28deb09303783fb6cdff09f773450d179440b)

### DIFF
--- a/src/servers/http/v1/handlers/announce.rs
+++ b/src/servers/http/v1/handlers/announce.rs
@@ -78,6 +78,11 @@ async fn handle(
    See https://github.com/torrust/torrust-tracker/discussions/240.
 */
 
+/* code-review (da2ce7): The IP that is supplied with the announce request is completely dropped
+   in favour of the IP supplied by the `client_ip_sources`.
+   Is this intended behavior, is the IP in the client announce request completely untrusted?
+*/
+
 async fn handle_announce(
     tracker: &Arc<Tracker>,
     announce_request: &Announce,
--- a/tests/servers/http/asserts.rs
+++ b/tests/servers/http/asserts.rs
@@ -4,6 +4,7 @@ use reqwest::Response;
 
 use super::responses::announce::{Announce, Compact, DeserializedCompact};
 use super::responses::scrape;
+use crate::servers::http::responses::announce::DictionaryPeer;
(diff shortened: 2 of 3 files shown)


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/343

josecelano:
Hi @da2ce7 I think in the past we have discussed having two types of configuration structs:

- The DTO: the ones representing the information in the config files or in general the configuration injected into the app when it was started.
- The domain configuration. After receiving the configuration in most cases the application only fails when it starts using the wrong values. For example, if the configuration for one service is wrong you do not know it unless you start that service. For example, if you want to use TLS for the API service and you don't provide the certificates you would not realise the configuration was wrong until you enable that service.

In the Index I've recently introduced a validation for the configuration:

https://github.com/torrust/torrust-index/blob/develop/src/config.rs#L571-L601

I wanted the application to fail if you use an invalid tracker configuration. In this case, you can't not have a public UDP tracker. We don't support it.

I'm still using the same struct. For the time being, I've just added a validate function.

```rust
    /// # Errors
    ///
    /// Will return an error if the configuration is invalid.
    pub async fn validate(&self) -> Result<(), ValidationError> {
        self.validate_tracker_config().await
    }

    /// # Errors
    ///
    /// Will return an error if the `tracker` configuration section is invalid.    
    pub async fn validate_tracker_config(&self) -> Result<(), ValidationError> {
        let settings_lock = self.settings.read().await;

        let tracker_mode = settings_lock.tracker.mode.clone();
        let tracker_url = settings_lock.tracker.url.clone();

        let tracker_url = match parse_url(&tracker_url) {
            Ok(url) => url,
            Err(err) => {
                return Err(ValidationError::InvalidTrackerUrl {
                    source: Located(err).into(),
                })
            }
        };

        if tracker_mode.is_close() && (tracker_url.scheme() != "http" && tracker_url.scheme() != "https") {
            return Err(ValidationError::UdpTrackersInPrivateModeNotSupported);
        }

        Ok(())
    }
```

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/pull/790

da2ce7:
closed by #401

PR #561, opened 2023-12-31
Cleanup Service Initialization

### LINKED ISSUE #560: Cleanup Service Initialization Code
Currently Torrust Tracker uses various different code paths for initializing the services, instead it should use a consistent code-path for starting services.

da2ce7:
Extracted from: 
- #557

Closes #560


src/bootstrap/jobs/http_tracker.rs
```diff
@@ -7,88 +7,79 @@
...
+        let cfg = Arc::new(ephemeral_mode_public());
+        let config = &cfg.http_trackers[0];
+        let tracker = initialize_with_configuration(&cfg);
+        let version = Version::V1;
 
-    // Wait until the HTTP tracker server job is running
-    match rx.await {
-        Ok(_msg) => info!("Torrust HTTP tracker server started"),
-        Err(e) => panic!("the HTTP tracker server was dropped: {e}"),
+        start_job(config, tracker, version).await.expect("some https server");
```
josecelano: Hi @da2ce7, I think I would add an explicit assert:

```rust
    #[tokio::test]
    async fn it_should_start_http_tracker() {
        let cfg = Arc::new(ephemeral_mode_public());
        let config = &cfg.http_api;
        let tracker = initialize_with_configuration(&cfg);
        let version = Version::V1;

        let opt_handle = start_job(config, tracker, version).await;

        assert!(opt_handle.is_some());
    }
```

Becuase you don't know if that this line:

`start_job(config, tracker, version).await.expect("some https server");`

is only the ACT part or the ACT and ASSERT of the test. I had the feeling the assert was missing.

If you want to use `expect` becuase you want to add a message you could do this:

```rust
    #[tokio::test]
    async fn it_should_start_http_tracker() {
        let cfg = Arc::new(ephemeral_mode_public());
        let config = &cfg.http_api;
        let tracker = initialize_with_configuration(&cfg);
        let version = Version::V1;

        let opt_handle = start_job(config, tracker, version).await;

        opt_handle.expect("http server job should be started");
    }
```

https://doc.rust-lang.org/nightly/std/error/index.html#common-message-styles

src/bootstrap/jobs/mod.rs
```diff
@@ -11,3 +11,86 @@ pub mod http_tracker;
...
+        } else {
+            Some(Err(Error::MissingTlsConfig {
+                location: Location::caller(),
+            }))
+        }
+    } else {
+        info!("tls not enabled");
+        None
+    }
+}
```
josecelano: Hi @da2ce7 I think there are too many nested conditions.In general I prefer early return on errors:

```rust
pub async fn make_rust_tls(enabled: bool, cert: &Option<String>, key: &Option<String>) -> Option<Result<RustlsConfig, Error>> {
    if !enabled {
        info!("tls not enabled");
        return None;
    }

    match (cert, key) {
        (Some(cert), Some(key)) => {
            info!("Using HTTPS: cert path: {cert}, key path: {key}.");

            Some(
                RustlsConfig::from_pem_file(cert, key)
                    .await
                    .map_err(|err| Error::BadTlsConfig {
                        source: (Arc::new(err) as DynError).into(),
                    }),
            )
        }
        _ => Some(Err(Error::MissingTlsConfig {
            location: Location::caller(),
        })),
    }
}
```

I would like to also return if the `cert` or the `key` are missing but then you would need an `unwrap`:

```rust
pub async fn make_rust_tls(enabled: bool, cert: &Option<String>, key: &Option<String>) -> Option<Result<RustlsConfig, Error>> {
    if !enabled {
        info!("TLS not enabled");
        return None;
    }

    if cert.is_none() || key.is_none() {
        return Some(Err(Error::MissingTlsConfig {
            location: Location::caller(),
        }));
    }

    let cert = cert.as_ref().unwrap(); // Safe to unwrap as we checked above
    let key = key.as_ref().unwrap(); // Safe to unwrap as we checked above

    info!("Using HTTPS: cert path: {cert}, key path: {key}.");

    RustlsConfig::from_pem_file(cert, key)
        .await
        .map_err(|err| Error::BadTlsConfig {
            source: Arc::new(err).into(),
        })
        .into()
}
```

Besides, this makes me wonder if we are doing this check in the right place. Maybe we should not even accept the invalid configuration. In the configuration, we have:

```rust
pub struct HttpTracker {
    /// Weather the HTTP tracker is enabled or not.
    pub enabled: bool,
    /// The address the tracker will bind to.
    /// The format is `ip:port`, for example `0.0.0.0:6969`. If you want to
    /// listen to all interfaces, use `0.0.0.0`. If you want the operating
    /// system to choose a random port, use port `0`.
    pub bind_address: String,
    /// Weather the HTTP tracker will use SSL or not.
    pub ssl_enabled: bool,
    /// Path to the SSL certificate file. Only used if `ssl_enabled` is `true`.
    #[serde_as(as = "NoneAsEmptyString")]
    pub ssl_cert_path: Option<String>,
    /// Path to the SSL key file. Only used if `ssl_enabled` is `true`.
    #[serde_as(as = "NoneAsEmptyString")]
```

And there is a data structure that is repeated in the API, and HTTP trackers:

```rust
    pub ssl_enabled: bool,
    /// Path to the SSL certificate file. Only used if `ssl_enabled` is `true`.
    #[serde_as(as = "NoneAsEmptyString")]
    pub ssl_cert_path: Option<String>,
    /// Path to the SSL key file. Only used if `ssl_enabled` is `true`.
    #[serde_as(as = "NoneAsEmptyString")]
```

Maybe we should extract a new struct that enforces the configuration rule:

Rule: if SSL is enabled SSL cert and SSL Key are mandatory:

```rust
pub struct TLSConfig {
    pub ssl_enabled: bool,
    /// Path to the SSL certificate file. Only used if `ssl_enabled` is `true`.
    pub ssl_cert_path: Option<String>,
    /// Path to the SSL key file. Only used if `ssl_enabled` is `true`.
    pub ssl_key_path: Option<String>,
}

impl TLSConfig {
    pub fn new(ssl_enabled: bool, ssl_cert_path: Option<String>, ssl_key_path: Option<String>) -> Result<Self, &'static str> {
        if ssl_enabled && (ssl_cert_path.is_none() || ssl_key_path.is_none()) {
            Err("Both cert and key must be provided when SSL is enabled")
        } else {
            Ok(TLSConfig {
                ssl_enabled,
                ssl_cert_path,
                ssl_key_path,
            })
        }
    }
}
```

That would make parsing configuration harder but we can keep the flat structure for the toml file and add a new method to `HttpTracker` to return the `TLSConfig`.

Alternatively, we could add a validate method to all configuration structs. I guess the configuration structs are only DTOs which contain the same information as the toml file, but that does not mean the configuration is valid. Maybe we should have two types of structs for configuration:

- The DTO: the way we transfer configuration to JSON, toml, etc.
- The actual domain config structs that are validated. You cannot instantiate invalid configuration (which is possible right now).

I think we should do something similar to what Nginx does:

We could statically validate a configuration file without running it. With the current solution, we need to run the tracker with a given configuration to know if the configuration is valid.

The only validation we are doing now is the parsing process. WE try to deserialize the toml file into Rust structs with certain types. But for example, socket addresses are of type `String`:

```rust
pub bind_address: String,
```

That confirms that a second level of validation and a higher-level of domain structs 
for configuration could be a nice thing.

- Higer-level: domain configuration structs. Reach types and valid config with invariants.
- Lower-level: DTOs for configuration to be able to represent the configuration in different formats: toml, JSON, yaml.

src/bootstrap/jobs/tracker_apis.rs
```diff
@@ -49,51 +51,56 @@ pub struct ApiServerJobStarted();
...
+        let cfg = Arc::new(ephemeral_mode_public());
+        let config = &cfg.http_api;
+        let tracker = initialize_with_configuration(&cfg);
+        let version = Version::V1;
 
-    // Wait until the APIs server job is running
-    match rx.await {
-        Ok(_msg) => info!("Torrust APIs server started"),
-        Err(e) => panic!("the API server was dropped: {e}"),
+        start_job(config, tracker, version).await.expect("some https server");
```
josecelano: Hi @da2ce7 there is a typo here:

```rust
    #[tokio::test]
    async fn it_should_start_the_tracker_api() {
        let cfg = Arc::new(ephemeral_mode_public());
        let config = &cfg.http_api;
        let tracker = initialize_with_configuration(&cfg);
        let version = Version::V1;

        start_job(config, tracker, version).await.expect("some tracker API server");
```

It should be the API, not the HTTP tracker.

src/bootstrap/jobs/udp_tracker.rs
```diff
@@ -8,30 +8,34 @@
...
+/// It will panic if the task did not finish successfully.
 #[must_use]
-pub fn start_job(config: &UdpTracker, tracker: Arc<core::Tracker>) -> JoinHandle<()> {
-    let bind_addr = config.bind_address.clone();
+pub async fn start_job(config: &UdpTracker, tracker: Arc<core::Tracker>) -> JoinHandle<()> {
+    let bind_to = config
+        .bind_address
+        .parse::<std::net::SocketAddr>()
+        .expect("Tracker API bind_address invalid.");
+
```
josecelano: Hi @da2ce7, this relates to my previous comment. Maybe at this point `UdpTracker` configuration should contain a valid `SocketAddr`.

josecelano (dismissed):
Good work @da2ce7 I've only added a couple of comments. Regarding my comments on configuration it's something to consider for future bif refactors in configuration.

src/bootstrap/jobs/mod.rs
da2ce7: @josecelano In this pull request I have moved the `RustlsConfig` tls config up the chain. However to rework the tls config handling is not the focus of this pull request.

Good ideas for having validated structures for the config. I likewise think it is best to validate as much of the config as early as possible.

src/bootstrap/jobs/http_tracker.rs
da2ce7: @josecelano In this pull request I was too lazy to expand the error system for the bootstrap. So I just kept the behavior of panicking. (Maybe it is a little neater now).

src/bootstrap/jobs/tracker_apis.rs
da2ce7: @josecelano thanks, and errors like it have also been fixed.

src/bootstrap/jobs/udp_tracker.rs
da2ce7: @josecelano fixed.

src/bootstrap/jobs/http_tracker.rs
josecelano: Hi @da2ce7 my option also panics, it was just to make the assert more explicit.

josecelano:
ACK a12fdeefed89f1995329b5ad19ca7b63d1001a57

da2ce7:
ACK 13140f60dbd461fafb2ce234c6c968d7f23052ec

PR #563, opened 2024-01-01
Cleanup Announce Response

da2ce7:
Ready For Review.


josecelano:
Hi @da2ce7 I have to do a more formal review but there are some smells (first impression):

1. Empty traits:

src/servers/http/v1/responses/announce.rs
```diff
@@ -2,62 +2,145 @@
...
-/// Normal (non compact) `announce` response.
+/// Trait that defines the Announce Response Format
+pub trait Response: From<AnnounceData> + axum::response::IntoResponse {
+    /// Returns the Body of the Announce Response
+    ///
+    /// # Errors
+    ///
+    /// If unable to generate the response, it will return an error.
+    fn body(&self) -> Result<Vec<u8>, responses::error::Error>;
+}
```
josecelano: Hi @da2ce7 I like this trait. A response is something that can bi built from the `AnnounceData` and converted into an Axum response.

josecelano (commented):
Hi @da2ce7, in general I'm OK with this PR because there was a duplication which had to be removed, however, I have the feeling that  empty traits and structs should be avoided (if possible).

For me, traits should represent behavior, although I've seen this pattern: [Marker Trait](https://stackoverflow.com/questions/76445707/what-is-a-marker-trait-in-rust).

Same for structs. Empty implementations look like inheritance.

In general, it seems there is a "hidden" enum in this implementation.

If the goal was just to remove duplication I think we could do it without adding empty traits and empty implementations. I've explored that path a little bit [here](https://github.com/torrust/torrust-tracker/pull/585).

Additionally, I think both responses are the same except for the peer list. I think we should take advantage of that property. I think we don't need to mark the parent struct as normal or compact with a different type than the one we use for the peer list. What makes the response compact is the peer list, if I'm not wrong.

The alternative I've proposed It's also aligned with other structs that already exist like AnnouncePolicy and SwarmStats. So we could simplify it even more.

da2ce7:
@josecelano 

I used 

``` rust
pub struct Announce<E, P>
where
    E: Encoding<P>,
    P: PeerEncoding,
```
An `Announce` is not typed based upon the `PeerEncoding`, but a general `Encoding`.

I would have liked to have the signature something like:

``` rust
pub struct Announce<E>
where
    E: Encoding,
```
However, I wanted to have the types being auto-detected based upon supplying form of the Encoding of the peers:
https://github.com/torrust/torrust-tracker/pull/563/files#diff-a73aff059d2186689d35963b70948e4c9e98739953692f7d3d817ecc7e2884c4R481-R494

This is where some of the trait complexity comes from in my implementation.

josecelano:
Hi @da2ce7, there is something I don't like about this PR and [mine](https://github.com/torrust/torrust-tracker/pull/585). When I wrote part of that code I wanted to have a clear representation of what an announce encoded response contains (normal and compact) so the Rust structs have the same structure as the bencoded format.

Notice that we already have an [AnnounceData](https://github.com/torrust/torrust-tracker/blob/5f6eed7d7fce465bfade63eb79c3b74661f0932b/src/core/mod.rs#L502-L517) struct which is a generic struct holding the data needed by any announce response format.

My idea was to have DTOs that could be the exact representation of the encoded data. After removing the duplicity it's not clear anymore what the response contains. You have to read the serialization function.

I thought it was a good idea to have a direct mapping between:

Rust struct <-> Bencoded response <-> Json representation of the bencoded format (like in [here](https://chocobo1.github.io/bencode_online/))

In fact, maybe we have the need to remove duplication because of the comments. If we consider these structures just DTOs the duplication would not be bad, in fact, it could be good, since those structures (even having the same fields) represent different concepts defined in different BEPs. For example, if in the future, a new BEP is introduced and they add a new field to the compact response but not to the normal response we would need to undo a lot of what it's done here.

And regarding my implementation there is still some duplicity, but I think it could be removed making some functions generic. I just wanted to explore a little bit that option. But as I mentioned I've started feeling I'm removing duplication but introducing "domain" entities and making less clear what the encoded response contains.

da2ce7:
@josecelano

I have reworked the code to try and better take into account you concerns. Now there is are `Normal` and `Compact` structs that contain their own prepared data that is ready for serialization.

It is possible to add new forms of serialization, or to remove or add fields without modifying the other structures.

src/core/databases/mod.rs
```diff
@@ -134,7 +134,7 @@ pub trait Database: Sync + Send {
     /// # Errors
     ///
     /// Will return `Err` if unable to save.
-    async fn save_persistent_torrent(&self, info_hash: &InfoHash, completed: u32) -> Result<(), Error>;
+    async fn save_persistent_torrent(&self, info_hash: &InfoHash, downloaded: u32) -> Result<(), Error>;
```
josecelano, line 137: Hi @da2ce7 I see you have renamed stats fields.

- `SwarmMetadata`: https://github.com/torrust/torrust-tracker/blob/develop/src/core/torrent/mod.rs#L55-L67
- `SwarmStats`: https://github.com/torrust/torrust-tracker/blob/develop/src/core/torrent/mod.rs#L76-L87

`SwarmMetadata` contains the names defined in the [BRP 48](https://www.bittorrent.org/beps/bep_0048.html). And `SwarmStats` the names we were using in the Tracker.

src/core/mod.rs
```diff
@@ -500,20 +501,14 @@ pub struct TorrentsMetrics {
...
-    pub swarm_stats: SwarmStats,
-    /// The interval in seconds that the client should wait between sending
-    /// regular requests to the tracker.
-    /// Refer to [`announce_interval`](torrust_tracker_configuration::Configuration::announce_interval).
-    pub interval: u32,
-    /// The minimum announce interval in seconds that the client should wait.
-    /// Refer to [`min_announce_interval`](torrust_tracker_configuration::Configuration::min_announce_interval).
-    pub interval_min: u32,
+    pub stats: SwarmStats,
+    pub policy: AnnouncePolicy,
```
josecelano, line 511: 👍🏼

src/core/torrent/mod.rs
josecelano, line 1: I like removing this duplication but we are using the other names in many places. Maybe we could use methods with the other names instead of the comment.

```rust
pub struct SwarmMetadata {
    /// The number of peers that have ever completed downloading
    pub downloaded: u32,
    /// The number of active peers that have completed downloading (seeders)
    pub complete: u32,
    /// The number of active peers that have not completed downloading (leechers)
    pub incomplete: u32,
}

impl SwarmMetadata {
    /// Alias for the
    fn seeders(&self) -> u32 {
       self.complete
    }
}
```

But I don't like how similar are `complete` (seeders) and `completed` (downloaded).

src/servers/http/v1/responses/announce.rs
josecelano, line 1: Awesome! @da2ce7 I really like this version.

josecelano (dismissed):
Hi @da2ce7 I like this version. I think it's self-explanatory and a good documentation for BEPs. It would be very easy to see what the response is returning and also make changes if needed.

There is only one more thing that I could do. Maybe I would convert the `announce.rs` into a folder with three files:

- `mod.rs`
- `normal.rs`
- `compact.rs`

That would make even easier to read only one response type, I guess.

josecelano:
ACK 7bf84f2f7a4852091a33f0c98776525c56ffb30b

src/core/torrent/mod.rs
da2ce7: I don't like this. I would remove the `SwarmStats`, and rename all the variables to match the `SwarmMetadata` format... but that can happen another time.

josecelano: Hi @WarmBeer I'm not sure. I think it's a good thing to follow BEP names but on the other hand "seeders" and "leechers" are also widely adopted, but apparently not in BEPs.

And "downloaded" and "complete" could be confusing, you do not know by the name which one is still active.

When I was thinking about this I decided that maybe the best option would be to use "seeders/leechers" wording in the whole app except for the scrape response.

da2ce7:
@josecelano Ready for Merge

josecelano:
ACK 0e6fe3309b65a674e6bcb968dde4b2e59af536d2

PR #568, opened 2024-01-02
Refactor: split global consts for limits

### LINKED ISSUE #262: `MAX_SCRAPE_TORRENTS` (74) limit applied to returned peer list in announce request?
There is a `MAX_SCRAPE_TORRENTS` constant declared [here](https://github.com/torrust/torrust-tracker/blob/develop/src/shared/bit_torrent/common.rs#L4).

The [BEP 15. UDP Tracker Protocol for `BitTorrent`](https://www.bittorrent.org/beps/bep_0015.html)
defines this limit:

"Up to about 74 torrents can be scraped at once. A full scrape can't be done with this protocol."

The [BEP 48. Tracker Protocol Extension: Scrape](https://www.bittorrent.org/beps/bep_0048.html)
does not explicitly mention this limit, but the limit is being used for
both the UDP and HTTP trackers since it's applied at the domain level:

- [Here](https://github.com/torrust/torrust-tracker/blob/develop/src/tracker/torrent.rs#L129)
- And [here](https://github.com/torrust/torrust-tracker/blob/develop/src/tracker/torrent.rs#L129) for limiting the number of peers in the `announce` response.

@WarmBeer is this limit defined in any BEP or just a Torrust Tracker limit?

I would use a different const for the peer list limit. It's confusing.

josecelano:
Current const: `crate::shared::bit_torrent::common::MAX_SCRAPE_TORRENTS`

`MAX_SCRAPE_TORRENTS` is the limit only for the number of torrents in a `scrape`request.

New const: `crate::core::TORRENT_PEERS_LIMIT`

`TORRENT_PEERS_LIMIT` is now the limit for the number of peers in an announce request (UDP and HTTP tracker).

Besides, the endpoint to get the torrent details in the API does not limit the number of peers for the torrent. So the API returns all peers in the tracker. This could lead to performance issues and we might need to paginate results, but the API should either return all peers and paginate them in a new endpoint.

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -458,6 +458,9 @@ use crate::core::databases::Database;
 use crate::core::torrent::{SwarmMetadata, SwarmStats};
 use crate::shared::bit_torrent::info_hash::InfoHash;
 
+/// The maximum number of returned peers for a torrent.
+pub const TORRENT_PEERS_LIMIT: usize = 74;
+
 /// The domain layer tracker service.
 ///
 /// Its main responsibility is to handle the `announce` and `scrape` requests.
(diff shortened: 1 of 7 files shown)


josecelano:
ACK 46e67a86c7e2861cc924713df0e049468b434817

PR #579, opened 2024-01-03
Preload info hashes in wrk benchmark

mickvandijke:
Preload info hashes in wrk benchmark instead of generating them on the fly.


mickvandijke:
ACK fc9bd77e236cad3106c90aeed708760f667ea621

Issue #567, opened 2024-01-02
Add a configurable memory limit to the torrent repository


mickvandijke:
Currently the torrent repository is not bound by an upper limit of torrents or memory usage. When the memory size of the torrent repository exceeds the amount of memory available on the host system, the tracker will [restart](https://github.com/torrust/torrust-tracker/issues/566).

To solve this problem, I want to add a configurable memory limit for the torrent repository. When the torrent repository is full, it should let new torrents replace old low priority torrents. For the first implementation I will base the priority on the time it was last accessed in an announce request.

### Relevant discussions/issues
https://github.com/torrust/torrust-tracker/discussions/499
https://github.com/torrust/torrust-tracker/issues/566

### Needs issues
https://github.com/torrust/torrust-tracker/issues/565

cc @josecelano @da2ce7

josecelano:
Hi, @WarmBeer I think an explicit limit is a nice feature but it would also be good to set a no-limit option so the tracker can consume as much memory as possible without panicking.

mickvandijke:
> Hi, @WarmBeer I think an explicit limit is a nice feature but it would also be good to set a no-limit option so the tracker can consume as much memory as possible without panicking.

Will add a limit of `0` that will allow the torrent repository to grow until there is no more space available.

josecelano:
> > Hi, @WarmBeer I think an explicit limit is a nice feature but it would also be good to set a no-limit option so the tracker can consume as much memory as possible without panicking.
> 
> Will add a limit of `0` that will allow the torrent repository to grow until there is no more space available.

@WarmBeer maybe `-1` it's a better option.

ChatGPT:

> When choosing a value to represent 'no limit' for a configuration option like memory_usage_limit in an application configuration file, it's essential to consider clarity, standard practices, and the potential for misinterpretation. Let's evaluate the options you've provided:
> 
> 0: Typically, '0' implies no allocation or disabling the feature. In the context of memory usage, it might be misunderstood as not allowing the application to use any memory, which could confuse users.
> 
> -1: This is a common convention in many programming and configuration contexts to represent 'unlimited' or 'no limit'. It's widely recognized by developers and system administrators, making it a good choice for clarity and standard practice.
> 
> null: Using 'null' can be a good option, as it explicitly represents the absence of a value. However, it might require additional handling in the code to interpret 'null' as 'no limit' and not as an uninitialized or error state.
> 
[...]

PR #581, opened 2024-01-04
Add DashMap implementation for torrent repository and set as used one

mickvandijke:
DashMap does not support ordering of keys that we use right now for our scrape responses, since DashMap works using multiple inner HashMaps (sharded). 

When I implement the torrent repository limit, I plan on adding a new HashMap that keeps track of the storage priority of torrents. We could also use the ordering of that HashMap for our torrents and scrape responses.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -835,6 +835,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mickvandijke:
I will include https://github.com/torrust/torrust-tracker/issues/567 as a part of this PR as well to fix any related failing tests. Especially for the ordering of torrents in the API response.

josecelano:
I guess this was superseded by https://github.com/torrust/torrust-tracker/pull/645, right @mickvandijke?

Issue #620, opened 2024-01-16
Add E2E tests


josecelano:
We are not covering the bootstrapping of the app because we use a lot of "spwan". It would be convenient to write some E2E tests for basic features like the following:

- Call the healthcheck endpoints
- Make an announce request using the HTTP tracker
- Make an announce request using the UDP tracker

We can use the docker image like in the [Index](https://github.com/torrust/torrust-index/blob/develop/contrib/dev-tools/container/e2e/sqlite/run-e2e-tests.sh).

### Subtasks

- [x] https://github.com/torrust/torrust-tracker/issues/634
- [x] https://github.com/torrust/torrust-tracker/issues/647
- [x] https://github.com/torrust/torrust-tracker/issues/637

josecelano:
We now run all checks:

- UDP tracker: announce and scrape requests.
- HTTP tracker: announce and scrape requests.
- HealthCheck: request to the health checker endpoint.

PR #623, opened 2024-01-17
Extract Config from Tracker and Heath Check

### LINKED ISSUE #524: Fix container `HEALTHCHECK` when port `0` is used in confguration
### Context

We have a `HEALTHCHECK` instruction in the `Containerfile`:

```dockerfile
HEALTHCHECK --interval=5s --timeout=5s --start-period=3s --retries=3 \  
  CMD /usr/bin/http_health_check http://localhost:${HEALTH_CHECK_API_PORT}/health_check \
    || exit 1
```

The implementation is [here](https://github.com/torrust/torrust-tracker/blob/develop/src/servers/health_check_api/handlers.rs#L23C1-L37C2). It checks that all services are running:

- API
- 1 or more HTTP trackers
- 1 or more UDP trackers

For the API, for example, it tries to connect to the API healthcheck endpoint.

```rust
async fn api_health_check(config: &HttpApi) -> Option<Json<Report>> {
    // todo: when port 0 is specified in the configuration get the port from the
    // running service, after starting it as we do for testing with ephemeral
    // configurations.

    if config.enabled {
        let addr: SocketAddr = config.bind_address.parse().expect("invalid socket address for API");

        if addr.port() != UNKNOWN_PORT {
            let health_check_url = format!("http://{addr}/health_check");

            if !get_req_is_ok(&health_check_url).await {
                return Some(responses::error(format!(
                    "API is not healthy. Health check endpoint: {health_check_url}"
                )));
            }
        }
    }

    None
}
```

In the API configuration (`Config.toml`):

```toml
[http_api]
bind_address = "127.0.0.1:1212"
enabled = true
#...
```

you can specify the port = like this:

```toml
[http_api]
bind_address = "127.0.0.1:0"
enabled = true
#...
```

In that case, the application will assign a free port. 

### Problem

The healthcheck endpoint relies on static configuration. If you use port 0 the healthcheck skippies that service. We need to pass the list of actual ports to the "healthchecker" instead of the configuration.

### Solution

When we run the services we have to send back the port used to the main app. And the main app has to inject that configuration (list of services and ports) to the "healthchecker".

### Relates To

- https://github.com/torrust/torrust-tracker/issues/274
- https://github.com/torrust/torrust-tracker/issues/213
- https://github.com/torrust/torrust-tracker/issues/186

da2ce7:
This is the first three commits from #557 

It includes an significant improvement of the Heath Check logic, as it now takes the active port from the binding, instead of the configuration.


packages/configuration/src/lib.rs
```diff
@@ -804,7 +798,7 @@ mod tests {
     fn http_api_configuration_should_check_if_it_contains_a_token() {
         let configuration = Configuration::default();
 
-        assert!(configuration.http_api.contains_token("MyAccessToken"));
-        assert!(!configuration.http_api.contains_token("NonExistingToken"));
+        assert!(configuration.http_api.access_tokens.values().any(|t| t == "MyAccessToken"));
+        assert!(!configuration.http_api.access_tokens.values().any(|t| t == "NonExistingToken"));
```
josecelano, line 802: Hi @da2ce7 I think I would keep the `contains` abstraction:

```rust
pub struct AccessTokens(HashMap<String, String>)

impl AccessTokens {
    /// Checks if the given token is one of the tokens.
    #[must_use]
    pub fn contains(&self, token: &str) -> bool {
        self.0.values().any(|t| t == "MyAccessToken")
    }
}

// ...

assert!(configuration.http_api.access_tokens.contains("MyAccessToken"));
assert!(!configuration.http_api.access_tokens.contains("NonExistingToken"));

```

But It's not important. Maybe it's because I don't know the HashMap API or because I think it's easy to read or because it's against the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

packages/test-helpers/src/configuration.rs
```diff
@@ -156,3 +156,15 @@ pub fn ephemeral_with_no_services() -> Configuration {
...
 }
+
+/// Ephemeral with only udp
+#[must_use]
+pub fn ephemeral_udp_only() -> Configuration {
+    let mut cfg = ephemeral();
+
+    cfg.http_api.enabled = false;
+    cfg.http_trackers[0].enabled = false;
+    cfg.udp_trackers[0].enabled = false;
```
josecelano: Hi @da2ce7 should this line `cfg.udp_trackers[0].enabled = false;` be `cfg.udp_trackers[0].enabled = true;`?

src/app.rs
```diff
@@ -21,24 +21,106 @@
...
+
+/// A [`Listing`] is provided to the [`Registar`] for registration.
+#[derive(Clone, Debug, Constructor)]
+pub struct Listing {
+    binding: SocketAddr,
+    check_fn: FnCheck,
+}
+
+/// A [`RegistrationForm`] is completed with a [`Listing`] and then is sent.
+pub type RegistrationForm = tokio::sync::oneshot::Sender<Listing>;
```
josecelano: hi @da2ce7 I'm a little bit concerned about the name `RegistrationForm`. I think it's used very often for signups. We could call it:

- `ServiceRegistrationForm`

By the way, the main tracker app runs different services:

- Tracker API
- HTTP tracker
- UDP tracker

I've been calling them services.

On the other hand, why did you use the name `Listing`?

image

Maybe `RegistryEntry` would be more explicit. For me `listing` is more a list than an entry on a list, but maybe because I'm not a native English speaker.

src/app.rs
```diff
@@ -21,24 +21,106 @@
...
+/// Returned by the service check.
+#[derive(Debug, Constructor)]
+pub struct CheckJob {
+    pub binding: SocketAddr,
+    pub info: String,
+    pub job: JoinHandle<Result<String, String>>,
+}
+
+/// The function type that is called to check a service.
+pub type FnCheck = fn(&SocketAddr) -> CheckJob;
```
josecelano: Hi @da2ce7 rename `FnCheck` to `FnHealthCheck`?

src/app.rs
```diff
@@ -21,24 +21,106 @@
...
 use crate::bootstrap::jobs::{health_check_api, http_tracker, torrent_cleanup, tracker_apis, udp_tracker};
 use crate::{core, servers};
 
+/// Returned by the service check.
+#[derive(Debug, Constructor)]
+pub struct CheckJob {
+    pub binding: SocketAddr,
+    pub info: String,
+    pub job: JoinHandle<Result<String, String>>,
+}
```
josecelano: hi @da2ce7 rename to `HealthCheckResult`?

josecelano (dismissed):
Wonderful job @da2ce7! I've only added some comments for things that were "harder" to read for me. I mean, I had to read the details to understand what they do, but only in a few cases and it was fast to get on track. So it's perfectly mergeable.

josecelano:
ACK f073af89bc5d94d59c382a2d0f089c425215a263

packages/configuration/src/lib.rs
da2ce7: @josecelano this is using the standard iterator api, I don't think that it is surprising for any map to provide an iterator.

src/app.rs
da2ce7: Done :smile:

src/app.rs
da2ce7: Done :smile:

src/app.rs
da2ce7: Done :smile:

packages/test-helpers/src/configuration.rs
da2ce7: Fixed :+1:

josecelano:
ACK 3f0dcea464bb4fd7cf7424a02dcc9f9329295caf

josecelano:
Hi @da2ce7 after merging this PR there are [some new problems running the services](https://github.com/torrust/torrust-tracker/issues/626).

I want to [add E2E tests](https://github.com/torrust/torrust-tracker/issues/620) to avoid this type of regression.

PR #635, opened 2024-01-19
Reset default configuration for development

### LINKED ISSUE #590: Tracker API enabled in default configuration
When I implemented the Health Check API I accidentally enabled the tracker API for the development container configuration in this commit e1a45a2.

This line: https://github.com/torrust/torrust-tracker/blob/develop/share/default/config/tracker.development.sqlite3.toml#L28

It's also enabled in the other container configurations. If I'm not wrong initially all services were disabled so the admin has to enable them explicitly.

Should we disable the API by default in all configurations?

josecelano:
- Using wildcard IPs for UDP and HTTP tracker.
- Revert port 0 (unintentionally changed). Use predefined ports.

### DIFF
--- a/share/default/config/tracker.development.sqlite3.toml
+++ b/share/default/config/tracker.development.sqlite3.toml
@@ -13,18 +13,18 @@ remove_peerless_torrents = true
 tracker_usage_statistics = true
 
 [[udp_trackers]]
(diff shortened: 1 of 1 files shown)


josecelano:
ACK 16224da08318361ea3b53aa7766ae2832073a189

PR #643, opened 2024-01-23
E2E tests runner

### LINKED ISSUE #634: E2E tests: scaffolding
Parent issue: https://github.com/torrust/torrust-tracker/issues/620

Scaffolding for adding E2E tests.

We can use a shared docker environment like in the index. See https://github.com/torrust/torrust-index/blob/develop/contrib/dev-tools/container/e2e/sqlite/run-e2e-tests.sh

josecelano:
A new binary to run E2E tests:

- [x] Build the docker image.
- [x] Run the docker image.
- [x] Wait until the container is healthy.
- [x] Parse logs to get running services.
- [x] Build config file for the tracker_checker.
- [x] Run the tracker_checker.
- [x] Stop the container.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -110,3 +110,32 @@ jobs:
       - id: test
         name: Run Unit Tests
         run: cargo test --tests --benches --examples --workspace --all-targets --all-features
+
+  e2e:
+    name: E2E
+    runs-on: ubuntu-latest
+    needs: unit
+
+    strategy:
+      matrix:
+        toolchain: [nightly]
+
+    steps:
+      - id: setup
+        name: Setup Toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          components: llvm-tools-preview
+
+      - id: cache
+        name: Enable Job Cache
+        uses: Swatinem/rust-cache@v2
+
+      - id: checkout
+        name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - id: test
+        name: Run E2E Tests
+        run: cargo run --bin e2e_tests_runner ./share/default/config/tracker.e2e.container.sqlite3.toml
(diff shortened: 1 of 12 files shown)


josecelano:
Hi @da2ce7 I'm also experimenting with using Rust instead of Bash for the CI scripting. We have discussed in the past that it would be nice to write those scripts in Rust too. I think in this case it's a perfect fit.

josecelano:
Hi @da2ce7 I'm using Docker for the E2E test environment like in the Index. In the Index we needed because we wanted to also run the Tracker. Here we don't need it, we could simply run `cargo run` but I'm using it to test also containerization. This way E2E tests will cover everything. The problem is the container build is made again in the testing workflow and it takes 10 minutes.

Another problem with Docker is, we can not use port 0 to assign a free port becuase we need to publish the port when we run docker.

So, I'm considering not using Docker. What do you think?

josecelano:
Hi @da2ce7, some comments:

1. Independent crate for each CI script?

Maybe this script should be an independent crate in the `contrib/dev-tools` dir.

It's like [this script in Bash](https://github.com/torrust/torrust-index/blob/develop/contrib/dev-tools/container/e2e/sqlite/run-e2e-tests.sh) but in Rust.

Rigth now, there is no coupling between the main crate and this binary. And I don't like having  CI/Testing code in the production package.

2. Use docker?

See https://github.com/torrust/torrust-tracker/pull/643#issuecomment-1907576015

3. Use `test-helpers` package to build the configuration for the tracker?

I'm using a file in `share/default/config`. We could use `test-helpers` or `configuration` package. However I prefer to do it similar to what I would do it manually to test the tracker.

4. Check that there is no panic in the logs

I could add one more test. I could check that there are no "panics" in the logs. That was one of the reasons I started adding E2E tests. There were some panics in the bootstrapping of the services that made the halt signal not work. But the servers were running anyway, so these panics were not detected by tests.

5. Use an env var to pass conf to the `tracker_checker`

That way we don't even need the temp dir. We could also remove the file at the end and do all in the root folder.

josecelano:
ACK ec13fb41cfef0d68af67b44aee8c83696754a519

josecelano:
With an independent crate for CI scripts, we need to run their tests in new workflows. Maybe that's too complex. We could use one single crate for all the scripts with one binary per script.

Issue #566, opened 2024-01-02
Tracker restarted when memory is full


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/567

We are running the live demo in a droplet with 1GB of RAM.

The tracker domain is https://tracker.torrust-demo.com/health_check.

On the 29th of December I changed the tracker configuration from

```toml
remove_peerless_torrents = true
```

To:

```toml
remove_peerless_torrents = false
```

With this change, the tracker does not clean up torrents without peers. That makes the data structure containing the torrent data grow indefinitely.

That makes the process to run out of memory every 5 hours approx:

![image](https://github.com/torrust/torrust-tracker/assets/58816/11578229-656e-4041-9308-a0919741068f)

I guess, the Tracker container crashes and it's [restarted again](https://github.com/torrust/torrust-compose/blob/main/droplet/compose.yaml#L97), otherwise I suppose the process would simply die.

We should limit the memory usage. This could happen even with the option `remove_peerless_torrents` enabled.

cc @WarmBeer @da2ce7 

### Last 24 hours

![torrust-demo-DigitalOcean-24-hours png](https://github.com/torrust/torrust-tracker/assets/58816/24323fe6-a4d9-4899-a3b8-9be28fc8a1c5)

### Last 7 days

![torrust-demo-DigitalOcean-7-days](https://github.com/torrust/torrust-tracker/assets/58816/aae96262-0bcf-435c-99a6-875edf1db42d)

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/567

@WarmBeer is working on [it](https://github.com/torrust/torrust-tracker/issues/567).

josecelano:
Hi @WarmBeer I've been thinking about this problem. I want to share some thoughts.

### Is the tracker actually crashing?

The tracker was restarted on the demo environment but I suppose it was because of the container healthcheck. we are using this `HEALTHCHECK` instruction:

```dockerfile
HEALTHCHECK --interval=5s --timeout=5s --start-period=3s --retries=3 \  
  CMD /usr/bin/http_health_check http://localhost:${HEALTH_CHECK_API_PORT}/health_check \
    || exit 1
```

And this compose configuration:

```yml
  tracker:
    image: torrust/tracker:develop
    container_name: tracker
    tty: true
    restart: unless-stopped
    environment:
      - USER_ID=${USER_ID}
[...]

PR #644, opened 2024-01-24
udp: handle udp requests concurrently

### LINKED ISSUE #611: Process UDP requests concurrently
Relates to: https://github.com/torrust/torrust-tracker/issues/565, https://github.com/torrust/torrust-tracker/issues/596
Depends on: https://github.com/torrust/torrust-tracker/issues/565

The current version of the UDP server processes one request at a time.

```rust
let listen = async move {
    debug!(target: "UDP Tracker", "Waiting for packets on socket address: udp://{address} ...");

    loop {
        let socket_clone = socket.clone();

        match socket_clone.recv_from(&mut data).await {
            Ok((valid_bytes, remote_addr)) => {

                debug!(target: "UDP Tracker", "Received {} bytes", payload.len());
                debug!(target: "UDP Tracker", "From: {}", &remote_addr);
                debug!(target: "UDP Tracker", "Payload: {:?}", payload);

                let response_fut = handle_packet(remote_addr, payload, &tracker);

                match tokio::time::timeout(Duration::from_secs(5), response_fut).await {
                    Ok(response) => {
                        Udp::send_response(socket_clone, remote_addr, response).await;
                    }
                    Err(_) => {
                        error!("Timeout occurred while processing the UDP request.");
                    }
                }
            }
            Err(err) => {
                error!("Error reading UDP datagram from socket. Error: {:?}", err);
            }
        }
    }
};
```

For the time being, that's not a bottleneck because we also write sequentially into the Tracker repository. @WarmBeer is working on the issue:

- [x] https://github.com/torrust/torrust-tracker/issues/565

That would allow concurrent writes/reads but to take advantage of that change I think we need to change the UDP server.

One implementation proposal could be using workers. I think @da2ce7 was working on a refactor where he was using a pool of active requests handled by spawned tasks (when he was working on the graceful shutdown).

This is the draft ChatGTP implementation:

```rust
// ... [existing imports] ...

impl Udp {
    // ... [other methods] ...

    /// It starts the UDP server instance with graceful shutdown.
    async fn start_with_graceful_shutdown(
        tracker: Arc<Tracker>,
        bind_to: SocketAddr,
        tx_start: Sender<Started>,
        rx_halt: Receiver<Halted>,
    ) -> JoinHandle<()> {
        let socket = Arc::new(UdpSocket::bind(bind_to).await.expect("Could not bind to UDP socket."));
        let address = socket.local_addr().expect("Could not get local_addr from binding.");

        info!(target: "UDP Tracker", "Starting on: udp://{}", address);

        let (tx, rx) = mpsc::channel::<(Vec<u8>, SocketAddr)>(100);
        let rx = Arc::new(Mutex::new(rx));

        let n_workers = 10; // Set the number of workers as needed
        for _ in 0..n_workers {
            let worker_rx = rx.clone();
            let worker_socket = socket.clone();
            let worker_tracker = tracker.clone();
            tokio::spawn(async move {
                worker_process(worker_rx, worker_socket, worker_tracker).await;
            });
        }

        let running = tokio::task::spawn(async move {
            let mut halt = tokio::task::spawn(async move {
                shutdown_signal_with_message(
                    rx_halt,
                    format!("Shutting down UDP server on socket address: udp://{address}"),
                )
                .await;
            });

            loop {
                tokio::select! {
                    Ok(_) = &mut halt => {
                        debug!(target: "UDP Tracker", "Halt signal received, stopping listening for new requests.");
                        break;
                    }
                    result = Udp::receive_packet(&socket) => {
                        if let Ok((payload, remote_addr)) = result {
                            if tx.send((payload, remote_addr)).await.is_err() {
                                error!("Failed to send request to worker");
                            }
                        } else if let Err(e) = result {
                            error!("Error reading UDP datagram from socket. Error: {:?}", e);
                        }
                    }
                }
            }

            debug!(target: "UDP Tracker", "Main listener stopped, waiting for worker tasks to complete.");
        });

        info!(target: "UDP Tracker", "Started on: udp://{}", address);

        running
    }

    async fn receive_packet(socket: &Arc<UdpSocket>) -> std::io::Result<(Vec<u8>, SocketAddr)> {
        let (valid_bytes, remote_addr) = socket.recv_from(&mut data).await?;

        Ok((payload, remote_addr))
    }

    // ... [existing Udp methods] ...
}

async fn worker_process(
    rx: Arc<Mutex<mpsc::Receiver<(Vec<u8>, SocketAddr)>>>,
    socket: Arc<UdpSocket>,
    tracker: Arc<Tracker>,
) {
    while let Some((payload, remote_addr)) = rx.lock().await.recv().await {
        let response_fut = handle_packet(remote_addr, payload, &tracker);

        // Implementing a 5-second timeout for each request
        match tokio::time::timeout(Duration::from_secs(5), response_fut).await {
            Ok(response) => {
                Udp::send_response(socket.clone(), remote_addr, response).await;
            }
            Err(_) => {
                error!("Timeout occurred while processing the UDP request from {}", remote_addr);
                // Optionally, send a timeout response or error message back to the client
                // Udp::send_timeout_response(socket.clone(), remote_addr).await;
            }
        }
    }
}

// ... [existing tests and other code] ...


```

da2ce7:
Extracted from #557 

Upgrades the udp implementation to handle requests concurrently, no timeout is needed as udp is non-blocking.

Hard-coded limit of 50 concurrent requests. With trivial changes can make this limit configurable, however I don't think that it would really be much benefit for it, the main concern is memory usage, and since udp is no blocking each request should return very quickly, could set to 5000 and it should still be fine.


josecelano:
Hi @da2ce7 did you read my [comment about dynamically limiting the number of concurrent requests](https://github.com/torrust/torrust-tracker/issues/566#issuecomment-1900105745)?

da2ce7:
> Hi @da2ce7 did you read my [comment about dynamically limiting the number of concurrent requests](https://github.com/torrust/torrust-tracker/issues/566#issuecomment-1900105745)?

This is not related to udp requests, as udp requests are non-blocking.

da2ce7:
@josecelano I mean, I doubt there is any real improvement above 5 requests at the same time. The limit is the locking speed of our data-structures, and the kernel sending the udp packets.

josecelano:
> @josecelano I mean, I doubt there is any real improvement above 5 requests at the same time. The limit is the locking speed of our data-structures, and the kernel sending the udp packets.

@da2ce7 my point is:

- The main bottleneck is accessing the data in the core tracker service (our data-structures).
- All HTTP and UDP trackers use the same core service.
- The high-level services (HTTP and UDP) make requests to the low-level core service. And they have to wait (lock) until they can access the core data.
- The more pending requests at the core level the more response time (from the core service and from the HTTP and UDP to the final users) and the more memory consumption.

The idea is to reject requests at the high level when they detect the core level is taking too long to respond. This should balance the load and avoid server degradation.

Anyway, that limit could be implemented on top of this PR once is merged. I think it can be easily changed.

src/servers/udp/server.rs
```diff
@@ -209,80 +237,103 @@ impl Udp {
...
+        result: Result<UdpRequest, Box<std::io::Error>>,
+        tracker: Arc<Tracker>,
+        socket: Arc<UdpSocket>,
+    ) -> JoinHandle<()> {
+        // timeout not needed, as udp is non-blocking.
+        tokio::task::spawn(async move {
+            match result {
+                Ok(udp_request) => {
+                    trace!("Received Request from: {}", udp_request.from);
+                    Self::make_response(tracker.clone(), socket.clone(), udp_request).await;
```
josecelano, line 318: Hi @da2ce7  I would rename `make_response` to `handle_request`.

josecelano (approved):
Hi @da2ce7 I think it's a good functionality and good implementation but it's not compatible with dynamically changing the number of active requests since the 50 limit is hardcoded in the type.

I would merge it but if we want to make it dynamic in the future we will need to change this code.

josecelano:
ACK 72c8348559f45f936266823d576f8d8798b32d66

josecelano:
Hi @da2ce7 another concern is whether we should spawn a new thread for each request or it would be better to have a fixed amount of workers each one processing many requests (sequentially).

See https://github.com/torrust/torrust-tracker/issues/611#issue-2083543706

If I'm not wrong this is the approach for some webservers like Nginx and Apache.

Below ChatGPT feedback:

### Spawning a New Thread per Request (up to 50)

**Advantages:**

1. Responsiveness: Each request is immediately assigned its own thread, potentially leading to faster response times as each request is processed in parallel.
2. Simplicity: The logic can be simpler as each thread only handles one request, avoiding the complexities of managing a shared worker pool.
3. Scalability: This approach can handle spikes in requests efficiently, as new threads are created as needed (up to the limit).

**Disadvantages:**

1. Resource Intensive: Threads consume system resources (memory and CPU). Having too many threads can lead to high context-switching overhead and increased memory usage.
2. Scalability Limit: The maximum cap of 50 threads may not be sufficient for high loads, and increasing this limit can strain the system.
3. Lack of Control: It's harder to manage and monitor individual threads as compared to a fixed pool of workers.

### Using a Fixed Number of Worker Threads

**Advantages:**

1. Resource Efficiency: A fixed number of threads can be more efficient in terms of resource usage, as there's a limit to how many threads are running at any given time.
2. Predictability: This model offers more predictable performance, as the system is not continuously creating and destroying threads.
3. Easier to Manage: Managing a fixed pool of threads is typically simpler, especially for monitoring and debugging purposes.
4. Load Balancing: You can implement more sophisticated load balancing strategies among the fixed set of worker threads.

**Disadvantages:**

1. Potential for Bottlenecks: If the number of worker threads is too low, they might become a bottleneck during peak loads.
2. Complexity in Implementation: Managing a pool of worker threads and the distribution of requests to them can be more complex than simply spawning a new thread per request.
3. Less Responsive to Spikes: Unlike the dynamic approach, a fixed number of threads might not handle sudden spikes in traffic as responsively.

### Conclusion

The choice between these two approaches depends on your specific requirements and constraints:

- If your application experiences highly variable loads with significant spikes, dynamically creating threads (up to a limit) might be more beneficial.
- However, if the load is relatively predictable and you want more control over resource usage, a fixed number of worker threads would be more suitable.

In many modern applications, especially those that are I/O-bound like network applications, an asynchronous or event-driven model (such as using an event loop and non-blocking I/O) can often provide a more scalable solution than either of these threading models. This might be worth considering depending on the specifics of your application and its environment.

da2ce7:
@josecelano Tokio has a [light-weight threading model](https://docs.rs/tokio/latest/tokio/runtime/), with a global thread pool that matches the number of execution cores of the system.

While there is some overhead in creating tokio tasks, they are far lighter-weight than system-threads. I think that having a fixed set of worker threads dedicated to udp, while more optimal, would not give the significant benefit above light-weight threads used here.

The "costly" part in my code is the "yield_now()" call, this happens when there is around 50 active tasks, it will temporally pause the handling of new requests, and give an opportunity for the other tasks to finish up... one attack/abuse would be that the task-pool is deliberately exhausted, causing many unnecessarily pauses for the yield...

... however the answer for this is to have a two-level task-pool, where the tasks are grouped by client that created the request:

```text
pool (tasks grouped by client) <- pool (tasks for client)
```

With such a design you would need to connect with many different clients to fill the pool, each client gets its own pool... the client local pool would be blocking if full... but the task grouped by client would yield when full.

josecelano:
> @josecelano Tokio has a [light-weight threading model](https://docs.rs/tokio/latest/tokio/runtime/), with a global thread pool that matches the number of execution cores of the system.
> 
> While there is some overhead in creating tokio tasks, they are far lighter-weight than system-threads. I think that having a fixed set of worker threads dedicated to udp, while more optimal, would not give the significant benefit above light-weight threads used here.
> 
> The "costly" part in my code is the "yield_now()" call, this happens when there is around 50 active tasks, it will temporally pause the handling of new requests, and give an opportunity for the other tasks to finish up... one attack/abuse would be that the task-pool is deliberately exhausted, causing many unnecessarily pauses for the yield...
> 
> ... however the answer for this is to have a two-level task-pool, where the tasks are grouped by client that created the request:
> 
> ```
> pool (tasks grouped by client) <- pool (tasks for client)
> ```
> 
> With such a design you would need to connect with many different clients to fill the pool, each client gets its own pool... the client local pool would be blocking if full... but the task grouped by client would yield when full.

Hi @da2ce7 I did not remember that Tokyo tasks are not equal to threads. Thank you for your explanation. I'm going to merge it, my other concern it's not clear yet. @WarmBeer is working on limiting memory consumption. I have the feeling that we need a global solution for limiting both the CPU and memory consumption. If @WarmBeer solution works we can later limit the CPU consumption to avoid system degradation.

PR #648, opened 2024-01-26
E2E tests: Improve E2E tests runner

### LINKED ISSUE #647: E2E tests: Improve E2E test runner
Parent issue: https://github.com/torrust/torrust-tracker/issues/620

- Detect "panics" in logs
- Make sure the tracker configuration enables all services (UDP tracer, HTTP tracker and tracker API).

WHen you run the E2E tests:

```console
cargo run --bin e2e_tests_runner ./share/default/config/tracker.e2e.container.sqlite3.toml
```

The provided configuration `./share/default/config/tracker.e2e.container.sqlite3.toml` could not enable all services, or there could be a bug starting services. After starting the tracker we should check that all services are running. And at least one service per type (UDP, HTTP, HealtCheck).

We would also check that the logs don't contain any panics after starting the tracker and after stopping it.

josecelano:
- [x] Check that we run all services (UDP, HTTP, HealCheck API).
- [x] Check that the logs don't contain any panics.

### DIFF
--- a/src/bin/tracker_checker.rs
+++ b/src/bin/tracker_checker.rs
@@ -7,5 +7,5 @@ use torrust_tracker::checker::app;
 
 #[tokio::main]
 async fn main() {
-    app::run().await;
+    app::run().await.expect("Some checks fail");
 }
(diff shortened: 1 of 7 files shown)


josecelano:
ACK f18e68cc498fa98ebb2b23b0605dd8ec60dbf579

PR #719, opened 2024-02-27
Add ADR about when to use plural for mod names

### LINKED ISSUE #155: Normalize module names containing multiple objects
Sometimes single-file and folder module names containing more than one public struct are singular and sometimes plural.

For example:

- `src/http/request.rs` contains more than one request type.
- `src/http/response.rs` contains more than one response type.

In general, the plural is the most used option.

The folder name in the plural is better because it contains a collection of that "thing", but sometimes, when you use the namespace, the singular is easier to read, for example:

- `request::Announce` instead of `requests::Announce`.

Another advantage of the plural is you do not get Clippy warnings when a function in the module has a suffix which is the mod name, For example, mod `handlers` and an `xxx_handler` function inside.

I prefer the plural.

josecelano:
Add ADR about when to use plural for mod names.

### DIFF
--- a/docs/adrs/20240227164834_use_plural_for_modules_containing_collections.md
+++ b/docs/adrs/20240227164834_use_plural_for_modules_containing_collections.md
@@ -1,0 +1,35 @@
+# Use plural for modules containing collections of types
+
+## Description
+
+In Rust, the naming conventions for module names (mod names) generally lean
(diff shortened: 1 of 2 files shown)


josecelano:
ACK e77d89f848c4525a8dba7081584e6b14db732c95

PR #740, opened 2024-03-19
dev: bugfix completed download stat

da2ce7:
A peer could repetitively  announce that it is completed, making the download stat higher. This changes the behavior to check if the event for the peer is not already completed.


josecelano (approved):
Good catch!

josecelano:
ACK 439821ca29451a79bb2005ba70a8a5145ee0bcbe

PR #741, opened 2024-03-19
Fix error messages starting UDP tracker

josecelano:
String interpolation was not being doing well.


josecelano:
ACK 3a9c52f96ea8c9250eac786663f7a1a3e7ca6474

PR #778, opened 2024-04-05
Performance optimization: create a new torrent repository using a `SkipMap` instead of a `BTreeMap`

josecelano:
This PR implements a new Torrent Repository replacing the outer BTreeMap for torrents with a [SkipMap](https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/).

### Why

- It is straightforward to implement because the API is very similar. In fact, SkipMap is a replacement for BTreeMap that allows concurrency in adding new torrents.
- One problem with BTreeMap is that you have to lock the entire structure to add a new torrent. SkipMap is a lock-free structure.
- It is a lock-free structure that uses Atomics internally, which is more lightweight than locks.
- Unlike the DashMap implementation, it does not use unsafe code. However:

> NOTICE: Race conditions could be introduced if the implementation was incorrect. See https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/#concurrent-access

### Benchmarking

Running the Aquatic UDP load test gives the same results:

Current best implementation:

```output
  
Requests out: 397287.37/second
Responses in: 357549.15/second
  - Connect responses:  177073.94
  - Announce responses: 176905.36
  - Scrape responses:   3569.85
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 104
  - p99.9: 287
  - p100: 371  
```

SkipMap:

```output
Requests out: 396788.68/second
Responses in: 357105.27/second
  - Connect responses:  176662.91
  - Announce responses: 176863.44
  - Scrape responses:   3578.91
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 105
  - p99.9: 287
  - p100: 351  
```

However, benchmarking the repositories using Criterion shows better results in adding and updating multiple torrents in parallel.

image

image

image

image

### Conclusion

I think we car merge it but I would also continue with the DashMap implementation to compare.

PR #645, opened 2024-01-25
Add a configurable memory limit to the torrent repository and implement dashmap [#567, #565]

mickvandijke:
Closes https://github.com/torrust/torrust-tracker/issues/567
Closes https://github.com/torrust/torrust-tracker/issues/565

PR #784, opened 2024-04-09
Performance optimization: create a new torrent repository using `DashMap`

### LINKED ISSUE #565: Use dashmap for torrent in memory storage instead of the std hashmap
DashMap should be significantly faster in concurrent reads and writes compared to the std HashMap.

https://github.com/xacrimon/dashmap
https://github.com/xacrimon/conc-map-bench

josecelano:
Relates to: 

- https://github.com/torrust/torrust-tracker/pull/778
- https://github.com/torrust/torrust-tracker/issues/567#issuecomment-2018926623

This PR adds a new torrent repository implementation where the outer collection for torrents uses a [DashMap](https://docs.rs/dashmap/latest/dashmap/).

This is something @mickvandijke was working on this [PR](https://github.com/torrust/torrust-tracker/pull/645).

There have been many changes. @da2ce7 has extracted a [package for the repositories](https://github.com/torrust/torrust-tracker/tree/develop/packages/torrent-repository).

This PR adds a new repo using DashMap to the new package. However, it does not implement some of the extra features @mickvandijke added to the other [PR](https://github.com/torrust/torrust-tracker/pull/645). For example, it does not limit memory consumption. None of the other repos have that feature, so I suggest merging this PR and implementing that feature in the future for all repos.

### Why

The current repository used in production is the one using a [SkipMap](https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/struct.SkipMap.html) data structure. DashMap was the first type we considered to allow adding new torrents in parallels.

The implementation with DashMap has not been merged because it does not guarantee the order when you iterate over the torrents. The tracker API returns torrents ordered by InfoHash. To avoid breaking the API that PR would need to add a new data structure (kind of Index) to keep the torrent list ordered.

This implementation helps us run performance tests with this option without spending much time fixing its limitations. It looks like the performance is similar to the SkiMap, so we don't need to use it in production now. We only need to keep it for benchmarking in the future if other versions are better.

### Benchmarking

Running the Aquatic UDP load test, the DashMap looks slightly better.

SkipMap:

```output
Requests out: 396788.68/second
Responses in: 357105.27/second
  - Connect responses:  176662.91
  - Announce responses: 176863.44
  - Scrape responses:   3578.91
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 105
  - p99.9: 287
  - p100: 351  
```
  
DashMap with initial capacity 0 (best result. On average is lower, similar to SkipMap):

```output
Requests out: 410658.38/second
Responses in: 365892.86/second
  - Connect responses:  181258.91
  - Announce responses: 181005.95
  - Scrape responses:   3628.00
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 104
  - p99.9: 295
  - p100: 363
```

With Criterion:

image

image

image

image

### Conclusion

From my point of view, other [performance optimisations](https://github.com/torrust/torrust-tracker/discussions/774) have more potential than this, considering this implementation is not finished. The changes needed to finish this implementation will probably decrease the performance obtained in this benchmarking. In the future, we can review this option if we change the [tracker API behaviour to getting all torrents](https://github.com/torrust/torrust-tracker/discussions/775).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1101,6 +1101,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "978747c1d849a7d2ee5e8adc0159961c48fb7e5db2f06af6723b80123bb53856"
+dependencies = [
+ "cfg-if",
+ "hashbrown 0.14.3",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "deranged"
 version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3916,6 +3929,7 @@ dependencies = [
  "colored",
  "config",
  "crossbeam-skiplist",
+ "dashmap",
  "derive_more",
  "fern",
  "futures",
@@ -4022,6 +4036,7 @@ dependencies = [
  "async-std",
  "criterion",
  "crossbeam-skiplist",
+ "dashmap",
  "futures",
  "rstest",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ url = "2"
 uuid = { version = "1", features = ["v4"] }
 
 [package.metadata.cargo-machete]
-ignored = ["serde_bytes", "crossbeam-skiplist"]
+ignored = ["serde_bytes", "crossbeam-skiplist", "dashmap"]
 
 [dev-dependencies]
 local-ip-address = "0"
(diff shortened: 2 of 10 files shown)


josecelano:
ACK 00ee9db340c5274a63aa5f321e3bd5064e967b6c

josecelano:
ACK c653eb3e68d32026fba09595123fca149d049c72

josecelano:
ACK 4030fd12b575c039dbb1507fbe99e0aaec58fd5f

PR #826, opened 2024-05-02
refactor: [#680] http return errors instead of panicking

### LINKED ISSUE #680: Refactor: `bit_torrent::tracker::http::client::Client` to return errors instead of panicking
The HTTP Tracker Client struct contains some `unwrap` calls.

```rust
pub struct Client {
    base_url: Url,
    reqwest: ReqwestClient,
    key: Option<Key>,
}

impl Client {
    // ...

    pub fn bind(base_url: Url, local_address: IpAddr) -> Self {
        Self {
            base_url,
            reqwest: reqwest::Client::builder().local_address(local_address).build().unwrap(),
            key: None,
        }
    }

    // ...

    pub fn authenticated(base_url: Url, key: Key) -> Self {
        Self {
            base_url,
            reqwest: reqwest::Client::builder().build().unwrap(),
            key: Some(key),
        }
    }

    // ...

    pub async fn get(&self, path: &str) -> Response {
        self.reqwest.get(self.build_url(path)).send().await.unwrap()
    }

    // ...

    pub async fn get_with_header(&self, path: &str, key: &str, value: &str) -> Response {
        self.reqwest
            .get(self.build_url(path))
            .header(key, value)
            .send()
            .await
            .unwrap()
    }

    // ...
}
```

That makes it impossible to catch those errors upstream. For example, the Tracker Checker can not use this `Client` and continue its execution if an HTTP Tracker request fails. See, for example, this issue: https://github.com/torrust/torrust-tracker/issues/678.

We should return the `Result` type and let the caller handle the problem.

This Client was introduced for testing purposes and later moved to production, so it lacks some important features like:

- Testing.
- Better error handling.

hungfnt:
Refactor: HTTP tracker client returns errors instead of panicking.

### DIFF
--- a/src/console/clients/checker/checks/http.rs
+++ b/src/console/clients/checker/checks/http.rs
@@ -61,9 +61,19 @@ async fn check_http_announce(tracker_url: &Url) -> Result<(), CheckError> {
     // We should change the client to catch that error and return a `CheckError`.
     // Otherwise the checking process will stop. The idea is to process all checks
     // and return a final report.
-    let response = Client::new(tracker_url.clone())
+    let Ok(client) = Client::new(tracker_url.clone()) else {
+        return Err(CheckError::HttpError {
+            url: (tracker_url.to_owned()),
+        });
+    };
+    let Ok(response) = client
         .announce(&QueryBuilder::with_default_values().with_info_hash(&info_hash).query())
-        .await;
+        .await
+    else {
+        return Err(CheckError::HttpError {
+            url: (tracker_url.to_owned()),
+        });
+    };
 
     if let Ok(body) = response.bytes().await {
         if let Ok(_announce_response) = serde_bencode::from_bytes::<Announce>(&body) {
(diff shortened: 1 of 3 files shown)


josecelano (approved):
Hi @ngthhu, I'm going to merge this becuase this is better than what we have currently. However, It would be good to open a discussion to  make your choice explicit and document it. You decided to use the [anyhow](https://github.com/dtolnay/anyhow) crate. I'm not sure if in this case that's the best option.

We could use two alternatives:

- Enums
- Type-erased errors

Even [anyhow recommends using thiserror crate](https://github.com/dtolnay/anyhow?tab=readme-ov-file#comparison-to-thiserror) "_if you are a library that wants to design your own dedicated error type(s) so that on failures the caller gets exactly the information that you choose_".

In this case, it seems clients might be interested in handling the different errors, but looking at another similar library like [request](https://docs.rs/reqwest/latest/src/reqwest/error.rs.html#16-18) they also used a BoxError. Maybe because there could be a lot of them and it would be hard to maintain an `Enum`.

Since this is not part of the public API (yet) it's not critical. If we decide to publish a crate we have to think about it carefully.

So it looks like a good choice to me.

To know more about "Enum" vs "Type-erased" errors I recommend the chapter 2 (ERROR HANDLING) of the Jon Gjengset' book: 

https://rust-for-rustaceans.com/

josecelano:
ACK 895efe9c4154a6883008e9cf26592f3fed38602d

josecelano:
Thank your @ngthhu, and remember to sign commits 🙏🏼.

PR #841, opened 2024-05-07
feat: log aborted UDP requests

### LINKED ISSUE #830: Monitor how many UDP requests are aborted when the maximum number of concurrent requests is reached
We handle a maximum of 50 UDP requests at the same time. When we have more than 50, we overwrite the latest one, and we give a second chance to the finish.

```rust
struct ActiveRequests {
    rb: StaticRb<AbortHandle, 50>, // the number of requests we handle at the same time.
}

// ...

async fn run_udp_server(tracker: Arc<Tracker>, socket: Arc<UdpSocket>) {
    let tracker = tracker.clone();
    let socket = socket.clone();

    let reqs = &mut ActiveRequests::default();

    // Main Waiting Loop, awaits on async [`receive_request`].
    loop {
        if let Some(h) = reqs.rb.push_overwrite(
            Self::spawn_request_processor(Self::receive_request(socket.clone()).await, tracker.clone(), socket.clone())
                .abort_handle(),
        ) {
            if !h.is_finished() {
                // the task is still running, lets yield and give it a chance to flush.
                h.abort();
            }
        }
    }
}

```

[The demo server has been busy recently](https://github.com/torrust/torrust-demo/issues/1) and I'm curious how many UDP requests we are discarding (not even replying).

By the way @da2ce7 we are using the `push_overwrite` [method](https://docs.rs/ringbuf/latest/ringbuf/traits/ring_buffer/trait.RingBuffer.html#method.push_overwrite). 

image

That method overwrites the latest item, not the oldest. Does that make sense in our case? Would not be better if we overwrite the oldest request? That's the one that has had a longer time to be processed.

I'm planning to add a log (warning) before the `h.abort()` just to track the number of requests that are not processed. Then I can parse the logs to get the data like [this](https://github.com/torrust/torrust-demo/issues/3).

josecelano:
This will add a warning to the logs when a UDP request is aborted.

### DIFF
--- a/src/servers/udp/server.rs
+++ b/src/servers/udp/server.rs
@@ -291,7 +291,14 @@ impl Udp {
                 if !h.is_finished() {
                     // the task is still running, lets yield and give it a chance to flush.
                     tokio::task::yield_now().await;
+
                     h.abort();
+
+                    let server_socket_addr = socket.local_addr().expect("Could not get local_addr for socket.");
+
+                    tracing::span!(
+                        target: "UDP TRACKER",
+                        tracing::Level::WARN, "request-aborted", server_socket_addr = %server_socket_addr);
                 }
             }
         }


josecelano:
ACK e3143f779c40e590cfbdcfd34ee3d972f94ca1b3

PR #846, opened 2024-05-08
dev: fix udp ring-buffer not looping

### LINKED ISSUE #842: Review process to abort UDP request when the ring buffer is full
Relates to: https://github.com/torrust/torrust-tracker/issues/830

```rust
    async fn run_udp_server(tracker: Arc<Tracker>, socket: Arc<UdpSocket>) {
        let tracker = tracker.clone();
        let socket = socket.clone();

        let reqs = &mut ActiveRequests::default();

        // Main Waiting Loop, awaits on async [`receive_request`].
        loop {
            if let Some(h) = reqs.rb.push_overwrite(
                Self::spawn_request_processor(Self::receive_request(socket.clone()).await, tracker.clone(), socket.clone())
                    .abort_handle(),
            ) {
                if !h.is_finished() {
                    // the task is still running, lets yield and give it a chance to flush.

                    h.abort();

                    let server_socket_addr = socket.local_addr().expect("Could not get local_addr for socket.");

                    tracing::span!(
                        target: "UDP TRACKER",
                        tracing::Level::WARN, "request-aborted", server_socket_addr = %server_socket_addr);
                }
            }
        }
    }
```

We are using the `push_overwrite` [method](https://docs.rs/ringbuf/latest/ringbuf/traits/ring_buffer/trait.RingBuffer.html#method.push_overwrite). 

image

That method overwrites the latest item, not the oldest. Does that make sense in our case? Would not be better if we overwrite the oldest request? That's the one that has had a longer time to be processed.

cc @da2ce7

josecelano:
@da2ce7's previous version would be limited to a single thread, as `push_overwrite` would keep on returning the last element when the ring-buffer was full.

Now, the ring-buffer is pre-filled and is looped over with a mutating iterator.

New handles are progressively swapped-in when the old entries are finished.

Note: I think that this implementation can be replaced with a standard vector with the same effect.

### DIFF
--- a/src/servers/udp/server.rs
+++ b/src/servers/udp/server.rs
@@ -24,7 +24,7 @@ use std::sync::Arc;
 use aquatic_udp_protocol::Response;
 use derive_more::Constructor;
 use log::{debug, error, info, trace};
-use ringbuf::traits::{Consumer, Observer, RingBuffer};
+use ringbuf::traits::{Consumer, Observer, Producer};
 use ringbuf::StaticRb;
 use tokio::net::UdpSocket;
 use tokio::sync::oneshot;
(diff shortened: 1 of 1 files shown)


josecelano:
ACK abb47b2def67d59dbb57bace4182f91d5ace075f

josecelano:
The new test is [failing on the runner](https://github.com/torrust/torrust-tracker/actions/runs/9000357499/job/24724464525?pr=846#step:7:478) but not on my machine 🤔

image

da2ce7:
@josecelano I've added a little delay to the test so it will pass:

https://github.com/da2ce7/torrust-tracker/commit/fc7707ceb671aa289987fa0f508779a662c3d6bd

josecelano:
> @josecelano I've added a little delay to the test so it will pass:
> 
> [da2ce7@fc7707c](https://github.com/da2ce7/torrust-tracker/commit/fc7707ceb671aa289987fa0f508779a662c3d6bd)

Updated and merged.

PR #854, opened 2024-05-09
Refactor: enrich field types in configuration structs

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/pull/790

Refactor: enrich field types in configuration structs.

This will cause the app to fail earlier while loading invalid configurations and simplify the code by reducing conversions to get the rich type from the primitive when it's used.

- [x] `HealthCheckApi`
- [x] `UdpTracker`
- [x] `HttpTracker`
- [x] `HttpApi`
- [x] `Configuration`

Output example when you provide an invalid socket address:

```output
$ cargo run
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.10s
     Running `target/debug/torrust-tracker`
Loading default configuration file: `./share/default/config/tracker.development.sqlite3.toml` ...
thread 'main' panicked at src/bootstrap/config.rs:45:32:
called `Result::unwrap()` on an `Err` value: ConfigError { source: LocatedError { source: Error { tag: Tag(Default, 2), profile: Some(Profile(Uncased { string: "default" })), metadata: Some(Metadata { name: "TOML source string", source: None, provide_location: Some(Location { file: "packages/configuration/src/v1/mod.rs", line: 397, col: 14 }), interpolater:  }), path: ["health_check_api", "bind_address"], kind: Message("invalid socket address syntax"), prev: None }, location: Location { file: "packages/configuration/src/v1/mod.rs", line: 400, col: 41 } } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

You get that error proving this config option:

```toml
[health_check_api]
bind_address = "127.0.0.1-1313"
```

The error contains all the information needed, although it could be more user-friendly. Maybe we can map that error to a simpler explanation like this:

```
Configuration error: invalid socket address provided in toml file PATH in section `health_check_api` option `bind_address`.
```

PR #857, opened 2024-05-10
Refactor: tracker core service only needs the core config

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/401

After [extracting the `Core` config type](https://github.com/torrust/torrust-tracker/pull/854/commits/b545b33c17b3f71591a98dba9aeab84294cd1a62), we don't need to pass the whole configuration to the tracker core service. We can pass only its config.


josecelano:
ACK abf79e8129d8d9ecaccff87d3e7ebe2df12cf071

josecelano:
ACK ae77ebc50f2286e4c23b3f539759ebe4d5a908d1

PR #858, opened 2024-05-13
Use only non-defaults in `toml` config templates

### LINKED ISSUE #856: Config overhaul: remove default values from config file templates

The new Figment configuration implementation allows you not to define config values when you want to use the default one. For example, you can run the tracker without providing any value:

```console
TORRUST_TRACKER_CONFIG="" cargo run
```

And it will use default values. For the time being, all options have a default value.

Since we don't need to specify the default values, we should refactor the default config files to remove options with default values:

```output
$ tree share/default/config
share/default/config
├── tracker_checker.json
├── tracker.container.mysql.toml
├── tracker.container.sqlite3.toml
├── tracker.development.sqlite3.toml
├── tracker.e2e.container.sqlite3.toml
└── tracker.udp.benchmarking.toml
```

That makes it easier to maintain them.

josecelano:
After implementing the configuration with Figment, it's possible to omit values if they are defaults. Therefore, we don't need to add all options in templates. We only need to add values that overwrite the default values.

For example, the new configuration file for development environment (`share/default/config/tracker.development.sqlite3.toml`) is:

```toml
[[udp_trackers]]
enabled = true

[[http_trackers]]
enabled = true
```

The only difference with the default values is UDP and HTTP services are enabled.

This makes it easier to maintain these templates and to see what changes have been made compared to the default values.

### DIFF
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -120,9 +120,19 @@ pub struct AnnouncePolicy {
 impl Default for AnnouncePolicy {
     fn default() -> Self {
         Self {
-            interval: 120,
-            interval_min: 120,
+            interval: Self::default_interval(),
+            interval_min: Self::default_interval_min(),
         }
+    }
+}
+
+impl AnnouncePolicy {
+    fn default_interval() -> u32 {
+        120
+    }
+
+    fn default_interval_min() -> u32 {
+        120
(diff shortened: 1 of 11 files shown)


josecelano:
ACK 445bd5361607d8dcf78fcab48450a8d10c6753b3

PR #865, opened 2024-05-15
Add timeouts for Axum server in the Tracker API

### LINKED ISSUE #612: Tracker API: Add a timeout to the TLS handshake, waiting for headers, and waiting for receiving headers
Parent issue: https://github.com/torrust/torrust-tracker/issues/603

Related to:

- <https://github.com/programatik29/axum-server/issues/116>
- <https://github.com/torrust/torrust-tracker/issues/612>
- <https://github.com/torrust/torrust-tracker/issues/613>
- <https://github.com/tokio-rs/axum/discussions/1383>
- <https://github.com/hyperium/hyper/issues/3178>
- <https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.client_request_timeout>
- <https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.tls_handshake_timeout>
- <https://gist.github.com/programatik29/36d371c657392fd7f322e7342957b6d1>

Adds a timeout to:

- The TLS handshake.
- The time the server waits for a request after opening a connection.
- The time waiting for the client to send the headers after starting sending them.

Recently we moved to [axum-server](https://crates.io/crates/axum-server) and [it allows setting TLS handshake timeout](https://github.com/programatik29/axum-server/pull/39).

josecelano:
Add timeouts for Axum server in the Tracker API:

1. Timeout for the first request after opening the HTTP connection. [408 Request Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408).
2. HTTP1 header read timeout. Set a timeout for reading client request headers. If a client does not transmit the entire header within this time, the connection is closed.
3. HTTP2 keep alive timeout: Sets a timeout for receiving an acknowledgement of the keep-alive ping. If the ping is not acknowledged within the timeout, the connection will be closed. Does nothing if `http2_keep_alive_interval` is disabled.
4. HTTP2 keep alive interval: Sets an interval for HTTP2 Ping frames should be sent to keep a connection alive.
5. [Tower middleware which applies a timeout to requests](https://docs.rs/tower-http/latest/tower_http/timeout/index.html).

The current implementation for the first timeout does not send any message ([408 Request Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408)). [It seems is also common practice](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408).

To test the first case, you can open a connection to the API without sending any request with:

```output
telnet 127.0.0.1 1212
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.
```

After 5 seconds you will see:

```output
Connection closed by foreign host.
```

For advanced manual testing, you can use: https://github.com/josecelano/axum-server-timeout


josecelano:
ACK 9e42a1a452804f1f14f5e6a99c38bdccabbc5001

PR #871, opened 2024-05-17
Add privacy methods to the `TrackerMode` enum

### LINKED ISSUE #870: Align TrackerMode with Index
In the Index the `TrackerMode` is defined with:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
pub enum TrackerMode {
    // todo: use https://crates.io/crates/torrust-tracker-primitives
    /// Will track every new info hash and serve every peer.
    Public,
    /// Will only serve authenticated peers.
    Private,
    /// Will only track whitelisted info hashes.
    Whitelisted,
    /// Will only track whitelisted info hashes and serve authenticated peers.
    PrivateWhitelisted,
}

impl Default for TrackerMode {
    fn default() -> Self {
        Self::Public
    }
}

impl TrackerMode {
    #[must_use]
    pub fn is_open(&self) -> bool {
        matches!(self, TrackerMode::Public | TrackerMode::Whitelisted)
    }

    #[must_use]
    pub fn is_close(&self) -> bool {
        !self.is_open()
    }
}
```

We should get this type from the `torrust-tracker-primitives` crate and update the Index. Howeveer the Index has added the concept "open" and "close". We need to add this concept to the enum in the Tracker before we update the Index.

`TrackerMode` in the Tracker.

```rust
#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
pub enum TrackerMode {
    /// Will track every new info hash and serve every peer.
    #[serde(rename = "public")]
    Public,

    /// Will only track whitelisted info hashes.
    #[serde(rename = "listed")]
    Listed,

    /// Will only serve authenticated peers
    #[serde(rename = "private")]
    Private,

    /// Will only track whitelisted info hashes and serve authenticated peers
    #[serde(rename = "private_listed")]
    PrivateListed,
}
```

josecelano:
The tracker mode can be:

- Public (Non-whitelisted)
- Listed (Whitelisted)
- Private (Non-whitelisted)
- PrivateListed (Whitelisted)

There should have been two different flags (in my opinion):

- Visibility: public or private
- Whitelisted: true or false

So we would have the same four combinations:

- Not whitelisted:
  - Public
  - Private
- Whitelisted
  - Public
  - Private

That's a pending refactor. For this PR, the goal is just to align this enum with what we added to the Index so we can use it in the Index via the primitive crate.

See https://github.com/torrust/torrust-index/blob/develop/src/config.rs#L140-L171

### DIFF
--- a/packages/primitives/src/lib.rs
+++ b/packages/primitives/src/lib.rs
@@ -67,7 +67,7 @@ pub type PersistentTorrents = BTreeMap<InfoHash, u32>;
 ///
 /// Refer to [Torrust Tracker Configuration](https://docs.rs/torrust-tracker-configuration)
 /// to know how to configure the tracker to run in each mode.
-#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
 pub enum TrackerMode {
     /// Will track every new info hash and serve every peer.
     #[serde(rename = "public")]
--- a/packages/primitives/src/lib.rs
+++ b/packages/primitives/src/lib.rs
@@ -85,3 +85,21 @@ pub enum TrackerMode {
     #[serde(rename = "private_listed")]
     PrivateListed,
 }
+
+impl Default for TrackerMode {
+    fn default() -> Self {
+        Self::Public
+    }
+}
+
+impl TrackerMode {
+    #[must_use]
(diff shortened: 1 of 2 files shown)


josecelano:
ACK 08179e317c0b7bd71fa72eeb71944267f2f1a985

josecelano:
ACK 0c9da2f48b04e149991f5cc9a86b25ee8d2ceb43

PR #873, opened 2024-05-18
Implment Request Stream for UDP Tracker

da2ce7:
This pull request improves the UDP tracker `server.rs` request processing logic; now using a `tokio::stream` for getting the requests for the ring buffer.

I hope this pull request will improve the performance of the tracker in multi-core setups.

@josecelano This code will require a special code-review.


da2ce7:
@josecelano This pull request is ready for review.

josecelano (commented):
Hi @da2ce7 I haven't reviewed the code yet but I have executed the UDP load lest:

### Current version

**MAX**

```
Requests out: 435973.31/second
Responses in: 392346.07/second
  - Connect responses:  194332.01
  - Announce responses: 194108.16
  - Scrape responses:   3905.90
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 105
  - p99.9: 315
  - p100: 391
```  

**MIN**

```
Requests out: 371407.92/second
Responses in: 334267.17/second
  - Connect responses:  165563.03
  - Announce responses: 165413.11
  - Scrape responses:   3291.02
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 104
  - p99.9: 271
  - p100: 339
```

### This version

**MAX**

```
Requests out: 157926.53/second
Responses in: 142133.56/second
  - Connect responses:  70206.70
  - Announce responses: 70498.45
  - Scrape responses:   1428.41
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 10
  - p99: 78
  - p99.9: 125
  - p100: 156s
```  

**MIN**

```
Requests out: 79844.49/second
Responses in: 71860.52/second
  - Connect responses:  35539.87
  - Announce responses: 35614.59
  - Scrape responses:   706.05
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 4
  - p95: 16
  - p99: 49
  - p99.9: 68
  - p100: 88
```

josecelano:
Hi @da2ce7 I found this challenge:

https://mas-bandwidth.com/writing-highly-scalable-backends-in-udp/

And solutions:

- https://mas-bandwidth.com/writing-scalable-backends-in-udp-the-solution/
- https://github.com/mas-bandwidth/udp/tree/main
- https://www.youtube.com/watch?v=NogIoDZdYqA

da2ce7:
@josecelano reimplmented with a much simpler, (and not-slower?), version. :)

josecelano:
> @josecelano reimplmented with a much simpler, (and not-slower?), version. :)

Hey @da2ce7 cool! I will review it tomorrow after the meeting. I'm almost finished with the configuration refactorings.

josecelano:
> @josecelano reimplmented with a much simpler, (and not-slower?), version. :)

Hi @da2ce7, this version is even better than the previous one 🥇 !!.

At least with this concrete performance test that I'm running. I will review the code.

I have to double-check that the current version has the same result. The computer is exactly the same. However, we updated the BIOS. I guess that should not affect it.

### This version

**MAX**

```
Requests out: 478727.77/second
Responses in: 431048.72/second
  - Connect responses:  213289.40
  - Announce responses: 213400.69
  - Scrape responses:   4358.63
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 104
  - p99.9: 343
  - p100: 431
```

**MIN**

```
Requests out: 433379.95/second
Responses in: 393697.26/second
  - Connect responses:  194904.16
  - Announce responses: 194860.20
  - Scrape responses:   3932.91
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 105
  - p99.9: 313
  - p100: 391
```

josecelano:
The current version (target branch) is slightly better (the max) but, on average, much slower than the new one (this PR).

The new MIN is almost as good as the previous MAX. I will run again the whole [benchmarking](https://torrust.com/benchmarking-the-torrust-bittorrent-tracker).

NEW MIN

Requests out: 433379.95/second
Responses in: 393697.26/second

CURRENT MAX

Requests out: 448806.70/second
Responses in: 403873.92/second

### Current version

**MAX**

```
Requests out: 448806.70/second
Responses in: 403873.92/second
  - Connect responses:  200015.65
  - Announce responses: 199875.73
  - Scrape responses:   3982.54
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 104
  - p99.9: 323
  - p100: 403
```

**MIN**

```
Requests out: 381837.70/second
Responses in: 343653.57/second
  - Connect responses:  170156.07
  - Announce responses: 170132.54
  - Scrape responses:   3364.96
  - Error responses:    0.00
Peers per announce response: 0.00
Announce responses per info hash:
  - p10: 1
  - p25: 1
  - p50: 1
  - p75: 1
  - p90: 2
  - p95: 3
  - p99: 105
  - p99.9: 277
  - p100: 355
```

src/servers/udp/server.rs
```diff
@@ -252,124 +329,178 @@ impl Udp {
...
-
-        info!(target: "UDP TRACKER", "Starting on: udp://{}", address);
-
-        let running = tokio::task::spawn(async move {
-            debug!(target: "UDP TRACKER", "Started: Waiting for packets on socket address: udp://{address} ...");
-            Self::run_udp_server(tracker, socket).await;
-        });
+        let halt_task = tokio::task::spawn(shutdown_signal_with_message(
+            rx_halt,
+            format!("Halting Http Service Bound to Socket: {bind_to}"),
```
josecelano, line 334: Hi @da2ce7 I think the message should be "Halting UDP Service Bound to Socket: {bind_to}"

src/servers/udp/server.rs
```diff
@@ -252,124 +329,178 @@ impl Udp {
...
+        };
+
+        let address = socket.local_addr().expect("it should get the locally bound address");
+        let local_addr = format!("udp://{address}");
+
+        // note: this log message is parsed by our container. i.e:
+        //
+        // `[UDP TRACKER][INFO] Starting on: udp://`
+        //
+        tracing::info!(target: "UDP TRACKER", "Starting on: {local_addr}");
```
josecelano, line 356: Hi @da2ce7, maybe we could make this dependency explicit. We could move the const from:

- https://github.com/torrust/torrust-tracker/blob/develop/src/console/ci/e2e/logs_parser.rs#L5-L8
- https://github.com/torrust/torrust-tracker/blob/develop/src/console/ci/e2e/logs_parser.rs#L67-L69

to the UDP server mod. Ideally, we should also use a field in the log now that we are using tracing. But that could be implemented in a new PR.

src/servers/udp/server.rs
```diff
@@ -235,6 +234,84 @@ impl Drop for ActiveRequests {
     }
 }
 
+/// Wrapper for Tokio [`UdpSocket`][`tokio::net::UdpSocket`] that is bound to a particular socket.
+struct Socket {
```
josecelano, line 238: Hi @da2ce7, I would call it `BoundSocket` and add a `local_addr()` method.

src/servers/udp/server.rs
```diff
@@ -252,124 +329,178 @@ impl Udp {
...
+        //
+        tracing::info!(target: "UDP TRACKER", "Starting on: {local_addr}");
+
+        let socket = socket.socket;
+
+        let direct = Receiver {
+            socket,
+            tracker,
+            data: RefCell::new([0; MAX_PACKET_SIZE]),
+        };
```
josecelano, line 364: Hi @da2ce7, why did you call this "direct"?

src/servers/udp/server.rs
```diff
@@ -235,6 +234,84 @@ impl Drop for ActiveRequests {
...
+            Ok(from) => {
+                let payload = buf.filled().to_vec();
+                let request = UdpRequest { payload, from };
+
+                Some(Ok(tokio::task::spawn(Udp::process_request(
+                    request,
+                    self.tracker.clone(),
+                    self.socket.clone(),
+                ))
+                .abort_handle()))
```
josecelano, line 306: @da2ce7 I like this trait implementation. However, I like the idea of having a more "pure" receiver that only returns the request. I would move this spawn to the main loop or maybe the Reciever could have another wrapper `StreamProcessor` or something like that that takes the requests and launches the worker. I think in the future, we could have a limited number of workers instead of spawning a new task, although I guess Tokio handles that.

src/servers/udp/server.rs
```diff
@@ -252,124 +329,178 @@ impl Udp {
...
+                        }
+                        tracing::error!(target: "UDP TRACKER: Udp::run_udp_server::loop", local_addr, err = %e,  "break: (got error)");
+                        break;
+                    }
+                };
+
+                if req.is_finished() {
+                    continue;
+                }
+
```
josecelano, line 420: Hi @da2ce7 I guess from this point on, we have created a new task that has not been finished yet and we need to add it to the "pending to finish" limited buffer. Why don't we extract a type "buffer" for that? or move that logic to the `ActiveRequests` Struct.

**ChatGPT refactor (not tested).**

Given that you already have an `ActiveRequests` struct with a ring buffer (`StaticRb`), it makes sense to move the task management logic into `ActiveRequests`.

### Step 1: Update `ActiveRequests` Struct

We'll move the logic for making space for new tasks and managing the buffer into `ActiveRequests`. Here's the updated struct:

```rust
use ringbuf::StaticRb;
use tokio::task::AbortHandle;

#[derive(Default)]
struct ActiveRequests {
    rb: StaticRb<AbortHandle, 50>, // the number of requests we handle at the same time.
}

impl ActiveRequests {
    fn new(max_size: usize) -> Self {
        Self {
            rb: StaticRb::new(max_size),
        }
    }

    async fn add_task(&mut self, task: AbortHandle) -> Result<(), AbortHandle> {
        if self.rb.is_full() {
            self.make_space_for_new_task(task).await
        } else {
            self.rb.push(task).expect("Ring buffer should not be full");
            Ok(())
        }
    }

    async fn make_space_for_new_task(&mut self, new_task: AbortHandle) -> Result<(), AbortHandle> {
        let mut finished_count = 0;
        let mut unfinished_task = None;

        for _ in 0..self.rb.len() {
            if let Some(mut task) = self.rb.pop() {
                if task.is_finished() {
                    finished_count += 1;
                    continue;
                }

                if task.is_finished() {
                    finished_count += 1;
                    continue;
                }

                if finished_count == 0 {
                    task.abort();
                    return Err(new_task);
                }

                unfinished_task = Some(task);
            }
        }

        if let Some(task) = unfinished_task {
            self.rb.push(task).expect("Ring buffer should not be full");
        }

        self.rb.push(new_task).expect("Ring buffer should not be full");
        Ok(())
    }
}

impl std::fmt::Debug for ActiveRequests {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let (left, right) = &self.rb.as_slices();
        let dbg = format!("capacity: {}, left: {left:?}, right: {right:?}", &self.rb.capacity());
        f.debug_struct("ActiveRequests").field("rb", &dbg).finish()
    }
}

impl Drop for ActiveRequests {
    fn drop(&mut self) {
        for h in self.rb.pop_iter() {
            if !h.is_finished() {
                h.abort();
            }
        }
    }
}
```

### Step 2: Refactor the Main Loop

Update the main loop to use the `ActiveRequests` for managing the tasks:

```rust
impl Udp {
    async fn run_udp_server_main(mut direct: Receiver) {
        let addr = direct.socket.local_addr().expect("it should get local address");
        let local_addr = format!("udp://{addr}");

        // Create an ActiveRequests instance
        let mut active_requests = ActiveRequests::new(MAX_BUFFER_SIZE);

        loop {
            if let Some(req) = {
                tracing::trace!(target: "UDP TRACKER: Udp::run_udp_server", local_addr, "(wait for request)");
                direct.next().await
            } {
                tracing::trace!(target: "UDP TRACKER: Udp::run_udp_server::loop", local_addr, "(in)");

                let req = match req {
                    Ok(req) => req,
                    Err(e) => {
                        if e.kind() == std::io::ErrorKind::Interrupted {
                            tracing::warn!(target: "UDP TRACKER: Udp::run_udp_server::loop", local_addr, err = %e,  "(interrupted)");
                            return;
                        }
                        tracing::error!(target: "UDP TRACKER: Udp::run_udp_server::loop", local_addr, err = %e,  "break: (got error)");
                        break;
                    }
                };

                if req.is_finished() {
                    continue;
                }

                // Try to add the request to the buffer
                if let Err(req) = active_requests.add_task(req.abort_handle()).await {
                    tracing::warn!(target: "UDP TRACKER: Udp::run_udp_server::loop", local_addr, "aborting request: (buffer full)");
                }
            } else {
                tracing::error!(target: "UDP TRACKER: Udp::run_udp_server", local_addr, "breaking: (ran dry, should not happen in production!)");
                break;
            }
        }
    }

    // Other methods...
}
```

### Summary

1. **Update `ActiveRequests`**:
   The `ActiveRequests` struct now includes methods for adding tasks and making space for new tasks by processing or aborting old tasks.

2. **Refactor Main Loop**:
   Use the `ActiveRequests` struct in the main loop to manage the task buffer and ensure there is space for new tasks.

This refactoring moves the buffer management and task completion logic into `ActiveRequests`, making the main loop cleaner and more focused on handling UDP requests.

josecelano (approved):
Hi @da2ce7, I think this improves two important things:

- It removes the mess of "processor" functions.
- It introduces some abstractions (Receiver, BoundSocket, ...) that make the code more straightforward and clear.

And it improves significantly the performance, so I think we should merge it. All my comments are only suggestions to potentially improve the readability, and they can be done later.

josecelano:
ACK 84cc1a1d39432456b2b0c306d8289f6a618da7b1

josecelano:
Hi @da2ce7, just for the record, this PR behaves the same way as the current version. We have a limit of 50 active requests. When the server has more than 50 simultaneously, it starts aborting the oldest tasks. As far as I can see, this enabled a kind of indirect timeout. Maybe we should increase the buffer and set an explicit timeout to process UDP requests. I guess that would add an overhead, and in the end, the effect would be the same; some requests are aborted when the server is overloaded. The advantage of this solution is the timeout is somehow dynamic. Just out of curiosity, I've run the test load for 1 minute on my PC, and there were 6 aborted requests. We have to check that this is not affecting the benchmarking. I don't think so because the UDP load tests count the responses.

src/servers/udp/server.rs
da2ce7: @josecelano yes, this is a poor name, it is from a experiment where the requests were 'directly' made...

josecelano: Hi, @da2ce7 ok.

josecelano:
Hi @da2ce7 I'm working on the refactorings I've suggested in this PR: https://github.com/torrust/torrust-tracker/pull/913

Now that I'm working on the code and not just reviewing it, I've noticed one thing. When we receive a new request, we immediately spawn a new task to handle it. However, we limit the number of active requests to 50. If we are already handling 50 requests and a new request comes in, we spawn a new task and remove the oldest tasks in the active requests buffer. Would not it make sense not to spawn new tasks if the active request buffer is already full? 

We could have a "pending to activate requests" buffer in front of the active reqs buffer. When tasks in the active buffer finish we get more requests from the first requests buffer and spwan a new task to handle them.

**Pros:**

- We don't waste more resources spawning tasks that will not be executed immediately.
- We don't need to remove active requests (we don't need to make a place for new requests if the active reqs buffer is complete).

**Cons:**

- We indirectly increase the timeout. This could be a good or a bad thing. It would be good if the timeout was too short, and bad if it was too long.

By the way, do you want to continue working on this PR? I can stop working on the [new one ](https://github.com/torrust/torrust-tracker/pull/913), but I think it's a good thing to alternately work on the same to share the code ownership.

src/servers/udp/server.rs
da2ce7: Merged in: https://github.com/torrust/torrust-tracker/pull/913

PR #921, opened 2024-06-28
Add comments to the UDP server

### LINKED ISSUE #918: UDP server: alternative implementation to avoid aborting too many requests
I've udpated the [demo](https://index.torrust-demo.com/torrents), including the tracker:

udp://tracker.torrust-demo.com:6969/announce

@da2ce7 added a new log in this [PR](https://github.com/torrust/torrust-tracker/pull/873):

```rust
tracing::warn!(target: UDP_TRACKER_LOG_TARGET, local_addr, "Udp::run_udp_server::loop aborting request: (no finished tasks)");
```

### Context

When the tracker receives a new UDP request:

1. It spawns a new task to handle the requests.
2. It gets the abort handle for that task.
3. It tries to insert that handle in the active requests buffer where only 50 concurrent requests are allowed.
4. If the buffer is full, it tries to remove finished tasks.
5. If all the tasks are unfinished tasks, it removes the oldest task.

That's the point where we log the new warning:

In the demo environment, we have aprox 42 tasks aborted in 2 minutes. It means a log entry like this:

```s
tracker  | 2024-06-26T10:08:08.933194Z  WARN UDP TRACKER: Udp::run_udp_server::loop aborting request: (no finished tasks) local_addr="udp://0.0.0.0:6969"
```

Of course, it's not a regular number.

### Problem

I have described the problem in:

- https://github.com/torrust/torrust-tracker/pull/873#issuecomment-2186011239
- https://github.com/torrust/torrust-tracker/pull/873#issuecomment-2188470903

### Proposal 1: New buffer for request pending to activate

I have also described a proposal in:

- https://github.com/torrust/torrust-tracker/pull/873#issuecomment-2188470903

The basic idea is to add a new buffer for requests that can't be handled because we already have 50 active requests. I'm copying my comment from the PR:

>  When we receive a new request, we immediately spawn a new task to handle it. However, we limit the number of active requests to 50. If we are already handling 50 requests and a new request comes in, we spawn a new task and remove the oldest tasks in the active requests buffer. Would not it make sense not to spawn new tasks if the active request buffer is already full? 
> 
> We could have a "pending to activate requests" buffer in front of the active reqs buffer. When tasks in the active buffer finish we get more requests from the first requests buffer and spwan a new task to handle them.
> 
> **Pros:**
> 
> - We don't waste more resources spawning tasks that will not be executed immediately.
> - We don't need to remove active requests (we don't need to make a place for new requests if the active reqs buffer is complete).
> 
> **Cons:**
> 
> - We indirectly increase the timeout. This could be a good or a bad thing. It would be good if the timeout was too short, and bad if it was too long.
> 

I think we need to add an explicit timeout to the time spent processing the single requests (this is needed anyway) and we also need to remove aborted tasks for the active request buffer periodically.

I have to think deeper about it.

### Proposal 2: Increase the number of active requests

We could also make it a config option.

I think we need to guarantee that we handle the requests we accept on time and we reject the ones we can not handle. In the current solution with high load, we can abort a request that has not had enough time to be processed even if it was the only task on the server.

### Proposal 3: Abort the new request

If we receive a new request and we can not handle it because we are already handling 50 requests, we abort the new request.

New request will be rejected, and the server can start accepting new requests when other previous tasks finish.

This is the simplest solution. In fact, we can reject the new request even before inserting it in the active requests buffer.

Even if we implement solution 1, at some point, we have to implement something like this because we can not grow the first bugger indefinitely.

This solution requires cleaning finished tasks anyway.

1. If the buffer is full we attempt to clean finished tasks.
2. If none of the tasks are finished we ignore the new requests.
3. If some tasks were finished, we spawn the new tasks and insert the new abort handle.

### Extra

I've added a comment on the code [here](https://github.com/torrust/torrust-tracker/blob/develop/src/servers/udp/server/launcher.rs#L130-L148).

### Conclusion

- I would implement proposal 3 now and proposal 1 in the future.
- I would also research what other trackers are doing. There is an ongoing discussion about how to avoid overload: https://github.com/torrust/torrust-tracker/discussions/825

josecelano:
After a dicussion with @da2ce7 I understood the reasons why he implemented the active requests buffer in the way it's implemented. This PR adds some clarifications.

My main concern was tasks starvation, which means if we always accept new requests (spawning new tasks) we could not give enough time to current active tasks to finish (a reasonable time).

@da2ce7 clarify that the yieldinf `tokio::task::yield_now().await;` should give the task enough time to finish.

My second concern was why we immediately spawn new tasks for the incomming requests instead of waiting until we have place in the buffer. @da2ce7 replied that we are going to run that's tasks anyway because we force a place for the new tasks.

### DIFF
--- a/src/servers/udp/server/request_buffer.rs
+++ b/src/servers/udp/server/request_buffer.rs
@@ -4,10 +4,15 @@ use tokio::task::AbortHandle;
 
 use crate::servers::udp::UDP_TRACKER_LOG_TARGET;
 
-/// Ring-Buffer of Active Requests
+/// A ring buffer for managing active UDP request abort handles.
+///
+/// The `ActiveRequests` struct maintains a fixed-size ring buffer of abort
+/// handles for UDP request processor tasks. It ensures that at most 50 requests
+/// are handled concurrently, and provides mechanisms to handle buffer overflow
+/// by removing finished or oldest unfinished tasks.
 #[derive(Default)]
 pub struct ActiveRequests {
-    rb: StaticRb<AbortHandle, 50>, // the number of requests we handle at the same time.
+    rb: StaticRb<AbortHandle, 50>, // The number of requests handled simultaneously.
 }
 
 impl std::fmt::Debug for ActiveRequests {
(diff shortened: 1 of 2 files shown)


josecelano:
### Starvation

Hi @da2ce7, I'm still not sure if we can't have a starvation issue with this implementation. It seems we are assuming that by just giving the task a second chance to finish with "yield", the task can finish, but in the end, resources are limited, and the tasks can have a lock. It could be waiting to get the lock to update the torrent repository. I have the impression that we can have scenarios where:

- We continue spawning tasks, consuming more memory faster than what the server can handle.
- We abort tasks without giving them a minimum time to be processed.

I have the impression that we should only reject new requests when the buffer is full. Maybe we can try to simulate that scenario with a shorter buffer.

### LocalRb

@da2ce7 should not be possible to use a [LocalRb](https://docs.rs/ringbuf/latest/ringbuf/rb/local/struct.LocalRb.html) instead of a [StaticRb](https://docs.rs/ringbuf/latest/ringbuf/type.StaticRb.html). If I'm not wrong, only one thread owns the `ActiveRequests` variable and it's not sent between threads.

josecelano:
ACK a897de43dd22aa15b63800e2fd1ade8d5d424533

src/servers/udp/server/request_buffer.rs
```diff
@@ -29,67 +34,103 @@ impl Drop for ActiveRequests {
...
+                // Reinsert the unfinished task if any.
+                if let Some(h) = unfinished_task {
+                    self.rb.try_push(h).expect("it was previously inserted");
+                }
+
+                // Insert the new task, ensuring there's space.
+                if !abort_handle.is_finished() {
+                    self.rb
+                        .try_push(abort_handle)
+                        .expect("it should remove at least one element.");
```
da2ce7: "it should have space for this new task"

src/servers/udp/server/request_buffer.rs
```diff
@@ -29,67 +34,103 @@ impl Drop for ActiveRequests {
...
             }
-
-            // task is unfinished.. give it another chance.
-            tokio::task::yield_now().await;
-
-            // if now finished, we continue.
-            if h.is_finished() {
-                finished += 1;
-                continue;
+            Err(abort_handle) => {
```
da2ce7: `abort_handle`, ->  `new_task`

src/servers/udp/server/request_buffer.rs
```diff
@@ -29,67 +34,103 @@ impl Drop for ActiveRequests {
...
+                // Buffer is full, attempt to make space.
+
+                let mut finished: u64 = 0;
+                let mut unfinished_task = None;
+
+                for removed_abort_handle in self.rb.pop_iter() {
```
da2ce7: `removed_abort_handle`, -> `old_task`

src/servers/udp/server/request_buffer.rs
```diff
@@ -29,67 +34,103 @@ impl Drop for ActiveRequests {
...
+                // one thread owns the active buffer. There is no way for the
+                // buffer to be full again. That means the "expects" should
+                // never happen.
+
+                // Reinsert the unfinished task if any.
+                if let Some(h) = unfinished_task {
+                    self.rb.try_push(h).expect("it was previously inserted");
+                }
+
+                // Insert the new task, ensuring there's space.
```
da2ce7:
``` text
Space has already been made for this new task in the buffer.
One or many old task have already been finished or yielded, freeing space in the buffer. OR
A single unfinished task has been aborted to make space for this new task.
```

da2ce7:
@josecelano

> @da2ce7 clarify that the yieldinf `tokio::task::yield_now().await;` should give the task enough time to finish.

In fact we can yield many times without aborting tasks if each yield leads to the completion of the old task.

> My second concern was why we immediately spawn new tasks for the incomming requests instead of waiting until we have place in the buffer. @da2ce7 replied that we are going to run that's tasks anyway because we force a place for the new tasks.

We start the process of removing the tasks after starting the new task, then we do not even insert the new task if it is already finished after making space for it.

josecelano:
> @josecelano
> 
> > @da2ce7 clarify that the yieldinf `tokio::task::yield_now().await;` should give the task enough time to finish.
> 
> In fact we can yield many times without aborting tasks if each yield leads to the completion of the old task.
> 
> > My second concern was why we immediately spawn new tasks for the incomming requests instead of waiting until we have place in the buffer. @da2ce7 replied that we are going to run that's tasks anyway because we force a place for the new tasks.
> 
> We start the process of removing the tasks after starting the new task, then we do not even insert the new task if it is already finished after making space for it.

Hey @da2ce7 I think I was missing a critical point. Since we are only receiving new requests in a single thread there is no way to increase the number of request processor tasks more than 51 tasks. The socket buffer is doing the contention. New "raw" requests have to wait until a current active request finishes or aborts.

Regarding the other topic "having enough time to be completed" before the force the taste to abort, the scenario I'm worried is:

- Request processors (spawned tasks to handle a single request) take 20 seconds to finish because all requests are announce requests that require locking the torrent repository, and for some reason (high-load also in the API, for example) they have to wait longer.
- The current 50 active requests are unfinished, but it's normal to wait long.
- Every new request is going to abort one of the pending requests not letting the whole system to advance.

In that case we are just switching old tasks with new tasks. In that scenario I would prefer to abort current tasks if the time we have to wait is longer that a reasonable response time for the client. I mean, it does not make sense to wait 120 seconds because the client will make a new request, but it makes sense to wait 2 seconds and reject all new raw requests. In general I would prefer the system to reply to fewer clients slower rather than not replying at all to any client. Am I missing something?

**UPDATE**

Maybe we can store in the buffer not only the abort handle but also a timestamp for when the task was added. We only remove the task after a timeout. If all 50 tasks are under that timeout we continue yielding.

da2ce7:
@josecelano I think that you are mixing up getting tasks throughput and anti-dos.

For the server to start thrashing (i.e. to not process any requests), any particular request will have to be yielded to 50 times.

If somebody is making lots of very-difficult to fulfill requests and filling the buffer so it thrashes; short-lived tasks created by a non-dossing user can still slip in and will get yielded to.

josecelano:
> @josecelano I think that you are mixing up getting tasks throughput and anti-dos.

What's the difference? I mean, I'm trying to analyse what would happen on a high load scenario whether it's unintended or an attack. Only having a lot of announce requests that are computing to acces the torrent repository with other services.

> For the server to start thrashing (i.e. to not process any requests), any particular request will have to be yielded to 50 times.
> 

Sorry, I haven't realized that because I was not assuming we are popping the eldest item. Of course that's the expected behaviour of the ring buffer but the documentation does not make explicit for this method and since there are other methods I was assuming this "pop" is somehow random.

On the other hand, I don't know if 50 thread context switches is enough or not compare to what it takes to process a request. I guess you are assuming so.

> If somebody is making lots of very-difficult to fulfill requests and filling the buffer so it thrashes; short-lived tasks created by a non-dossing user can still slip in and will get yielded to.

I suppose you assume the attacker uses a kind of heavy announce request that takes longer to process than the normal one. And I think you mean all tasks are going to have a chance to be executed and there is no way in this solution to avoid a bad actor. That's fine, I was not expecting to detect them and stop them. I was only trying to avoid wasting time just switching contexts. 

If 50 thread context switches are enough to process a request in a reasonable response time in a server under pressure, then the solution is perfect. It's a kind of indirect timeout where we are not using a timestamp bit a context switches counter. I have the impression that the timeout should be bound to user expectations (what a reasonable response time is). This is more empirical, if we can't handle a request in 50 context switches then the system is under heavy load and we abort tasks. I think this would be the perfect solution if the reason for the server to underperform is because there are requests that are heavy to process. 

But if the CPU load is high in general because there are a lot of requests in parallel that are waiting for a lock (for example) and 50 context switches is a relatively short time in comparison with a reasonable response time then it could lead to non replaying any client instead of sending responses with a longer response time.

Anyway, I think 🤔 I have to think deeper about it :-).

src/servers/udp/server/request_buffer.rs
```diff
@@ -29,67 +34,105 @@ impl Drop for ActiveRequests {
...
```
da2ce7: // Insert the new task: ~~, ensuring there's space.~~

da2ce7 (approved):
Good Comments and Clarification for the Task Ring Buffer Implementation :+1:

josecelano:
ACK 2c3edaa346f0ab7c60cb01a5249fbbdef8c84e44

josecelano:
ACK d1c2d15c7050ff75f04e026652b429077c8c2ace

PR #940, opened 2024-07-01
Config overhaul: define current config version as version 2

### LINKED ISSUE #939: Config overhaul: define current config version as version 2

Internally, the current version of the new configuration after the overhaul is version 1. I intended to start versioning from this version, but after a team discussion, we have decided to use version 2 even if we don't support version 1 anymore.

It was defined in the issues as version 2.

**NOTE**: We will not support version 1, although we can refer to it.

josecelano:
Config overhaul: define current config version as version 2.


josecelano:
ACK 632ad0d662f34070da38d9f82e1501a3b4f588bf

PR #941, opened 2024-07-01
Config overhaul: improve admin experience

### LINKED ISSUE #935: Config overhaul: improve admin experience

- Add a warning when the configuration does not enable any service (UDP tracker, HTTP tracker, or tracker API). It does not make sense to run the app without services enabled.
- Write the final used config to the logs when the tracker starts. We are only writing the config values when the source is an env var. We should do it always and print out the final configuration after emerging sources (defaults -> TOML file -> env vars).

josecelano:
- Add a warning when the configuration does not enable any service (UDP tracker, HTTP tracker, or tracker API). It does not make sense to run the app without services enabled.
- Write the final used config to the logs when the tracker starts. We are only writing the config values when the source is an env var. We should do it always and print out the final configuration after merging sources (defaults -> TOML file -> env vars).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4012,6 +4012,7 @@ dependencies = [
  "derive_more",
  "figment",
  "serde",
+ "serde_json",
  "serde_with",
  "thiserror",
  "toml",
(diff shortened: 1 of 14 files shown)


josecelano:
ACK 3244e53f0cc7158a1977bc124e57c7fe88944d2d

josecelano:
This is how it looks now when you run the tracker:

image

josecelano:
ACK 397ef0f6d003d4887272b6e25a2c5bbad9cc96c8

PR #947, opened 2024-07-02
Config overhaul: add version and namespace to the configuration

### LINKED ISSUE #937: Config overhaul: add version and namespace to the configuration

@da2ce7 proposed adding a version field to the configuration and also a namespace.

```toml
[torrust_tracker]
version = "1.0.0"

[torrust_tracker.logging]
log_level = "info"
```

```json
{
  "torrust_tracker": {
    "version": "1.0.0",
    "logging": {
      "log_level": "info"
    }
  }
}
```

@da2ce7, is this the way you thought about including the namespace? I think it's a good idea because it allows merging different app configurations into a single TOML file (for example, a single Tracker and Index). However, it adds one level of indentation to the whole field.

josecelano:
Add version to the configuration. It will fail if the provided version is not supported.

```toml
version = "2"
```

It only supports the exact match '2'.

### DIFF
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use camino::Utf8PathBuf;
-use derive_more::Constructor;
+use derive_more::{Constructor, Display};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


josecelano:
ACK 60c68762b64b7b909987fc6002447af7c823adbe

PR #959, opened 2024-07-13
dev: remove async trait dep

da2ce7:
Remove Unneeded Async Trait Dependency


da2ce7:
ACK 64850aff2d29ce368f7c2be2edb7c4c7b789441d

PR #961, opened 2024-07-13
dev: remove some internal types in preference for aquatics

da2ce7:
Prefer the protocol types in https://github.com/greatest-ape/aquatic/tree/master

- Removed `AnnounceEvent` for aquatic type.
- `InfoHash` uses aquatic type internally.
- Use aquatic `PeerId` and `peer::Id` uses aquatic type internally.


da2ce7:
ACK bef5680a02d43aacf11684ab3b3245e6cace7fb6

PR #962, opened 2024-07-13
use major version in toml files, and update major versions

da2ce7:


da2ce7:
ACK 9791427541ffba649b6803797df12117536585db

PR #980, opened 2024-07-30
Do not allow invalid tracker keys

### LINKED ISSUE #976: Bug: tracker keys can contain any char not only alphanumeric chars

We generate random 32-char keys like this: `YZSl4lMZupRuOpSRC3krIKR5BPB14nrJ`, with only alphanumeric chars.

This is the statement we use to generate the keys:

```rust
let random_id: String = thread_rng()
    .sample_iter(&Alphanumeric)
    .take(AUTH_KEY_LENGTH)
    .map(char::from)
    .collect();
```

And this is the `Key` struct.

```rust
/// A randomly generated token used for authentication.
///
/// It contains lower and uppercase letters and numbers.
/// It's a 32-char string.
#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Clone, Display, Hash)]
pub struct Key(String);
```

After merging this [PR](https://github.com/torrust/torrust-tracker/pull/975), it will be possible to upload preexisting keys. The uploaded keys should have the same restrictions:

- 32 chars
- Only alphanumeric chars: a-z, A-Z and 0-9.

However, the `Key` struct does not enforce these restrictions. It only checks that the string length is 32. That will allow uploading keys like this: `%QEs2ZNcCm9cwEV9dBpcPB5OwNFWFiRd`, which is not a valid key.

josecelano:
Do not allow invalid tracker keys. Keys cannot be constructed if they:

- Are longer than 32 chars.
- Contain non-alphanumeric ASCII chars.

### DIFF
--- a/src/core/auth.rs
+++ b/src/core/auth.rs
@@ -146,11 +146,11 @@ impl Key {
     /// Valid keys can only contain 32 chars including 0-9, a-z and A-Z.
     pub fn new(value: &str) -> Result<Self, ParseKeyError> {
         if value.len() != AUTH_KEY_LENGTH {
-            return Err(ParseKeyError);
+            return Err(ParseKeyError::InvalidKeyLength);
         }
 
         if !value.chars().all(|c| c.is_ascii_alphanumeric()) {
-            return Err(ParseKeyError);
+            return Err(ParseKeyError::InvalidChars);
         }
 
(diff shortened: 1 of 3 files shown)


josecelano:
ACK e81914b9dd1da75196f0de0a529262c16e4d1b2b

PR #983, opened 2024-07-31
Overhaul tracker keys: allow permanent keys

### LINKED ISSUE #979: Overhaul tracker keys: allow permanent keys

We can allow single keys to be permanent. 

```rust
/// An authentication key which has an expiration time.
/// After that time is will automatically become invalid.
#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Clone)]
pub struct ExpiringKey {
    /// Random 32-char string. For example: `YZSl4lMZupRuOpSRC3krIKR5BPB14nrJ`
    pub key: Key,
    /// Timestamp, the key will be no longer valid after this timestamp
    pub valid_until: DurationSinceUnixEpoch,
}
```

There are some possible implementations:

### Proposal 1: Set `valid_until` to 0

ExpiringKeys with the UnixEpoch value in the `valid_until` field don't expire.

### Proposal 2: Make `valid_until` optional

```rust
pub struct ExpiringKey {
    pub key: Key,
}
```

**NOTICE:** I would rename the struct to `TrackerKey` because it can be permanent. Alternatively, we could rename `Key` to `Token` and `ExpiringKey` to `Key`.

cc @da2ce7

josecelano:
This PR introduces a new feature. Tracker keys can be permanent.

#### Upload a pre-generated and expiring key

```
curl -X POST http://localhost:1212/api/v1/keys?token=MyAccessToken \
     -H "Content-Type: application/json" \
     -d '{
     	   "key": "Xc1L4PbQJSFGlrgSRZl8wxSFAuMa2110",
           "seconds_valid": 7200
         }'        
```

#### Upload a pre-generated and permanent key

```
curl -X POST http://localhost:1212/api/v1/keys?token=MyAccessToken \
     -H "Content-Type: application/json" \
     -d '{
     	   "key": "Xc1L4PbQJSFGlrgSRZl8wxSFAuMa2110",
           "seconds_valid": null
         }'        
```

#### Generate a random and expiring key

You can omit the key, and the tracker will generate a random one.

```
curl -X POST http://localhost:1212/api/v1/keys?token=MyAccessToken \
     -H "Content-Type: application/json" \
     -d '{
     	   "key": null,
           "seconds_valid": 7200
         }'        
```

#### Generate a random and permanent key

```
curl -X POST http://localhost:1212/api/v1/keys?token=MyAccessToken \
     -H "Content-Type: application/json" \
     -d '{
     	   "key": null,
           "seconds_valid": null
         }'        
```

### NOTES

- Fields in the json object are mandatory. You cannot omit them. Instead of omitting the field, I've decided to use the `null` value to represent the case where the value is not needed. I prefer to use always the same structure and make it explicit. However, omitting the field could have other advantages, like requiring less maintenance. For example, if we rename a field (breaking change), that would not affect users who are not using that field.
- I've introduced manual database migrations. The `second_valid` column in the `keys` table is now nullable (for permanent keys).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3886,6 +3886,7 @@ dependencies = [
  "serde_bytes",
  "serde_json",
  "serde_repr",
+ "serde_with",
  "thiserror",
  "tokio",
  "torrust-tracker-clock",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ serde_bencode = "0"
 serde_bytes = "0"
 serde_json = { version = "1", features = ["preserve_order"] }
 serde_repr = "0"
+serde_with = { version = "3.9.0", features = ["json"] }
 thiserror = "1"
 tokio = { version = "1", features = ["macros", "net", "rt-multi-thread", "signal", "sync"] }
 torrust-tracker-clock = { version = "3.0.0-alpha.12-develop", path = "packages/clock" }
--- a/migrations/README.md
+++ b/migrations/README.md
@@ -1,0 +1,5 @@
+# Database Migrations
+
+We don't support automatic migrations yet. The tracker creates all the needed tables when it starts. The SQL sentences are hardcoded in each database driver.
+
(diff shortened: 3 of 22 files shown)


josecelano:
ACK c5beff551e51f95e05e36a315350902d3b104153

josecelano:
Hi @da2ce7, this is ready for review.

PR #988, opened 2024-08-01
Overhaul tracker keys: option to disable auth key expiration 

### LINKED ISSUE #978: Overhaul tracker keys: option to disable auth key expiration

### Discussed in https://github.com/torrust/torrust-tracker/discussions/244

<div type='discussions-op-text'>

<sup>Originally posted by **rakema01** May 23, 2022</sup>
Would be a neat thing to have when the focus of the site is to seed for long periods of time. Also especially helpful when you've got too many torrents and having to change the tracker URL to match the key is a pain. Of course, there's the option to change the expiry length to a random huge number but having an option to outright disable expiry checks is a bit cleaner in my opinion. In its place, we could have an option for the key be manually regenerated when needed.</div>

I think we could add a config option like this:

```toml
[core]
private = true
check_keys_expiration = true     # The new option. It only makes sense when `private = true`
```

When `check_keys_expiration = false` we don't call `auth::verify(key)` in this function:

```rust
pub async fn verify_auth_key(&self, key: &Key) -> Result<(), auth::Error> {
    // code-review: this function is public only because it's used in a test.
    // We should change the test and make it private.
    match self.keys.read().await.get(key) {
        None => Err(auth::Error::UnableToReadKey {
            location: Location::caller(),
            key: Box::new(key.clone()),
        }),
        Some(key) => auth::verify(key),
    }
}
```

cc @da2ce7

josecelano:
When the tracker is running in private mode you can disable checking keys' expiration in the configuration with:

```toml
[core]
private = false

[core.private_mode]
check_keys_expiration = false
```

All keys will be valid as long as they exist in the database.

### DIFF
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -5,6 +5,7 @@
 //!
 //! The current version for configuration is [`v2`].
 pub mod v2;
+pub mod validator;
 
 use std::collections::HashMap;
 use std::env;
--- a/packages/configuration/src/v2/core.rs
+++ b/packages/configuration/src/v2/core.rs
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 
 use super::network::Network;
 use crate::v2::database::Database;
+use crate::validator::{SemanticValidationError, Validator};
 use crate::{AnnouncePolicy, TrackerPolicy};
 
 #[allow(clippy::struct_excessive_bools)]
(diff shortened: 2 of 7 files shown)


josecelano:
ACK ed40e4ad85f7c5ba3bbe986d8a19eaf1e3740d88

josecelano:
ACK d7dfc3bea813cec6b45aa9c6d31bfa129766b2a0

PR #991, opened 2024-08-02
Config overhaul: use an explicit section for metadata

### LINKED ISSUE #958: Config overhaul: use an explicit section for metadata

After a discussion today, we decided to make the "namespace" in the configuration explicit by using the project name or package name. The main benefit is you can parse the file to know which service the configuration belongs to.

### Current version

```toml
version = "2.0.0"

# ...
```

### Proposal 1

```toml
[torrust_tracker]
version = "2.0.0"

# ...
```

### Proposal 2

```toml
[project]
name = "torrust_tracker"
version = "2.0.0"

# ...
```

Hi @da2ce7, I came up with this solution after seeing this [Python feature](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/).

I think it's better to move the namespace value to an option value. The type in Rust would be the same for all Torrust projects.

### Proposal 3

```toml
[metadata]
name = "torrust_tracker"
version = "2.0.0"

# ...
```

I also think that maybe `version` is too ambiguous if we use the "project" keyword because you don't know if it's the app version of the config file version because there is no native way to add metadata in TOMl files. In our case the config version does not match the app version.

josecelano:
The metadata section in the configuration file is changed:

TOML:

```toml
[metadata]
app = "torrust-tracker"
purpose = "configuration"
schema_version = "2.0.0"
```

JSON:

```json
{
  "metadata": {
    "app": "torrust-tracker",
    "purpose": "configuration",
    "version": "2.0.0"
  }
}
```

- `app`: the application this config file is used for.
- `purpose`: the purpose of the file containing these metadata.
- `schema_version`: the schema version for the file being parsed.


josecelano:
ACK 287e48422003e2e1b890dd2de9cb5ce102601d8c

PR #992, opened 2024-08-02
Config overhaul: make some fields mandatory

### LINKED ISSUE #938: Config overhaul: make some fields mandatory
Depends on: https://github.com/torrust/torrust-tracker/issues/936

In the current version, all config values have a default value. You can run the app without providing any config value at all.

@da2ce7 suggested forcing the system admin (who is responsible to setup the app) to provide at least some critical options like:

```toml
version = "2"

[logging]
threshold = "info"

[core]
private = false
listed = false
```

They are already mandatory, but we have defined server default values, meaning you can omit the file in the TOML or JSON file, and it will use the default value. We have to remove the default values to force the users to provide the values.

Hi @da2ce7, is that list OK? (I'm already using new names even if, in some cases, those changes might not been applied yet)

**UPDATE (2024-08-02)**

Metadata section has changed.

```toml
[metadata]
app = "torrust-tracker"
purpose = "configuration"
schema_version = "2.0.0"

[logging]
threshold = "info"

[core]
private = false
listed = false
```

josecelano:
Some configuration options are mandatory. The tracker will panic if the user doesn't provide an explicit value for them from one of the configuration sources: TOML or ENV VARS.

The mandatory options are:

```toml
[metadata]
schema_version = "2.0.0"

[logging]
threshold = "info"

[core]
private = false
listed = false
```

### DIFF
--- a/packages/configuration/src/lib.rs
+++ b/packages/configuration/src/lib.rs
@@ -303,6 +303,9 @@ pub enum Error {
 
     #[error("Unsupported configuration version: {version}")]
     UnsupportedVersion { version: Version },
+
+    #[error("Missing mandatory configuration option. Option path: {path}")]
+    MissingMandatoryOption { path: String },
 }
 
 impl From<figment::Error> for Error {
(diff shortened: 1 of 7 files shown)


josecelano:
ACK 54d961c687f208b7f6bc2d7b0bc16442116d9eb6

josecelano:
ACK 90ef14d2d5450421fd2b59b78e09536ed30152e7

PR #1001, opened 2024-08-08
fix: benchmarking config template

josecelano:
Fix benchmarking config template:

- Missing [core.tracker_policy] section. The configuration was changed.
- Missing database config. Needed when the default DB folder does not exist. With this config, a storage folder is not needed. We are not using DB, which requires DB for benchmarking.


josecelano:
ACK e563bfbcd0b25004ce13bfefd448656606380c00

PR #1003, opened 2024-08-08
Fix bug: the tracker should always remove inactive peers

### LINKED ISSUE #1002: Bug? should we always remove inactive peers?
I was reviewing this function:

```rust
/// Remove inactive peers and (optionally) peerless torrents
///
/// # Context: Tracker
pub fn cleanup_torrents(&self) {
    // If we don't need to remove torrents we will use the faster iter
    if self.config.tracker_policy.remove_peerless_torrents {
        self.torrents.remove_peerless_torrents(&self.config.tracker_policy);
    } else {
        let current_cutoff =
            CurrentClock::now_sub(&Duration::from_secs(u64::from(self.config.tracker_policy.max_peer_timeout)))
                .unwrap_or_default();
        self.torrents.remove_inactive_peers(current_cutoff);
    }
}
```

If I'm not wrong, inactive peers are not removed unless `remove_peerless_torrents` is enabled.

If fact, the function comment does not match the code, `remove_inactive_peers` function is only called when `remove_peerless_torrents` is enabled.

Should not we always remove inactive peers whether we want to remove peerless torrents or not?

`remove_peerless_torrents` option depends on removing inactive peers. I think there is no way to remove a peer from a torrent but with the `remove_inactive_peers` function.

I guess the function should be:

```rust
/// Remove inactive peers and (optionally) peerless torrents
///
/// # Context: Tracker
pub fn cleanup_torrents(&self) {
    let current_cutoff =
        CurrentClock::now_sub(&Duration::from_secs(u64::from(self.config.tracker_policy.max_peer_timeout)))
            .unwrap_or_default();
    self.torrents.remove_inactive_peers(current_cutoff);

    if self.config.tracker_policy.remove_peerless_torrents {
        self.torrents.remove_peerless_torrents(&self.config.tracker_policy);
    }
}
```

On the other hand, the name for the config option `remove_peerless_torrents` is a little bit misleading. The torrent is not removed even if **it does not have peers** but the stats are been persisted and that torrent has been downloaded at least once. It's an exception.

```rust
fn is_good(&self, policy: &TrackerPolicy) -> bool {
    if policy.persistent_torrent_completed_stat && self.downloaded > 0 {
        return true;
    }

    if policy.remove_peerless_torrents && self.swarm.is_empty() {
        return false;
    }

    true
}
```

cc @da2ce7

josecelano:
**Fix bug:** the tracker should always remove inactive peers even when the config option `core.tracker_policy.remove_peerless_torrents` is disabled.

**Refactor:** rename the `is_good` function to `meets_retaining_policy` to be more specific.

### DIFF
--- a/packages/torrent-repository/src/entry/mod.rs
+++ b/packages/torrent-repository/src/entry/mod.rs
@@ -22,7 +22,7 @@ pub trait Entry {
     fn get_swarm_metadata(&self) -> SwarmMetadata;
 
     /// Returns True if Still a Valid Entry according to the Tracker Policy
-    fn is_good(&self, policy: &TrackerPolicy) -> bool;
+    fn meets_retaining_policy(&self, policy: &TrackerPolicy) -> bool;
 
     /// Returns True if the Peers is Empty
     fn peers_is_empty(&self) -> bool;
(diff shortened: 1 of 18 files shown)


josecelano:
ACK 3fbab31f741475596fcedd1527804efa72b02ead

PR #1022, opened 2024-08-25
Improve Tracing Part One

da2ce7:
make better use of the modern features in the `tracing` library, first part.


da2ce7:
ACK 14e672ec35bfd627d1829bec70a6643f417797e3

PR #1035, opened 2024-09-10
Allow UDP clients to limit peers in response

### LINKED ISSUE #569: The tracker is ignoring the `num_want` param in the announce request.
It seems the tracker is also ignoring the `num_want` param in the announce request.

`num_want` is the number of peers the client wants in the `announce` response (peer list).

See https://www.bittorrent.org/beps/bep_0015.html

Right now:

- We always return less than 74.
- We ignore the `num_want`.

_Originally posted by @josecelano in https://github.com/torrust/torrust-tracker/issues/262#issuecomment-1499050298_

josecelano:
The UDP tracker announce response always include all peers available up to a maxium of 74 peers, ignoring the `num_want` param in the request described in:

https://www.bittorrent.org/beps/bep_0015.html

This change applies that limit only when is lower than then TORRENT_PEERS_LIMIT (74).

It also adds the `numwant` GET param to the announce request for the HTTP tracker for the same purpose.

- [x] UDP tracker (`num_want`, positional param in the UDP packet).
- [x] HTTP tracker (`numwant` GET param).

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -448,6 +448,7 @@ pub mod torrent;
 
 pub mod peer_tests;
 
+use std::cmp::max;
 use std::collections::HashMap;
 use std::net::IpAddr;
 use std::panic::Location;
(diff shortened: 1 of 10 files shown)


josecelano:
ACK 084879e89f4632bd395dbca35460c08c03abdae9

PR #1062, opened 2024-10-04
docs: update release process

josecelano:
I've updated the release process. The process is the same, so I didn't bump the version. I just added a new check because I didn't realise in the recent releases that the deployment workflow failed.

I've also fixed markdown linter errors.


josecelano:
ACK c9f4dfdc3b0cba23ed508a2273beaf057d2433a5

PR #1095, opened 2024-11-19
udp: improve logging (and vairous changes)

da2ce7:
When making this pull request I found that there was logic for the UDP tracker when in private mode: that we do not support, so I removed this dead-code.

I also found that the `ConnectionId` was not being checked for scrape requests, and when parsing badly formatted requests, so I corrected for this.

I also added the statistic counting for how many error responses have been sent.


josecelano (approved):
Gi @da2ce7 I think having the error counter in stats is a good feature. The only way to check errors now is to log into the server and check logs. At least, I can load the API enpoint to get that info. Althougth It only includes only the handled protocol errors.

josecelano:
ACK 8c703955f010948199be75c2dcd3481d519d6bc4

PR #1110, opened 2024-11-27
Add a new job to the testing workflow to test compilation in different OSs

### LINKED ISSUE #1099: Test compilation in different OSs
Relates to: https://github.com/torrust/torrust-tracker/issues/325

We could create a new workflow only for that or change the `testing` workflow,

From:

```yaml
  unit:
    name: Units
    runs-on: ubuntu-latest
    needs: check

    strategy:
      matrix:
        toolchain: [nightly, stable]
```

To:

```yaml
  unit:
    name: Units on ${{ matrix.os }} with toolchain ${{ matrix.rust }}
    runs-on: ${{ matrix.os }}

    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
        toolchain: [nightly, stable]
```

It may be too slow to run all tests in all OSs. We can also add a new job only for building and test compilation, not test execution.

I have to check if it compiles in macOS. Apparently, it's fixed in [Windows](https://github.com/torrust/torrust-tracker/pull/1098).

cc @da2ce7

josecelano:
Add a new job to the testing workflow to test compilation in different operating systems.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -85,6 +85,27 @@ jobs:
         name: Check Unused Dependencies
         run: cargo machete
 
+  build:
+    name: Build on ${{ matrix.os }} (${{ matrix.toolchain }})
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


josecelano:
ACK 38baaea557fc4e045ab9eac7af70fa07543e6ba9

Issue #1140, opened 2024-12-18
Move logic to `http-protocol` and `primitives` packages


josecelano:
Parent issue: https://github.com/torrust/torrust-tracker/issues/753
Relates to: https://github.com/torrust/torrust-tracker/issues/1097#issuecomment-2550953294

[Some doc tests were very slow because they compile the main library](https://github.com/torrust/torrust-tracker/issues/1097).

They were disabled, but the right solution would be to move them to workspace packages so you don't need to compile the main lib.

[Moving that code to packages](https://github.com/torrust/torrust-tracker/issues/753) (pre-existing or new packages) was something [planned anyway](https://github.com/torrust/torrust-tracker/issues/753) because it makes possible to share that code with other projects and compilation should also be faster (you don't need to recompile them if you don't touch anything inside those packages).

New packages (dir and crate name):

- `packages/http-protocol` (`bittorrent-http-protocol`)
- `packages/tracker-core` (`torrust-tracker-core`)

Modules to extract:

To `http-protocol`:

- `src/servers/http/v1/query.rs`
- `src/servers/http/v1/requests/announce.rs`
- `src/servers/http/v1/responses/announce.rs`
- `src/servers/http/v1/responses/error.rs`
- `src/servers/http/v1/responses/scrape.rs`
- `src/servers/http/v1/services/peer_ip_resolver.rs` (I'm not sure about this)

To `tracker-core`:

- `crate::core::AnnounceData`
- `crate::core::ScrapeData`
...

See https://github.com/torrust/torrust-tracker/pull/1141/files

josecelano:
The package `http-protocol` was already created. We only need to move the logic there and extract the new `tracker-core` package.

josecelano:
> The package `http-protocol` was already created. We only need to move the logic there and extract the new `tracker-core` package.

I did not create the `tracker-core` package because I have only moved basic types. I moved them to the `primitives` package.

PR #1141, opened 2024-12-18
Don't execute slow doc tests

### LINKED ISSUE #1097: Some docs tests are very slow
Some doc tests are very slow now.

```console
cargo test --doc
```

These are the ones that take longer:

```output
test src/servers/http/percent_encoding.rs - servers::http::percent_encoding::percent_decode_info_hash (line 28)
test src/servers/http/percent_encoding.rs - servers::http::percent_encoding::percent_decode_peer_id (line 59)
test src/servers/http/v1/query.rs - servers::http::v1::query::Query::get_param (line 33)
test src/servers/http/v1/query.rs - servers::http::v1::query::Query::get_param (line 46)
test src/servers/http/v1/query.rs - servers::http::v1::query::Query::get_param_vec (line 62) 
test src/servers/http/v1/query.rs - servers::http::v1::query::Query::get_param_vec (line 75)
test src/servers/http/v1/requests/announce.rs - servers::http::v1::requests::announce::Announce (line 32)
test src/servers/http/v1/responses/announce.rs - servers::http::v1::responses::announce::CompactPeer (line 207)
test src/servers/http/v1/responses/announce.rs - servers::http::v1::responses::announce::NormalPeer (line 155)
test src/servers/http/v1/responses/error.rs - servers::http::v1::responses::error::Error::write (line 29)
test src/servers/http/v1/responses/scrape.rs - servers::http::v1::responses::scrape::Bencoded (line 14)
test src/servers/http/v1/services/peer_ip_resolver.rs - servers::http::v1::services::peer_ip_resolver::invoke (line 62) 
test src/servers/http/v1/services/peer_ip_resolver.rs - servers::http::v1::services::peer_ip_resolver::invoke (line 84)
```

It can even crash your computer if you don't have enough memory. You can run them with:

```console
cargo test -- --test-threads=1
```

It would be even slower, but you can execute them.

We have to find out why they are so slow now.

I guess the problem is:

> Each doc-test generates a standalone binary for execution. Since these binaries are isolated from each other, they may independently recompile parts of your crate, especially if there is no mechanism to reuse already compiled artifacts.

For example, in this test:

```
/// ```rust
/// use std::str::FromStr;
/// use torrust_tracker::servers::http::percent_encoding::percent_decode_info_hash;
/// use bittorrent_primitives::info_hash::InfoHash;
/// use torrust_tracker_primitives::peer;
///
/// let encoded_infohash = "%3B%24U%04%CF%5F%11%BB%DB%E1%20%1C%EAjk%F4Z%EE%1B%C0";
///
/// let info_hash = percent_decode_info_hash(encoded_infohash).unwrap();
///
/// assert_eq!(
///     info_hash,
///     InfoHash::from_str("3b245504cf5f11bbdbe1201cea6a6bf45aee1bc0").unwrap()
/// );
/// ```
```

We need to compile the main `torrust_tracker` package. FOr this particular case we can extract the functionality to a new package. In fact, it's duplicated. THe comment for the duplication:

```
/* code-review: this module is duplicated in torrust_tracker::servers::http::percent_encoding.
   Should we move it to torrust_tracker_primitives?
*/
```

But in other cases, we can also do the same. If, for some reason, we cannot do it, I would not compile the example for these slow cases.

cc @mario-nt @da2ce7

josecelano:
This PR disables slow doc tests. They are slow because they include the main library, and each test has to compile it.

They will be enabled again after [moving the code to workspace sub-packages](https://github.com/torrust/torrust-tracker/issues/1140) that do not depend on the main tracker lib.

### DIFF
--- a/src/core/auth.rs
+++ b/src/core/auth.rs
@@ -196,7 +196,7 @@ impl Key {
 
 /// Error returned when a key cannot be parsed from a string.
 ///
-/// ```rust,no_run
+/// ```text
 /// use torrust_tracker::core::auth::Key;
 /// use std::str::FromStr;
 ///
--- a/src/core/databases/driver.rs
+++ b/src/core/databases/driver.rs
@@ -29,7 +29,7 @@ pub enum Driver {
 ///
 /// Example for `SQLite3`:
 ///
-/// ```rust,no_run
+/// ```text
 /// use torrust_tracker::core::databases;
 /// use torrust_tracker::core::databases::driver::Driver;
 ///
@@ -40,7 +40,7 @@ pub enum Driver {
 ///
 /// Example for `MySQL`:
 ///
-/// ```rust,no_run
+/// ```text
 /// use torrust_tracker::core::databases;
 /// use torrust_tracker::core::databases::driver::Driver;
 ///
--- a/src/servers/http/v1/query.rs
+++ b/src/servers/http/v1/query.rs
@@ -30,7 +30,7 @@ impl Query {
     /// It return `Some(value)` for a URL query param if the param with the
     /// input `name` exists. For example:
     ///
-    /// ```rust
+    /// ```text
     /// use torrust_tracker::servers::http::v1::query::Query;
     ///
     /// let raw_query = "param1=value1&param2=value2";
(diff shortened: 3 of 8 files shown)


josecelano:
ACK 0e13ff72554507218d2a548b54c4478c1410f48e

Issue #1152, opened 2024-12-23
Log assertions: add assertions for API log errors


josecelano:
Parent issue: https://github.com/torrust/torrust-tracker/issues/1150

Some API endpoints produce log errors:

```
cargo test -- --test-threads=1 --nocapture | grep "ERROR" | grep "API"
```

Write assertions in the tests producing those errors.

Tests expecting a 500 response are tests using these asserts:

```rust
pub async fn assert_token_not_valid(response: Response) {
    assert_unhandled_rejection(response, "token not valid").await;
}

pub async fn assert_unauthorized(response: Response) {
    assert_unhandled_rejection(response, "unauthorized").await;
}

pub async fn assert_failed_to_remove_torrent_from_whitelist(response: Response) {
    assert_unhandled_rejection(response, "failed to remove torrent from whitelist").await;
}

pub async fn assert_failed_to_whitelist_torrent(response: Response) {
    assert_unhandled_rejection(response, "failed to whitelist torrent").await;
}

pub async fn assert_failed_to_reload_whitelist(response: Response) {
    assert_unhandled_rejection(response, "failed to reload whitelist").await;
}

pub async fn assert_failed_to_generate_key(response: Response) {
    assert_unhandled_rejection(response, "failed to generate key").await;
}

pub async fn assert_failed_to_delete_key(response: Response) {
    assert_unhandled_rejection(response, "failed to delete key").await;
}

pub async fn assert_failed_to_reload_keys(response: Response) {
    assert_unhandled_rejection(response, "failed to reload keys").await;
}

async fn assert_unhandled_rejection(response: Response, reason: &str) {
    assert_eq!(response.status(), 500);
    assert_eq!(response.headers().get("content-type").unwrap(), "text/plain; charset=utf-8");

    let reason_text = format!("Unhandled rejection: Err {{ reason: \"{reason}");
    let response_text = response.text().await.unwrap();
    assert!(
        response_text.contains(&reason_text),
        ":\n  response: `\"{response_text}\"`\n  does not contain: `\"{reason_text}\"`."
    );
}
```

We only need to identify tests using these assert functions:

```
assert_token_not_valid
assert_unauthorized
assert_failed_to_remove_torrent_from_whitelist
assert_failed_to_whitelist_torrent
assert_failed_to_reload_whitelist
assert_failed_to_generate_key
assert_failed_to_delete_key
assert_failed_to_reload_keys
```

We need to use the request-id in the logs to match the test with the log line.

[...]

Issue #1150, opened 2024-12-23
Write assertions for logged errors


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1069
Implement after: https://github.com/torrust/torrust-tracker/issues/1146

Some tests produce errors in logs. There are expected errors:

```console
cargo test -- --nocapture | grep "ERROR"
```

We want to add assertions for them. 

After merging this [PR](https://github.com/torrust/torrust-tracker/pull/1149), we will be able to capture logs in tests.

This issue is better implemented after [this one](https://github.com/torrust/torrust-tracker/issues/1146).

### Subtasks

- [x] https://github.com/torrust/torrust-tracker/issues/1152
- [x] https://github.com/torrust/torrust-tracker/issues/1162
- [x] https://github.com/torrust/torrust-tracker/issues/1164

### Errors in test execution output

These are the three types of errors:

**Tracker API:**

The tracker API returns a 500 response on some endpoints, in spite of not being a server error.

```output
2024-12-23T16:17:24.085791Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.097523Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.112349Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.134428Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.143296Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.157824Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.165816Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.183568Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.185476Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.187261Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
2024-12-23T16:17:24.237987Z ERROR API: response failed classification=Status code: 500 Internal Server Error latency=0 ms
[...]

PR #1153, opened 2024-12-23
Fix missing logs for HTTP requests

josecelano:
Relates to:
  - https://github.com/torrust/torrust-tracker/issues/1150
  - https://github.com/torrust/torrust-tracker/issues/1152

Some log lines (tracing spans) were missing. In addition to that, we were using new spans to log requests and responses instead of using the span provided by the `TraceLayer`. That enables you to join all the log events related to the same request.

This fix is needed to continue with this issue: https://github.com/torrust/torrust-tracker/issues/1150

Because it allows the use of a "request-id" header to identify logs.

We can write log assertions by matching lines with the `request-id` used in the request.

For example, if you make this request:

```console
curl -H "x-request-id: YOUR_REQUEST_ID" http://0.0.0.0:1212/api/v1/stats?token=InvalidToken
```

This is the new output (which was missing before this change):

```output
2024-12-23T17:53:06.530704Z  INFO request{method=GET uri=/api/v1/stats?token=InvalidToken version=HTTP/1.1}: API: request method=GET uri=/api/v1/stats?token=InvalidToken request_id=YOUR_REQUEST_ID
2024-12-23T17:53:06.530777Z ERROR request{method=GET uri=/api/v1/stats?token=InvalidToken version=HTTP/1.1}: API: response latency_ms=0 status_code=500 Internal Server Error request_id=YOUR_REQUEST_ID
2024-12-23T17:53:06.530785Z ERROR request{method=GET uri=/api/v1/stats?token=InvalidToken version=HTTP/1.1}: API: response failed failure_classification=Status code: 500 Internal Server Error latency=0 ms
```

As you can see, now we have the "request_id=YOUR_REQUEST_ID" field, which can be used to identify the test that made the request and write a log assertion for that concrete expected line.

### DIFF
--- a/src/servers/apis/routes.rs
+++ b/src/servers/apis/routes.rs
@@ -53,7 +53,7 @@ pub fn router(tracker: Arc<Tracker>, access_tokens: Arc<AccessTokens>) -> Router
         .layer(
             TraceLayer::new_for_http()
                 .make_span_with(DefaultMakeSpan::new().level(Level::INFO))
-                .on_request(|request: &Request<axum::body::Body>, _span: &Span| {
+                .on_request(|request: &Request<axum::body::Body>, span: &Span| {
                     let method = request.method().to_string();
                     let uri = request.uri().to_string();
                     let request_id = request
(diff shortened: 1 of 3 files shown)


josecelano:
ACK 97233f57aa9595e05bc6d44ad2dd41ae1087e3de

PR #1180, opened 2025-01-11
Refactor: Move logic to `http-protocol` and `primitives` packages

### LINKED ISSUE #1140: Move logic to `http-protocol` and `primitives` packages

josecelano:
Refactor: Move HTTP tracker logic to `http-protocol` package and core tracker types to `primitives` package. This will:

- Allow sharing that code with other projects.
- Allow re-enabling doc tests disabled [here](https://github.com/torrust/torrust-tracker/issues/1097).
- Decouple generic reusable logic from Axum framework.
- Improve compilation times.


josecelano:
ACK c2d134e792216fbc58f9df67d89762434cf7bae2

PR #1186, opened 2025-01-16
Overhaul core Tracker: extract stats

### LINKED ISSUE #1184: Overhaul core Tracker: extract stats

Extract the responsibility for handling stats (send events and return metrics) from the core tracker.

These two methods (`get_stats` and `get_stats`) will be called  directly exposing the `stats_repository` and `stats_event_sender`.

```rust
    /// It return the `Tracker` [`statistics::metrics::Metrics`].
    ///
    /// # Context: Statistics
    pub async fn get_stats(&self) -> tokio::sync::RwLockReadGuard<'_, statistics::metrics::Metrics> {
        self.stats_repository.get_stats().await
    }

    /// It allows to send a statistic events which eventually will be used to update [`statistics::metrics::Metrics`].
    ///
    /// # Context: Statistics
    pub async fn get_stats(
        &self,
        event: statistics::event::Event,
    ) -> Option<Result<(), SendError<statistics::event::Event>>> {
        match &self.stats_event_sender {
            None => None,
            Some(stats_event_sender) => stats_event_sender.send_event(event).await,
        }
    }
```

See the [previous similar refactor](https://github.com/torrust/torrust-tracker/issues/1182) for reference.

Since I'm extracting a lot of code from the core Tracker now the app bootstrapping is becoming more complex because the tracker dependencies are built before instantiating the tracker. At some point I think I will create a [IoC](https://en.wikipedia.org/wiki/Inversion_of_control) container to store all the references to services (Arc<...>).

cc @da2ce7

josecelano:
Overhaul core Tracker: extract stats

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -29,6 +29,8 @@ use torrust_tracker_configuration::Configuration;
 use tracing::instrument;
 
 use crate::bootstrap::jobs::{health_check_api, http_tracker, torrent_cleanup, tracker_apis, udp_tracker};
+use crate::core::statistics::event::sender::Sender;
+use crate::core::statistics::repository::Repository;
 use crate::servers::registar::Registar;
 use crate::servers::udp::server::banning::BanService;
 use crate::{core, servers};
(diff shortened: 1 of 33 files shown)


josecelano:
ACK 2a0bc4789fb13cd3e69c88d3cabf30642f6fb2d6

PR #1196, opened 2025-01-21
Overhaul core Tracker: refactor authentication module

### LINKED ISSUE #1195: Overhaul core Tracker: refactor authentication module

I have extracted an `authentication` service from the core tracker [here](https://github.com/torrust/torrust-tracker/pull/1192). However the authentication service is still used via the tracker public field `authentication`.

```rust
pub struct Tracker {
    /// The tracker configuration.
    config: Core,

    /// A database driver implementation: [`Sqlite3`](crate::core::databases::sqlite)
    /// or [`MySQL`](crate::core::databases::mysql)
    database: Arc<Box<dyn Database>>,

    /// The service to check if a torrent is whitelisted.
    pub whitelist_authorization: Arc<whitelist::authorization::Authorization>,

    /// The in-memory torrents repository.
    torrents: Arc<Torrents>,

    /// The service to authenticate peers.
    pub authentication: Arc<authentication::Facade>, // <--- This is the new authentication service
}
```

I called it `Facade` because I want to split the authentication service into more services. It still has many responsibilities.

I will implement that refactor in this issue. After this refactor I will be able to remove the `authentication` field from the tracker and use the new services directly in other parts of the code like Axum handlers.

At the end, what I expect to have it's something similar to what I did for the [whitelist](https://github.com/torrust/torrust-tracker/tree/develop/src/core/whitelist) service. I expect to have:

- A type to handle keys in database (database repository for keys).
- A type to handle keys in memory (in-memory repository for keys).
- A type to do some tasks like loading keys from database into memory, ... (keys manager).
- A type to do the authentication (verify key). I want to separate this behavior because I think is big enough to be encapsulated in its own type and also to be able to mock it in tests.

josecelano:
Overhaul core Tracker: refactor authentication module.


josecelano:
ACK dff6bca1bbf587074ad31ddd8a73f3c4c15ba058

PR #1225, opened 2025-01-29
Overhaul core Tracker: extract workspace package `tracker-core`

### LINKED ISSUE #1223: Overhaul core Tracker: extract workspace package for the tracker core lib

Create a new workspace package for the core tracker module: `src/core`.

- Package folder: `./packages/tracker-core-lib`
- Crate name: `bittorrent-tracker-core-lib`

This would allow other projects to use the same core tracker logic and I would make harder to pollute the core tracker with higher level login like web frameworks, etc.

Database migration are only used inside the core tracker module, so they have to be moved into the new package.

cc @da2ce7

josecelano:
Overhaul core Tracker: extract workspace package `tracker-core`.


josecelano:
ACK a5ca24460c5b06ad434f65f2e67f3aac47260b23

Issue #1251, opened 2025-02-07
Overhaul core Tracker: add tests for `databases` mod


josecelano:
Parent issue: https://github.com/torrust/torrust-tracker/issues/1226

Overhaul core Tracker: add tests for `databases` mod.

PR #1254, opened 2025-02-07
Some minor changes before adding unit tests to the tracker core `databases` module.

josecelano:
Some minor changes before [adding unit tests to the tracker core `databases` module](https://github.com/torrust/torrust-tracker/issues/1251).

- [x] Reorganize the module.
- [x] Remove the the `new` method from the `Database` trait.

I have removed the `new` method from the `Database` trait because:

- Drivers' dependencies might be different in the future for different drivers.
- You can have conflicts when mocking the trait. See https://github.com/asomers/mockall/commit/7c54ed1999c4c44cbc0c71701d5293e781c7d7a9
- The constructor is not part of the trait behavior in this case.

### DIFF
--- a/packages/tracker-core/src/databases/driver/mod.rs
+++ b/packages/tracker-core/src/databases/driver/mod.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 use sqlite::Sqlite;
 
 use super::error::Error;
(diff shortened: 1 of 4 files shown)


josecelano:
ACK 3f78459208aec0945b0671a92ebc2b31c6569e32

PR #1274, opened 2025-02-17
Overhaul core Tracker: review whitelist functionality (part 2)

### LINKED ISSUE #1270: Overhaul core Tracker: review whitelist functionality (part 2)
Parent issue: https://github.com/torrust/torrust-tracker/issues/1261
Relates to: 
 - https://github.com/torrust/torrust-tracker/pull/1269
 - https://github.com/torrust/torrust-tracker/issues/1268

This is second part of the implementation of this [issue](https://github.com/torrust/torrust-tracker/issues/1268).

I decided to merge the [PR](https://github.com/torrust/torrust-tracker/pull/1269) for the issue #1268 before continuing working of the refactor.

### Sub-tasks

- [x] In the tracker-core announce handler return a `Result<AnnounceData, AnnounceError>` when the torrent is not included in the whitelist.
- [x] In the tracker-core scrape handler return a `Result<ScrapeData, ScrapeError>` so we are able to return errors in the future without breaking the public API.
- [x] Move `packages::udp_tracker_core::peer_builder::from_request` to new package `udp-protocol`
- [x] Inline `invoke` function in handlers when possible.
- [x] ~~Move tests for the handlers (announce and scrape) from HTTP server to `http_tracker_core`~~. *1
- [x] ~~Move tests for the handlers (announce and scrape) from UDP server to `udp_tracker_core`~~. *1

**NOTE*1**: instead of moving the tests I will create new ones for the `http_tracker_core` after [refactoring all packages](https://github.com/torrust/torrust-tracker/issues/753). The tests contain logic from the delivery layer that it will not be moved to the `http_tracker_core` so It will be harder to refactor those tests than created new one for the logic that we will finally move to the `http_tracker_core`.

josecelano:
Overhaul core Tracker: review whitelist functionality (part 2).

- [x] Return an error in the announce handler if the tracker is running in `listed` mode and the infohash is not whitelisted. This is already implemented in the higher delivery layer but it should be also validated in the tracker core (domain layer).
- [x] Change scrape handler signature to return an error even if it's not needed now. That will allow us to return an error in the future whiteout breaking changes.
- [x] Move `packages::udp_tracker_core::peer_builder::from_request` to new package `udp-protocol`
- [x] Inline `invoke` function in handlers (`http-tracker-core` and `udp-tracker-core`) when possible.


josecelano:
ACK d0e693619f2322dde1866b20a4a66d453bde7504

PR #1305, opened 2025-02-20
docs: remove code-review comment

### LINKED ISSUE #1276: Overhaul core Tracker: review announce handler (peer mutability)
In the tracer core the announce handler requires the peer argument to be mutable. That's because the peer IP might change because the tracker ignores the IP in the request and assign the IP from the connection.

Some time ago I wrote a comment to review that. I'm considering making the peer immutable and returning the assigned peer IP. The function I want to change:

```rust
pub async fn announce(
    &self,
    info_hash: &InfoHash,
    peer: &mut peer::Peer,
    remote_client_ip: &IpAddr,
    peers_wanted: &PeersWanted,
) -> Result<AnnounceData, AnnounceError> {
    // code-review: maybe instead of mutating the peer we could just return
    // a tuple with the new peer and the announce data: (Peer, AnnounceData).
    // It could even be a different struct: `StoredPeer` or `PublicPeer`.

    self.whitelist_authorization.authorize(info_hash).await?;

    tracing::debug!("Before: {peer:?}");
    tracing::debug!("After: {peer:?}");

    let stats = self.upsert_peer_and_get_stats(info_hash, peer);

    let peers = self
        .in_memory_torrent_repository
        .get_peers_for(info_hash, peer, peers_wanted.limit());

    Ok(AnnounceData {
        peers,
        stats,
        policy: self.config.announce_policy,
    })
}
```

The new signature:

```rust
pub async fn announce(
    &self,
    info_hash: &InfoHash,
    peer: &peer::Peer,
    remote_client_ip: &IpAddr,
    peers_wanted: &PeersWanted,
) -> Result<(AnnounceData, IpAddr), AnnounceError> {}
```

I think (I have to confirm it) I made it mutable to return the final IP because it's used for testing purposes.

What do you think @da2ce7?

josecelano:
We decided not to change it.

- The returned value is simpler.
- We force the initial peer to change so there is no confusion about what was the final announced peer.

### DIFF
--- a/packages/tracker-core/src/announce_handler.rs
+++ b/packages/tracker-core/src/announce_handler.rs
@@ -162,10 +162,6 @@ impl AnnounceHandler {
         remote_client_ip: &IpAddr,
         peers_wanted: &PeersWanted,
     ) -> Result<AnnounceData, AnnounceError> {
-        // code-review: maybe instead of mutating the peer we could just return
-        // a tuple with the new peer and the announce data: (Peer, AnnounceData).
-        // It could even be a different struct: `StoredPeer` or `PublicPeer`.
-
         self.whitelist_authorization.authorize(info_hash).await?;
 
         tracing::debug!("Before: {peer:?}");


josecelano:
ACK bc95fc4c841674d837064099fedf040e059a20b8

PR #1308, opened 2025-02-20
Remove explicit declaration of workspace members from `Cargo.toml`

### LINKED ISSUE #1303: Refactor packages: remove explicit declaration of workspace members from `Cargo.toml`
I've been extracting new packages into the [packages](https://github.com/torrust/torrust-tracker/tree/develop/packages) folder but I forgot to add them the the `[workspace]` section in the main package [Cargo.toml](https://github.com/torrust/torrust-tracker/blob/develop/Cargo.toml) file.

```toml
[workspace]
members = [
    "console/tracker-client",
    "contrib/bencode",
    "packages/configuration",
    "packages/located-error",
    "packages/primitives",
    "packages/test-helpers",
    "packages/torrent-repository",
    "packages/tracker-api-client",
    "packages/tracker-client",
    "packages/tracker-core",
]
```

The application works anyway becuase _all [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members_.

See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

So what I'm going to do is to remove the whole section.

cc @da2ce7 

### Cargo docs

From the [cargo docs](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html):

_Cargo offers a feature called workspaces that can help manage multiple related packages that are developed in tandem._

_A workspace is a set of packages that share the same Cargo.lock and output directory._

_The workspace has one target directory at the top level that the compiled artifacts will be placed into; the adder package doesn’t have its own target directory. Even if we were to run cargo build from inside the adder directory, the compiled artifacts would still end up in add/target rather than add/adder/target_

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

_All [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members. Additional members can be listed with the members key, which should be an array of strings containing directories with Cargo.toml files._

josecelano:
The application works anyway because _all [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members_.

See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,18 +108,7 @@ torrust-tracker-api-client = { version = "3.0.0-develop", path = "packages/track
 torrust-tracker-test-helpers = { version = "3.0.0-develop", path = "packages/test-helpers" }
 
 [workspace]
-members = [
-    "console/tracker-client",
-    "contrib/bencode",
-    "packages/configuration",
-    "packages/located-error",
-    "packages/primitives",
-    "packages/test-helpers",
-    "packages/torrent-repository",
-    "packages/tracker-api-client",
-    "packages/tracker-client",
-    "packages/tracker-core",
-]
+members = ["console/tracker-client"]
 
 [profile.dev]
 debug = 1


josecelano:
ACK fc48f9a97041c37dd0c76deaaf3521d6ec05567f

josecelano:
ACK d48272f53fad95240660657c31b2ad47b5c38ea4

PR #1325, opened 2025-02-25
Refactor packages: normalize HTTP tracker core fn signatures for announce and scrape handlers

### LINKED ISSUE #1318: Refactor packages: normalize HTTP tracker core fn signatures for announce and scrape handlers
The application has these layers for the HTTP tracker:

```
        `axum-http-tracker-server`
                     ↓
`http-tracker-core` and `tracker-core`
         ↓
  `http-protocol`
```

And these layers for the UDP tracker:

```
        `axum-udp-tracker-server`
                     ↓
`udp-tracker-core` and `tracker-core`
         ↓
  `udp-protocol`
```

These are the announce and scrape handlers signatures (omitting parameters):

```
HTTP (bittorrent_http_tracker_core)

pub async fn handle_announce(...) -> Result<torrust_tracker_primitives::core::AnnounceData, bittorrent_http_tracker_protocol::v1::responses::error::Error>

pub async fn handle_scrape(...)   -> Result<torrust_tracker_primitives::core::ScrapeData,   bittorrent_http_tracker_protocol::v1::responses::error::Error>

UDP (bittorrent_udp_tracker_core)

pub async fn handle_announce(...) -> Result<torrust_tracker_primitives::core::AnnounceData, bittorrent_udp_tracker_core::services::announce::UdpAnnounceError>

pub async fn handle_scrape(...)   -> Result<torrust_tracker_primitives::core::ScrapeData,   bittorrent_udp_tracker_core::services::scrape::UdpScrapeError>

pub async fn handle_connect(...)  -> aquatic_udp_protocol::common::ConnectionId 
```

As you can see there is an smell in the way the HTTP tracker returns errors when the announce or scrape requests fail. All the functions return a service-level error (that might wrap the tracker-core error). However, the HTTP tracker directly returns a HTTP response. I think we should do the same as in other cases. We should return an error and let the upper layer (server) build the response.

That's a convention we have followed. Responses are built at the *-server package level.

So I'm proposing this change:

```
HTTP (bittorrent_http_tracker_core)

pub async fn handle_announce(...) -> Result<torrust_tracker_primitives::core::AnnounceData, bittorrent_http_tracker_core::services::announce::HttpAnnounceError>

pub async fn handle_scrape(...)   -> Result<torrust_tracker_primitives::core::ScrapeData,   bittorrent_http_tracker_core::services::announce::HttpScrapeError>
````

Alternatively we could return responses in all functions but I think it's better to return a domain type and let the server build the response. That would give the opportunity to the server to customize the response without re-editing it.

What do you think @da2ce7?

josecelano:
Refactor packages: normalize HTTP tracker core fn signatures for announce and scrape handlers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -563,6 +563,7 @@ dependencies = [
  "bittorrent-tracker-core",
  "futures",
  "mockall",
+ "thiserror 2.0.11",
  "tokio",
  "torrust-tracker-configuration",
  "torrust-tracker-primitives",
--- a/packages/axum-http-tracker-server/src/v1/handlers/announce.rs
+++ b/packages/axum-http-tracker-server/src/v1/handlers/announce.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use aquatic_udp_protocol::AnnounceEvent;
 use axum::extract::State;
 use axum::response::{IntoResponse, Response};
+use bittorrent_http_tracker_core::services::announce::HttpAnnounceError;
 use bittorrent_http_tracker_protocol::v1::requests::announce::{Announce, Compact, Event};
 use bittorrent_http_tracker_protocol::v1::responses::{self};
 use bittorrent_http_tracker_protocol::v1::services::peer_ip_resolver::ClientIpSources;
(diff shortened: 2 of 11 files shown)


josecelano:
ACK df4f7827e5b99ad3ad1e3fe6498850dce4c7731a

PR #1328, opened 2025-02-26
Refactor packages: review UDP events

### LINKED ISSUE #1319: Refactor packages: review UDP events
These are the events used for statistics:

```
HTTP core (http-tracker-core)

- Tcp4Announce  Tcp6Announce
- Tcp4Scrape    Tcp6Scrape
```

```
UDP core (udp-tracker-core)

- Udp4Announce  Udp6Announce
- Udp4Scrape    Udp6Scrape
- Udp4Connect   Udp6Connect

- Udp4Request   Udp6Request
- Udp4Response  Udp6Response

- Udp4Error     Udp6Error
- UdpRequestAborted
- UdpRequestBanned
```

And this is where those events are used::

```
HTTP core (http-tracker-core)

- Tcp4Announce  Tcp6Announce
- Tcp4Scrape    Tcp6Scrape
```

```
UDP core (udp-tracker-core)

- Udp4Announce  Udp6Announce
- Udp4Scrape    Udp6Scrape
- Udp4Connect   Udp6Connect

UDP server (udp-tracker-server)

- Udp4Request   Udp6Request
- Udp4Response  Udp6Response

- UdpRequestAborted
- UdpRequestBanned,

- Udp4Error     Udp6Error
```

Maybe we should split UDP events into:

- UDP tracker core events
- UDP tracker server events

UDP tracker core events would be the same as in the HTTP tracker core and the extra connect events that doesn't exist in the HTTP tracker.

For me those events are produced at different layers and it does not make sense to add events to the UDP tracker code that only make sense for the UDP server.

My proposal is to split them and move the UDP server events to the UDP server package (udp-tracker-server).

If we want to keep them in the UDP tracker core we should find the way to send them at the core level. For example, there is an special case, the `Udp4Error` and `Udp6Error` errors are sent when the server sends an error response. We could send them when the tracker core returns an error from the announce and scrape requests. However that would change a little bit the meaning of the event:

- Udp4Error and Udp6Error sent in the UDP tracker core: a request produced an error.
- Udp4Error and Udp6Error sent in the UDP tracker server: the server sent a response error.

Since we don't have events in the HTTP tracker core for this case I would keep those errors in the server for now.

I will not change drastically events in this issue. I just one to move events to the right layers (package), however I think we need to make a big refactor. For example, for the HTTP tracker:

```rust
pub enum Event {
    Tcp4Announce,
    Tcp4Scrape,
    Tcp6Announce,
    Tcp6Scrape,
}
```

I would do this:

```rust
pub enum Event {
    TcpAnnounce { scheme: HTTPorUDP, ip_version: V4orV6 },
    TcpScrape { scheme: HTTPorUDP, ip_version: V4orV6 },
}
```

That is out of the scope of this issue.

cc @da2ce7

josecelano:
This splits the UDP stats events into two different types of events (one for each layer):

- UDP core events
- UDP server event

### Subtasks

- [x] Step 1. Create UDP server events.
- [x] Step 2. Remove UDP server events from core events.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4450,6 +4450,7 @@ dependencies = [
  "torrust-tracker-configuration",
  "torrust-tracker-primitives",
  "torrust-tracker-test-helpers",
+ "torrust-udp-tracker-server",
  "tower 0.5.2",
  "tower-http",
  "tracing",
@@ -4525,6 +4526,7 @@ dependencies = [
  "torrust-tracker-configuration",
  "torrust-tracker-primitives",
  "torrust-tracker-test-helpers",
+ "torrust-udp-tracker-server",
 ]
 
 [[package]]
--- a/packages/axum-tracker-api-server/Cargo.toml
+++ b/packages/axum-tracker-api-server/Cargo.toml
@@ -38,6 +38,7 @@ torrust-tracker-api-core = { version = "3.0.0-develop", path = "../tracker-api-c
 torrust-tracker-clock = { version = "3.0.0-develop", path = "../clock" }
 torrust-tracker-configuration = { version = "3.0.0-develop", path = "../configuration" }
 torrust-tracker-primitives = { version = "3.0.0-develop", path = "../primitives" }
+torrust-udp-tracker-server = { version = "3.0.0-develop", path = "../udp-tracker-server" }
 tower = { version = "0", features = ["timeout"] }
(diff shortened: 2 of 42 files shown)


josecelano:
ACK b8d2f762a4a7b23bebe70f31f4dca172b022e9a0

PR #1331, opened 2025-02-27
Refactor packages:  review `AnnounceEvent` struct

### LINKED ISSUE #1309: Refactor packages:  review `AnnounceEvent` struct
We have an enum for announce events (`AnnounceEvent`) in the HTTP tracker:

```rust
/// The event that the peer is reporting: `started`, `completed` or `stopped`.
///
/// If the event is not present or empty that means that the peer is just
/// updating its status. It's one of the announcements done at regular intervals.
///
/// Refer to [BEP 03. The `BitTorrent Protocol` Specification](https://www.bittorrent.org/beps/bep_0003.html)
/// for more information.
#[derive(PartialEq, Debug)]
pub enum Event {
    /// Event sent when a download first begins.
    Started,
    /// Event sent when the downloader cease downloading.
    Stopped,
    /// Event sent when the download is complete.
    /// No `completed` is sent if the file was complete when started
    Completed,
}
```

That's the only place where we use our own type to represent announce events. In the rest of the code we are using the aquatic type for [UPD announce events](https://github.com/greatest-ape/aquatic/blob/master/crates/udp_protocol/src/request.rs#L218).

If I'm not wrong we had another type (domain type) that was removed in favor of using the one defined in the [aquatic UDP protocol](https://github.com/greatest-ape/aquatic/blob/master/crates/udp_protocol/src/request.rs#L218).

There are still some parts of the code that still reflect the previous state. For example:

```rust
#[must_use]
pub fn map_to_aquatic_event(event: &Option<Event>) -> aquatic_udp_protocol::AnnounceEvent {
    match event {
        Some(event) => match &event {
            Event::Started => aquatic_udp_protocol::AnnounceEvent::Started,
            Event::Stopped => aquatic_udp_protocol::AnnounceEvent::Stopped,
            Event::Completed => aquatic_udp_protocol::AnnounceEvent::Completed,
        },
        None => aquatic_udp_protocol::AnnounceEvent::None,
    }
}

#[must_use]
pub fn map_to_torrust_event(event: &Option<Event>) -> AnnounceEvent {
    match event {
        Some(event) => match &event {
            Event::Started => AnnounceEvent::Started,
            Event::Stopped => AnnounceEvent::Stopped,
            Event::Completed => AnnounceEvent::Completed,
        },
        None => AnnounceEvent::None,
    }
}
```

The second function is the same as the first one. The function is duplicated. We should remove the second one.

I see a lot of inconsistencies in how we use types to handle announce events. 

1. Mapping between our type and aquatic type is not infallible

```rust
impl From<aquatic_udp_protocol::request::AnnounceEvent> for Event {
    fn from(value: aquatic_udp_protocol::request::AnnounceEvent) -> Self {
        match value {
            AnnounceEvent::Started => Self::Started,
            AnnounceEvent::Stopped => Self::Stopped,
            AnnounceEvent::Completed => Self::Completed,
            AnnounceEvent::None => panic!("can't convert announce event from aquatic for None variant"),
        }
    }
}
```

I have to confirm it, but I think we assume `Started` event if the request does not contains that param as it's indicated in the comment of the `Event` enum.

2. We are using only the aquatic announce event for UDP

Aquatic defines two different enums for announce events:

- One for HTTP: https://github.com/greatest-ape/aquatic/blob/master/crates/http_protocol/src/common.rs#L28
- One for UDP: https://github.com/greatest-ape/aquatic/blob/master/crates/udp_protocol/src/request.rs#L218

UDP:

```rust
#[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
pub enum AnnounceEvent {
    Started,
    Stopped,
    Completed,
    None,
}
```

HTTP:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum AnnounceEvent {
    Started,
    Stopped,
    Completed,
    Empty,
}
```

They don't even have the same variant names.

3. Our tracker core depends on aquatic announce events (and its dependencies)

Our [bittorrent-tracker-core depends on the aquatic UDP protocol package](https://github.com/torrust/torrust-tracker/blob/develop/packages/tracker-core/Cargo.toml#L17). 

I think the tracker core should NOT depend on logic that is supposed to be in upper layers (delivery layers).

### Proposal

I think we should fix those problems by:

1. Using the announce event enum defined in the aquatic HTTP protocol in our HTTP tracker packages (http server and http-core).
2. Using the announce event enum defined in the aquatic UDP protocol in our UDP tracker packages (udp server and udp core).
3. Introducing our own enum for the announce event (and dependencies) in the `packages/primitives/src/core.rs` module and use it in the tracker core. So we have to map the event from delivery layers to the domain layer.

cc @da2ce7

josecelano:
Refactor packages:  review `AnnounceEvent` struct.

- Modify the `bittorrent_http_tracker_protocol::v1::requests::announce::Event` introducing a new variant `Empty`. This aligns our implementation with the aquatic one and make conversions between HTTP and UDP Announce events direct. It's also part of the protocol in the [BEP 3](https://www.bittorrent.org/beps/bep_0003.html).
- Keep the UDP Announce Event as the primitive type used everywhere. I wanted to removed it from packages that are not the `udp-tracker-core` package but we are using other primitives types from the aquatic UDP tracker protocol crate. We need to carefully study it and open a new issue if we want to remove that dependency.

### DIFF
--- a/packages/axum-http-tracker-server/src/v1/handlers/announce.rs
+++ b/packages/axum-http-tracker-server/src/v1/handlers/announce.rs
@@ -4,11 +4,10 @@
 //! and resolve the client IP address.
 use std::sync::Arc;
 
-use aquatic_udp_protocol::AnnounceEvent;
 use axum::extract::State;
 use axum::response::{IntoResponse, Response};
 use bittorrent_http_tracker_core::services::announce::HttpAnnounceError;
-use bittorrent_http_tracker_protocol::v1::requests::announce::{Announce, Compact, Event};
+use bittorrent_http_tracker_protocol::v1::requests::announce::{Announce, Compact};
 use bittorrent_http_tracker_protocol::v1::responses::{self};
 use bittorrent_http_tracker_protocol::v1::services::peer_ip_resolver::ClientIpSources;
 use bittorrent_tracker_core::announce_handler::AnnounceHandler;
@@ -155,30 +154,6 @@ fn build_response(announce_request: &Announce, announce_data: AnnounceData) -> R
         let response: responses::Announce<responses::Normal> = announce_data.into();
         let bytes: Vec<u8> = response.data.into();
         (StatusCode::OK, bytes).into_response()
-    }
-}
-
-#[must_use]
-pub fn map_to_aquatic_event(event: &Option<Event>) -> aquatic_udp_protocol::AnnounceEvent {
-    match event {
-        Some(event) => match &event {
-            Event::Started => aquatic_udp_protocol::AnnounceEvent::Started,
-            Event::Stopped => aquatic_udp_protocol::AnnounceEvent::Stopped,
-            Event::Completed => aquatic_udp_protocol::AnnounceEvent::Completed,
-        },
-        None => aquatic_udp_protocol::AnnounceEvent::None,
-    }
-}
-
-#[must_use]
-pub fn map_to_torrust_event(event: &Option<Event>) -> AnnounceEvent {
-    match event {
-        Some(event) => match &event {
-            Event::Started => AnnounceEvent::Started,
-            Event::Stopped => AnnounceEvent::Stopped,
-            Event::Completed => AnnounceEvent::Completed,
-        },
-        None => AnnounceEvent::None,
     }
 }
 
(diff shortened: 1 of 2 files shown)


josecelano:
ACK 81675988cb52b7722c027df070b805fb9c65b316

PR #1335, opened 2025-02-28
Refactor: improve REST API client

### LINKED ISSUE #689: Extract  Tracker API Client packages
Relates to: https://github.com/torrust/torrust-index/issues/806
Relates to: https://github.com/torrust/torrust-tracker/issues/255#issuecomment-1477478212

We are using a Tracker API Client in test code.

See: https://github.com/torrust/torrust-tracker/blob/develop/tests/servers/api/v1/client.rs

That client is duplicated in the Index because the Index calls the Tracker API:

https://github.com/torrust/torrust-index/blob/develop/src/tracker/api.rs

We should extract that login into an independent package so that the Index can use it and we remove that duplicate code.

Before extracting the package we should clean it and make it ready for production. Currently, is not ready because

- The URL schema is hardcoded to `http://`.
- It panics if the HTTP fails. We should return an error and let the user handle it.
- We should decide if we expose `request` errors or we wrap them.

I propose to create this packages:

- folder (crates.io package name)
- `packages/api-types` (`torrust-tracker-api-types`)
- `packages/api-client` (`torrust-tracker-api-client`) <- depends on `torrust-tracker-api-types`
- `packages/api-server` (`torrust-tracker-api-server`) <- depends on `torrust-tracker-api-types`.  Maybe in the future

**IMPORTANT:** the implementation in the Index was improved. We have to apply those improvements to this client.

josecelano:
Refactor: improve REST API client:

- Add a timeout to the requests.
- Return an error in the constructor if it can't build the HTTP client.
- Extract constants.

### DIFF
--- a/packages/axum-rest-tracker-api-server/tests/server/v1/contract/authentication.rs
+++ b/packages/axum-rest-tracker-api-server/tests/server/v1/contract/authentication.rs
@@ -16,6 +16,7 @@ async fn should_authenticate_requests_by_using_a_token_query_param() {
     let token = env.get_connection_info().api_token.unwrap();
 
     let response = Client::new(env.get_connection_info())
+        .unwrap()
         .get_request_with_query("stats", Query::params([QueryParam::new("token", &token)].to_vec()), None)
         .await;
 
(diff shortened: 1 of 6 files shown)


josecelano:
ACK 89b0bfd4549d4f7eb453389ec47880ad9956d8d8

PR #1340, opened 2025-03-03
Remove integration tests scaffolding for main app

josecelano:
Remove integration tests scaffolding for main app


josecelano:
I've open to PR only to proof that removing the integration tests causes the docker build to fail.

PR #1356, opened 2025-03-04
Fix bug: number of downloads for a torrent is reset 

### LINKED ISSUE #1264: It seems persisted data about torrents is not loaded from database
It seems persisted data about torrents (number of downloads) is not loaded from database when the tracker stars. The `start` function loads the authentication keys and the whitelist:

```rust
// Load peer keys
if config.core.private {
    app_container
        .keys_handler
        .load_peer_keys_from_database()
        .await
        .expect("Could not retrieve keys from database.");
}

// Load whitelisted torrents
if config.core.listed {
    app_container
        .whitelist_manager
        .load_whitelist_from_database()
        .await
        .expect("Could not load whitelist from database.");
}
```

However I don't see where we load torrent data. The method `TorrentsManager::load_torrents_from_database` is only used in test code.

Maybe that was the expected behavior. However that means the number of downloaded torrents in the database is reset every time the tracker is restarted because the `persist_stats` function only overwrite the in-memory counter. That in-memory counter is not preloaded with the database value when the tracker starts.

```
/// Persists torrent statistics to the database if persistence is enabled.
fn persist_stats(&self, info_hash: &InfoHash, swarm_metadata: &SwarmMetadata) {
    if self.config.tracker_policy.persistent_torrent_completed_stat {
        let completed = swarm_metadata.downloaded;
        let info_hash = *info_hash;

        drop(self.db_torrent_repository.save(&info_hash, completed));
    }
}
```

cc @da2ce7

josecelano:
### Problem

**The number of downloads for a torrent is not persisted correctly**. 

Even when persistence is enabled, it does not work. There are two cases when the counter is reset:

- When you restart the tracker.
- When a torrent entry is removed from the torrents repository because the last peer was removed (peerless torrents). This only happens if "removing peerless torrents" is enabled in the configuration.

Check the [issue](https://github.com/torrust/torrust-tracker/issues/1264) to know how to reproduce the bug.

### Solution

Every time a new torrent entry is added in the torrent repository we need to load stats (also called persistent torrent, which only includes number of downloads currently) from the database. If persistence for stats is enabled.

### Other considerations

**1. Unused code**

There is a method to load the counters for all torrents but it's not used in production code.

```rust
fn import_persistent(&self, persistent_torrents: &PersistentTorrents) {
    for (info_hash, completed) in persistent_torrents {
        if self.torrents.contains_key(info_hash) {
            continue;
        }

        let entry = EntryMutexStd::new(
            EntrySingle {
                swarm: PeerList::default(),
                downloaded: *completed,
            }
            .into(),
        );

        self.torrents.insert(*info_hash, entry);
    }
}
```

I guess it was added to load the information from the database when the tracker is restarted. However that's not enough because torrents are removed from the repository when they do not have peers (peerless torrents). That would work only if removing peerless torrents is disabled.

**2. Better solution**

I think there is a better solution than the one this PR applies. This PR is a temporary patch to fix the bug asap.

I would differentiate to different counters:

1. number of downloads per torrent while it was active in the swarm (since the tracker was restarted)
2. historical number of downloads per torrent per web service (since the service was put into production)

The counter 1 is what we have before merging this PR. The counter is reset every time the torrent is removed from the swarm. We can keep like it's now before merging this PR. This counter is handled at the in-memory torrent repository.

The new counter (counter 2) would be handled by the announce handler. After processing the announce request if the number of downloads has increased the handler also increases the historical counter. This new value should be the one exposed in the API.

```json
{
  "torrents": 379508,
  "seeders": 150440,
  "completed": 58992,  <- this should be the historical
  "leechers": 264865,
  "tcp4_connections_handled": 15,
  "tcp4_announces_handled": 15,
  "tcp4_scrapes_handled": 0,
  "tcp6_connections_handled": 0,
  "tcp6_announces_handled": 0,
  "tcp6_scrapes_handled": 0,
  "udp_requests_aborted": 556458,
  "udp_requests_banned": 24773536,
  "udp_banned_ips_total": 29868,
  "udp_avg_connect_processing_time_ns": 478212,
  "udp_avg_announce_processing_time_ns": 8724637,
  "udp_avg_scrape_processing_time_ns": 296950,
  "udp4_requests": 298728340,
  "udp4_connections_handled": 104369821,
  "udp4_announces_handled": 158354656,
  "udp4_scrapes_handled": 3664492,
  "udp4_responses": 273954123,
  "udp4_errors_handled": 7540716,
  "udp6_requests": 0,
  "udp6_connections_handled": 0,
  "udp6_announces_handled": 0,
  "udp6_scrapes_handled": 0,
  "udp6_responses": 0,
  "udp6_errors_handled": 0
}
```

This patches solves the problem but pre-loading the historical data every time the torrent is added to the repository.

### DIFF
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -26,7 +26,7 @@ impl<S> Environment<S> {
             .container
             .tracker_core_container
             .in_memory_torrent_repository
-            .upsert_peer(info_hash, peer);
+            .upsert_peer(info_hash, peer, None);
     }
 }
 
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -22,7 +22,7 @@ pub struct Environment<S> {
 impl<S> Environment<S> {
     /// Add a torrent to the tracker
     pub fn add_torrent_peer(&self, info_hash: &InfoHash, peer: &peer::Peer) {
-        let () = self
+        let _stats_updated = self
             .container
             .tracker_core_container
             .in_memory_torrent_repository
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -22,7 +22,7 @@ pub struct Environment<S> {
 impl<S> Environment<S> {
     /// Add a torrent to the tracker
     pub fn add_torrent_peer(&self, info_hash: &InfoHash, peer: &peer::Peer) {
-        let _stats_updated = self
+        let _number_of_downloads_increased = self
             .container
             .tracker_core_container
             .in_memory_torrent_repository
--- a/packages/axum-rest-tracker-api-server/src/environment.rs
+++ b/packages/axum-rest-tracker-api-server/src/environment.rs
@@ -37,7 +37,7 @@ where
             .container
             .tracker_core_container
             .in_memory_torrent_repository
-            .upsert_peer(info_hash, peer);
+            .upsert_peer(info_hash, peer, None);
     }
 }
 
--- a/packages/axum-rest-tracker-api-server/src/environment.rs
+++ b/packages/axum-rest-tracker-api-server/src/environment.rs
@@ -33,7 +33,7 @@ where
 {
     /// Add a torrent to the tracker
     pub fn add_torrent_peer(&self, info_hash: &InfoHash, peer: &peer::Peer) {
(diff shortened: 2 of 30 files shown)


josecelano:
ACK 06ef1d5fc6e3f0f06869d94e2b5e963bb6fa5fca

josecelano:
I've discarded the solution proposed in the PR description.

The new solution is described [here](https://github.com/torrust/torrust-tracker/issues/1264#issuecomment-2700929856). I will open a new issue for it.

PR #1363, opened 2025-03-07
Refactor:  remove counter for HTTP connections internally

### LINKED ISSUE #1342: Overhaul stats: Refactor HTTP metrics
There are a couple of comments like this:

```rust
/// > **NOTICE**: as the HTTP tracker does not requires a connection request
/// > like the UDP tracker, the number of TCP connections is incremented for
/// > each `announce` request.
```

The HTTP tracker core keeps a counter for the number of connections handled. However that metrics does not make sense. Internally the counter is the number of announce requests plus the number of scrape requests.

```rust
pub struct Metrics {
    /// Total number of TCP (HTTP tracker) connections from IPv4 peers.
    /// Since the HTTP tracker spec does not require a handshake, this metric
    /// increases for every HTTP request.
    pub tcp4_connections_handled: u64,

    /// Total number of TCP (HTTP tracker) `announce` requests from IPv4 peers.
    pub tcp4_announces_handled: u64,

    /// Total number of TCP (HTTP tracker) `scrape` requests from IPv4 peers.
    pub tcp4_scrapes_handled: u64,

    /// Total number of TCP (HTTP tracker) connections from IPv6 peers.
    pub tcp6_connections_handled: u64,

    /// Total number of TCP (HTTP tracker) `announce` requests from IPv6 peers.
    pub tcp6_announces_handled: u64,

    /// Total number of TCP (HTTP tracker) `scrape` requests from IPv6 peers.
    pub tcp6_scrapes_handled: u64,
}
```

I think it would make sense to delete those counters and add new counter to the `axum-http-tracker-server` package:

- `tcp4_requests`
- `tcp6_requests`

The `udp-tracker-server` has those metrics and also:

- `tcp4_responses`
- `tcp6_responses`

josecelano:
The number of HTTP tracker connections don't make sense. There are connection requests only in the UDP tracker.

That code is removed but, in order to keep backward compatibility, the API still exposes that value which is the:

number of announce requests + number of scrape requests


josecelano:
ACK 89607cc8025fe048a3b9ee41c8b8082b1a8d8321

PR #1367, opened 2025-03-10
New API feature: allow API clients to authenticate  via authentication header

### LINKED ISSUE #727: Remove API token from logs
We are using a `token` query param for API authentication and we are logging the whole request URL.

```output
2024-03-11T16:53:33.249051604+00:00 [API][INFO] request; method=GET uri=/api/v1/torrents?token=MyAccessToken&info_hash=2b66980093bc11806fab50cb3cb41835b95a0362 request_id=d99df52a-dfb8-4608-9974-b4d9c445ee41
2024-03-11T16:53:33.249113794+00:00 [API][INFO] response; latency=0 status=200 OK request_id=d99df52a-dfb8-4608-9974-b4d9c445ee41
```

That means tokens are included in the logs.

We should hide those tokens with **** or change the way we pass the token. We could use an HTTP header like in the Index. I prefer the second option because other proxies could also log the URLs.

josecelano:
The API allows client authentication via a `token` parameter in the URL query:

```console
curl http://0.0.0.0:1212/api/v1/stats?token=MyAccessToken | jq
```

Now it's also possible to do it via an `Authentication Header`:

```console
curl -H "Authorization: Bearer MyAccessToken" http://0.0.0.0:1212/api/v1/stats | jq
```

This is to avoid leaking the token in logs, etc.

For now, it's only optional and recommendable. It could be mandatory in future major API versions.

The API client uses by default the `Authentication Header`. It could be a breaking change if you use the newer client witn an old API that does not support it. However we have not released any crate for the API client yet. And [we are still using a different client in the Index](https://github.com/torrust/torrust-index/issues/806).

### DIFF
--- a/packages/axum-rest-tracker-api-server/src/v1/middlewares/auth.rs
+++ b/packages/axum-rest-tracker-api-server/src/v1/middlewares/auth.rs
@@ -1,7 +1,20 @@
 //! Authentication middleware for the API.
 //!
-//! It uses a "token" GET param to authenticate the user. URLs must be of the
-//! form:
+//! It uses a "token" to authenticate the user. The token must be one of the
+//! `access_tokens` in the tracker [HTTP API configuration](torrust_tracker_configuration::HttpApi).
+//!
+//! There are two ways to provide the token:
+//!
+//! 1. As a `Bearer` token in the `Authorization` header.
+//! 2. As a `token` GET param in the URL.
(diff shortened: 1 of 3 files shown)


josecelano:
ACK 34f2f437db7cfbce2fed5a94f906c39a7794b0f9

PR #1371, opened 2025-03-11
Benchmark flow (reworked)

josecelano:
This is a rework on top of https://github.com/torrust/torrust-tracker/pull/1370

It does not include moving the benchmark output report to the root dir until we implement the workflow to compare new results with the base branch results in a PR.


josecelano:
ACK c15573200caad686cbca63522ff1a8752760d645

PR #1391, opened 2025-03-18
Overhaul stats: Use broadcast channel for events in HTTP Core

### LINKED ISSUE #1388: Overhaul stats: Use broadcast channel for events in HTTP Core
Switch from [mpsc](https://docs.rs/tokio/1/tokio/sync/mpsc/index.html) to [broadcast](https://docs.rs/tokio/1/tokio/sync/broadcast/index.html) channel. That will allow multiple event listeners.

- [mpsc](https://docs.rs/tokio/1/tokio/sync/mpsc/index.html): multi-producer, single-consumer channel. Many values can be sent.
- [broadcast](https://docs.rs/tokio/1/tokio/sync/broadcast/index.html): multi-producer, multi-consumer. Many values can be sent. Each receiver sees every value.

It will make the code a little bit more complex becuase we have to handle "[Lagging](https://docs.rs/tokio/1.44.1/tokio/sync/broadcast/index.html#lagging)" but evens with one one subscriber are not very useful. They are limited for collecting global metrics or do other stuff but in a centralized way. Multiple listeners allows decoupling logic associated to events.

cc @da2ce7

josecelano:
Stats events were introduced to collect tracker metrics. We only have global metrics (aggregate metrics for all UDP and HTTP trackers).

This will change in the future. We will have:

- **Segregated metrics**: one listener per tracker (per socket).
- **Generic events**: there could be other events consumers. Events will be decoupled from stats.

This change allows multiple receivers in the channel.

For now, we one use one listener but with this change will be easy to add more.

### DIFF
--- a/packages/http-tracker-core/benches/helpers/util.rs
+++ b/packages/http-tracker-core/benches/helpers/util.rs
@@ -108,11 +108,11 @@ pub fn sample_info_hash() -> InfoHash {
 use bittorrent_http_tracker_core::statistics;
 use futures::future::BoxFuture;
 use mockall::mock;
-use tokio::sync::mpsc::error::SendError;
+use tokio::sync::broadcast::error::SendError;
 
 mock! {
     HttpStatsEventSender {}
     impl statistics::event::sender::Sender for HttpStatsEventSender {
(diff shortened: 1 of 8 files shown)


josecelano:
ACK c57e15144769b6524588f46110568c78749a2e7e

josecelano:
ACK 5f9c4d3f2a04289055e6ccfc2f530cb82d3f47ea

PR #1402, opened 2025-03-21
Overhaul stats: add a configuration option to enable/disable stats per server

### LINKED ISSUE #1401: Overhaul stats: add a configuration option to enable/disable stats per server
This feature is not implemented yet. In fact this is the first step. I want to add a new config option to enable/disable stats for each service (associated to a socket address).

```toml
[core]

[[udp_trackers]]
bind_address = "0.0.0.0:6969"
tracker_usage_statistics = true  <- the new option

  [udp_trackers.cookie_lifetime]
  secs = 120
  nanos = 0

[[http_trackers]]
bind_address = "0.0.0.0:7070"

[http_api]
bind_address = "0.0.0.0:1212"

  [http_api.access_tokens]
  admin = "***"

[health_check_api]
bind_address = "127.0.0.1:1313"
```

- I will use the same name we are using for the global metrics: `tracker_usage_statistics`.
- The default value will be `false`.
- For the time being, events are only used for stats even if they are already decoupled from stats. But for this issue and the parent issue we will continue enabling events when stats are enabled. We will introduce a new config option in the future to enable events even if you want to disable stats (why? because you might want to build your own stats).

This new config option will have no effect until the feature is implemented.

josecelano:
Add a configuration option to enable/disable stats per server. It does not have any effect yet.

### DIFF
--- a/packages/configuration/src/v2_0_0/core.rs
+++ b/packages/configuration/src/v2_0_0/core.rs
@@ -103,6 +103,7 @@ impl Core {
     fn default_tracker_policy() -> TrackerPolicy {
         TrackerPolicy::default()
     }
+
     fn default_tracker_usage_statistics() -> bool {
         true
     }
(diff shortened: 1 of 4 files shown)


josecelano:
ACK 57d884d8dd6128a019ad5253eb094507689cb83b

Issue #1263, opened 2025-02-12
Overhaul stats: Segregated metrics for each tracker running on a different socket address


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1341#issuecomment-2713624185
Depends on: https://github.com/torrust/torrust-tracker/issues/1385

**UPDATED ON 2025-03-20:** added sub-tasks

**IMPORTANT!!**: global metrics are not going to be removed (at least until the next major version)

While I was [reviewing the `tracker-core` package documentation](https://github.com/torrust/torrust-tracker/issues/1261) I've realized we have the `tracker_usage_statistics` config option in the core configuration:

```toml
[core]
tracker_usage_statistics = true
```

That value enables or disables the statistics.

[I've recently moved the statistics out of the tracker core package](https://github.com/torrust/torrust-tracker/issues/1228). Therefore It does not make sense to have that config option there because there is nothing related to statistics tin the `tracker-core` package now.

That value is used in the app bootstrap:

```
// HTTP stats
let (http_stats_event_sender, http_stats_repository) =
    http_tracker_core::statistics::setup::factory(configuration.core.tracker_usage_statistics);
let http_stats_event_sender = Arc::new(http_stats_event_sender);
let http_stats_repository = Arc::new(http_stats_repository);

// UDP stats
let (udp_stats_event_sender, udp_stats_repository) =
    udp_tracker_core::statistics::setup::factory(configuration.core.tracker_usage_statistics);
let udp_stats_event_sender = Arc::new(udp_stats_event_sender);
let udp_stats_repository = Arc::new(udp_stats_repository);
```

The same config option enables stats in both the UDP and HTTP trackers.

The current config omitting irrelevant values:

```toml
[core]
tracker_usage_statistics = true

[[udp_trackers]]
bind_address = "0.0.0.0:6969"

  [udp_trackers.cookie_lifetime]
  secs = 120
  nanos = 0

[[http_trackers]]
bind_address = "0.0.0.0:7070"

[http_api]
bind_address = "0.0.0.0:1212"

  [http_api.access_tokens]
  admin = "***"

[health_check_api]
bind_address = "127.0.0.1:1313"
```

Maybe we should create two new values for UDP and HTTP trackers. However in the current configuration there is not a place to put generic configuration for all HTTP trackers or all UDP trackers.

We can only add a config option per instance.

**NOTE:** That you can run more than one HTTP and/or UDP trackers.

If we do that the config could be:

```toml
[...]

PR #1404, opened 2025-03-21
Overhaul stats: start collecting stats per server instance (per socket)

josecelano:
Start collecting stats per server instance (per socket).

### Sub-tasks

- [x] Add E2E tests for global metrics.
- [ ] HTTP Tracker Instances.
  - [x] ~~Instantiate independent services (event sender and stat repo) for each instance~~. **DISCARDED**.
- [ ] UDP Tracker Instances.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -38,7 +38,7 @@ use crate::container::AppContainer;
 /// - Can't retrieve tracker keys from database.
 /// - Can't load whitelist from database.
 #[instrument(skip(config, app_container))]
-pub async fn start(config: &Configuration, app_container: &Arc<AppContainer>) -> Vec<JoinHandle<()>> {
+pub async fn start(config: &Configuration, mut app_container: AppContainer) -> Vec<JoinHandle<()>> {
     if config.http_api.is_none()
         && (config.udp_trackers.is_none() || config.udp_trackers.as_ref().map_or(true, std::vec::Vec::is_empty))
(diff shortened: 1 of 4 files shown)


josecelano:
[This commit](a9b1c14b8436d4eff2819ccc1a191f818d73c699) creates services per HTTP tracker instances (server bound to a socket address) to collect metrics only for that instance.

There should be tests failing because each HTTP tracker instance can use only one event sender and metrics repository. Therefore global metrics should not be updated after this commit.

I think the problem is there are no E2E tests for statistics. There are only integration tests at the HTTP Core package level. That means metrics per instance work.

I will add an E2E tests for global metrics even if we are planning to remove them in the future. In the mean time, it will be covered.

cc @da2ce7

josecelano:
There is another problem. You can not halt the application with CTRL+C. I think it might be a deadlock. Only the servers before the HTTP servers are started. I'm going to try a different approach. In the `AppContainer` initialization I initialize all the services needed: the global ones and the duplicates ones that we need for each tracker instance. This way the `start` function would be pretty much the same.

josecelano:
I have fixed the problem stopping the app and [added the E2E tests](https://github.com/torrust/torrust-tracker/issues/1407). I've also discarded using one event sender and stats repo per server instance. [I will change the singleton repository to collect stats for all server instances in a segregated way](https://github.com/torrust/torrust-tracker/issues/1263#issuecomment-2748955164).

Issue #1327, opened 2025-02-26
Overhaul stats


josecelano:
Depends on: https://github.com/torrust/torrust-tracker/issues/1319
Relates to: https://github.com/torrust/torrust-tracker/issues/1368

There are some inconsistencies in how we handle stats events and also some things that can be improved.

I'm also considering keeping metrics for each socket in addition to the aggregate metrics. We are currently collecting single aggregated metrics for for all servers (grouped by UDP or HTTP trackers).

**UPDATE 2025-03-17**

This refactor is needed to enrich events. The final goal is to expose those events to allow third-party consumers build their metrics. For example: you can create your event listeners to collects metrics for your private Index from Torrust Tracker.

cc @da2ce7 

### Notes

- The issue https://github.com/torrust/torrust-tracker/issues/1368 migth imply opening new sub-issues for this epic.

josecelano:
Finished.

josecelano:
Reopening becuase I have found a couple of new issues.

josecelano:
All subtasks done.

da2ce7:
Well done, this was a massive overhaul! Making the Torrust Tracker monitorable like this is a huge achievement 👍 Thank You for your hard work and focus on this task epic! 😄

PR #1410, opened 2025-03-24
Add a test for global metrics with 2 HTTP trackers

### LINKED ISSUE #1407: Overhaul stats: add E2E tests for the global metrics
Before implementing the [segregate metrics](https://github.com/torrust/torrust-tracker/issues/1403) I want to add some integration tests for the global metrics.

While I was working on https://github.com/torrust/torrust-tracker/pull/1404 I realized we don't have integration tests for that. There are tests for a single server. For example:

- HTTP tracker: https://github.com/torrust/torrust-tracker/blob/develop/packages/axum-http-tracker-server/tests/server/v1/contract.rs#L670-L752

The tests use the test environment to create a single instance of the HTTP or UDP trackers. 

We need to add a new tests at the application level at least with one test for each tracker type:

- Run the whole tracker with the HTTP tracker enabled.
- Make an `announce` request.
- Get the global metrics from the REST API.
- Check that the counter for announce request was increased.

Same for UDP tracker.

I don't know yet if we can use an E2E test:

https://github.com/torrust/torrust-tracker/blob/develop/src/console/ci/e2e/runner.rs

Or maybe we can create a new integration tests running the whole app.

I think the first option is faster to implement but slower to execute (it's using docker). I would prefer the second option. If it takes too long I would do it with E2E tests for now. Global metrics is something we are going to remove in future versions and the main app bootstrapping (containers and jobs) is not changed very often.

cc @da2ce7

josecelano:
A new integration test that checks that the global metrics are updated when you run 2 HTTP trackers. Only one metric is checked in this test.

It uses fixed ports that might conflict with other running instances in the future. We should use a random free port if we run more integration tests like this in the future.

The rest of the functionality is already tested at the package level with only one tracker.

I've added to make sure nothing is broken while working on the [stats overhaul](https://github.com/torrust/torrust-tracker/issues/1327).


josecelano:
ACK 4e59dd7879b96a8f07c49725b2ab930d241d834b

josecelano:
ACK b53da0736078719364fa98e07071ce8adc90b63c

PR #1412, opened 2025-03-25
Refactor: `AppContainer` to initialize all services in the setup phase

### LINKED ISSUE #1411: Refactor: `AppContainer` to initialize all services in the setup phase
The application runs in two phases:

1. Setup
2. Start

```rust
pub async fn run() -> (Arc<AppContainer>, Vec<JoinHandle<()>>, Registar) {
    let (config, app_container) = bootstrap::app::setup();

    let app_container = Arc::new(app_container);

    let (jobs, registar) = start(&config, &app_container).await;

    (app_container, jobs, registar)
}
```

This is a common approach in many frameworks and apps. First, we build the needed services and then we run them.

All services dependencies are resolves in the setup phase.

While I was working on [this refactor](https://github.com/torrust/torrust-tracker/pull/1406) I realized that we build the tracker and API containers in the second phase calling these methods:

- `AppContainer::http_tracker_container`
- `AppContainer::udp_tracker_container`
- `AppContainer::tracker_http_api_container`

I think we should move the instantiation to the `setup` phase.

Although most of the services are already build there is a container per server instance and some services are only specific to that instance. For example, the server configuration is cloned from the configuration and there is no copy of that service in the container at the main app level.

It would be good to create those container in the setup phase:

- To set a clear division between services instantiation (setup) and running the servers.
- To be able in the future to access services that are specific to instances.

This is also related to [this issue](https://github.com/torrust/torrust-tracker/issues/1263). I was considering using independent services for each server instance (each HTTP or UDP tracker running on a different socket address). Even thought I have discarded that approach (for the time being), in the future we might want to instantiate services per instance. Having a clear distinction between instantiation and execution for services now will allow to implement such changes easier in the future.

cc @da2ce7

josecelano:
Refactor `AppContainer` to initialize all services in the setup phase.

I've also removed duplicate code for services initialization.

The REST API container is still initialized in the "start" phase because requires more changes to handle the optional value. We don't need it at the root level for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4535,6 +4535,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "thiserror 2.0.12",
  "tokio",
  "torrust-axum-health-check-api-server",
  "torrust-axum-http-tracker-server",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ regex = "1"
 reqwest = { version = "0", features = ["json"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
+thiserror = "2.0.12"
 tokio = { version = "1", features = ["macros", "net", "rt-multi-thread", "signal", "sync"] }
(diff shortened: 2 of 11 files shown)


josecelano:
ACK 299d0f3ea7566a73aa1edeedef839d92d9545896

PR #1423, opened 2025-03-28
Refactor: extract type `ServiceBinding`

### LINKED ISSUE #1422: Refactor: extract type `ServiceBinding`
In this PR:

- https://github.com/torrust/torrust-tracker/pull/1416

I added a new field to the health check API endpoint `listen_url`:

```json
{
  "status": "Ok",
  "message": "",
  "details": [
    {
      "listen_url": "http://0.0.0.0:7070/",
      "binding": "0.0.0.0:7070",
      "service_type": "http_tracker",
      "info": "checking http tracker health check at: http://0.0.0.0:7070/health_check",
      "result": {
        "Ok": "200 OK"
      }
    }
  ]
}
```

I used a `Url` type but I have decided to introduce a new type, because it's a URL with some restrictions and actually it also allows some invalid URLs.

The idea of the ype is to represent the same thing as the "binding" socket address but including information about the URL: protocol and path.

I will use this new type in the `ServiceHealthCheckJob`:

```rust
pub struct ServiceHealthCheckJob {
    pub listen_url: Url,
    pub binding: SocketAddr,
    pub info: String,
    pub service_type: String,
    pub job: JoinHandle<ServiceHeathCheckResult>,
}
```

I will also rename the field in the health check API endpoint.

I will also use this new type in other places where we need the protocol in addition to the bind address. Like for example in this issue:

https://github.com/torrust/torrust-tracker/issues/1415

cc @da2ce7

josecelano:
Refactor: extract type `ServiceBinding`.

It's the protocol + socket binding address for a service.

- Protocols are only: udp, http and https (protocols used by the tracker
  now).
- The port must be greater than zero (already assigned).
- Any protocol and port combination is possible. For example: http
  protocol with a port different than 80.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4375,6 +4375,7 @@ dependencies = [
  "torrust-server-lib",
  "torrust-tracker-clock",
  "torrust-tracker-configuration",
+ "torrust-tracker-primitives",
  "torrust-tracker-test-helpers",
  "torrust-udp-tracker-server",
  "tower-http",
@@ -4510,10 +4511,11 @@ name = "torrust-server-lib"
 version = "3.0.0-develop"
 dependencies = [
  "derive_more",
+ "rstest",
  "tokio",
+ "torrust-tracker-primitives",
  "tower-http",
  "tracing",
- "url",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


josecelano:
ACK 007110356580f8bd976a1ae3ce0f4e73046a96ea

josecelano:
ACK 9d332f5bf5f28a3d71dd783010b74a1bb7c79fbf

josecelano:
It's almost done. I only have to fix two tests because the service URL and the service check URL are not the same.

josecelano:
ACK 29fa324417201eb6a8ec0103e7fb200ca46670d2

PR #1414, opened 2025-03-25
New API endpoint with extendable metrics

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1263#issuecomment-2748955164

It adds a new  REST API endpoint with a new extendable format for metrics.

In this PR I will only add the metrics for:

- Performance
- Announce requests
- Scrape requests

See the complete list [here](https://github.com/torrust/torrust-tracker/issues/1263#issuecomment-2766592064).

It requires emitting the new metrics from the:

- HTTP Tracker Core package
- UDP Tracker Core package
- UDP Tracker Server package

The remaining metrics (`torrents` and `peers`) require adding metrics to the tracker core package.

### JSON format

**URL:** http://0.0.0.0:1212/api/v1/metrics?token=MyAccessToken

**Sample response:**

```json
{
  "metrics":[
    {
      "kind":"counter",
      "name":"http_tracker_core_announce_requests_received_total",
      "samples":[
        {
          "value":1,
          "recorded_at":"2025-04-02T00:00:00+00:00",
          "labels":[
            {
              "name":"server_binding_ip",
              "value":"0.0.0.0"
            },
            {
              "name":"server_binding_port",
              "value":"7070"
            },
            {
              "name":"server_binding_protocol",
              "value":"http"
            }
          ]
        }
      ]
    },
    {
      "kind":"gauge",
      "name":"udp_tracker_server_performance_avg_announce_processing_time_ns",
      "samples":[
        {
          "value":1.0,
          "recorded_at":"2025-04-02T00:00:00+00:00",
          "labels":[
            {
              "name":"server_binding_ip",
              "value":"0.0.0.0"
            },
            {
              "name":"server_binding_port",
              "value":"7070"
            },
            {
              "name":"server_binding_protocol",
              "value":"http"
            }
          ]
        }
      ]
    }
  ]
}
```

### Prometheus format

**URL:** http://0.0.0.0:1212/api/v1/metrics?token=MyAccessToken&format=prometheus

```
http_tracker_core_announce_requests_received_total{server_binding_ip="0.0.0.0",server_binding_port="7070",server_binding_protocol="http"} 1
udp_tracker_server_performance_avg_announce_processing_time_ns{server_binding_ip="0.0.0.0",server_binding_port="7070",server_binding_protocol="http"} 1
```

### Manual Tests

You can increase the number of `announce` or `scrape` requests by using the console tracker client.

HTTP:

```console
[...]

PR #1426, opened 2025-03-31
POC: New API endpoint with extendable metrics

josecelano:
Relates to: 
  - https://github.com/torrust/torrust-tracker/issues/1263#issuecomment-2748955164
  - https://github.com/torrust/torrust-index/issues/809

This is only a Proof of Concept.

It adds a new  REST API endpoint with a new extendable format for metrics:

**URL:** http://0.0.0.0:1212/api/v1/metrics?token=MyAccessToken

**Sample response:**

```json
{
  "labeled_metrics": [
    {
      "metric": {
        "name": "announce_requests_received_total",
        "kind": "counter",
        "value": 1
      },
      "labels": {
        "ip_version": "ipv4",
        "protocol": "http",
        "url": "http://0.0.0.0:7070"
      }
    },
    {
      "metric": {
        "name": "scrape_requests_received_total",
        "kind": "counter",
        "value": 1
      },
      "labels": {
        "ip_version": "ipv4",
        "protocol": "http",
        "url": "http://0.0.0.0:7070"
      }
    }
  ]
}
```

It only supports JSON format for now. It doesn't support Prometheus format like the current stats endpoint:

**URL:** http://0.0.0.0:1212/api/v1/stats?token=MyAccessToken&format=prometheus

The Prometheus format response should be like the following:

```
announce_requests_received_total{ ip_version="ipv4", protocol="http", url="http://0.0.0.0:7070"} 1

scrape_requests_received_total{ ip_version="ipv4", protocol="http", url="http://0.0.0.0:7070"} 1
```

### TODO

- [ ] Replace primitive types in the new labeled metrics with new tpyes to enforce constraints. For example, metrics names, labels, metric types, etc.
- [ ] Inject the rigth URL scheme. It's hardcoded now.
- [ ] Implement for UDP metrics. It's only implemented for HTTP tracker metrics. This would require merging labeled metrics.
- [ ] Use a f64 for metric values instead of u64.
- [ ] The metric name and label set pair must be unique in the array of labeled metrics. Enforce constraint.
- [ ] Implement versioning. A given API version must contain a set of labeled metrics. Clients expects some labeled metrics to be included in a API version. We need to initialize the array of metrics with all the expected labeled metrics with the initial value.
- [ ] Review names for fields in the API response. For example rename `url` to `server_url`, ... 
- [ ] Review what labels we include per metric. For example the "protocol" label can be derived from the URL but it requires parsing the URL and it makes harder to build graphs in Grafana. I guess we should include labels for the type of aggregate date we want to get.

### IMPORTANT

- Each metric+label set values combination is a different time series in Prometheus. We can't use labels with high cardinality. See https://prometheus.io/docs/practices/naming/#labels. For example, we can't use client's IPs.

@da2ce7 it would be nice to have feedback before continuing working on this approach.


josecelano:
I have continued with the implementation of the POC [here](https://github.com/torrust/torrust-tracker/pull/1414).

PR #1431, opened 2025-04-04
Change rustdoc docs to only compile+run in release mode

victorb:
As-is, it asks the user to do a release mode build first, then after creating the directories, then do a debug mode build + run.

This commit simplifies it and avoids compiling the tracker in both debug and release mode, and instead only compiles+run with release mode


josecelano:
ACK a46011637ab11a0aef09ac70985069efe50364f9

Issue #1449, opened 2025-04-14
Overhaul stats: Handle lagging in event listeners channels


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1385

When I run the tracker in the demo server, stats stop working after a while, and I get this error in logs:

```
tracker  | 2025-04-14T08:43:30.045862Z ERROR torrust_udp_tracker_server::statistics::event::listener: Error receiving udp tracker server event: Lagged(9)
```

The problem was produced when I [switched to Tokio broadcast channels to allow multiple event listeners](https://github.com/torrust/torrust-tracker/issues/1385).

Lagging is not handled in the event listener. For example, the UDP tracker server event listener:

```rust
use tokio::sync::broadcast;
use torrust_tracker_clock::clock::Time;

use super::handler::handle_event;
use crate::event::Event;
use crate::statistics::repository::Repository;
use crate::CurrentClock;

pub async fn dispatch_events(mut receiver: broadcast::Receiver<Event>, stats_repository: Repository) {
    loop {
        match receiver.recv().await {
            Ok(event) => handle_event(event, &stats_repository, CurrentClock::now()).await,
            Err(e) => {
                tracing::error!("Error receiving udp tracker server event: {:?}", e);
                break;
            }
        }
    }
}
```

The channel capacity is 1024, so when unhandled events reach the channel capacity, we receive an error in the loop that breaks the loop, stopping the listener.

Since:

- We only have one listener now to update metrics.
- I want metrics to be exact.

As a temporary solution, I will increase the channel capacity to have the same effect as when we used a mspc channel.

I will also continue the loop when the listener receives that error.

IMPORTANT: The metrics will **NOT** be exact if there are lagged events.

cc @da2ce7 

### Alternative solution 1

Restore the mspc channel for metrics.

Pros:

- Precise metrics

Cons:

- Slower request processing? The request has to wait until the metrics are updated.
- We will need two channels: one mpsc channel for metrics and a broadcast channel for generic listeners.

**NOTE:** The main reason I wanted to have multiple listeners is that I want third-party software to be able to build their own metrics from tracker events (by registering their event listeners).

josecelano:
### Alternative solution 2

Maybe crossbeam [unbounded](https://docs.rs/crossbeam/latest/crossbeam/channel/fn.unbounded.html) channel?

https://docs.rs/crossbeam/latest/crossbeam/channel/index.html

josecelano:
[...]

PR #1451, opened 2025-04-14
Temporary solution to handle lagged stats listeners

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1449

When I switched the channels for statistics from [tokio mpsc](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html) to [tokio broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html), I introduced a bug that was only revealed after deploying the demo tracker.

The problem is that MPSC provides backpressure, so all events are processed. However, the broadcast channel has a limit, and listeners are informed with a `Lagged` error when the limit has been reached and events have been deleted.

We need to decide if:

1. We go back to mpsc channel
2. Or we implement another solution
3. Or we accept imprecise metrics on a high load

In the meantime, I have increased the channel capacity to be able to collect all events at the current demo peak load. That should avoid the problem of loosing events in practice.

I have also fixed the listener to handle the `Lagged` error. In this case, we don't have to break the loop but continue processing events to update metrics. The only difference is metrics will not be precise after this error happens because some events were lost.

### DIFF
--- a/packages/http-tracker-core/src/event/sender.rs
+++ b/packages/http-tracker-core/src/event/sender.rs
@@ -7,7 +7,7 @@ use tokio::sync::broadcast::error::SendError;
 
 use super::Event;
 
-const CHANNEL_CAPACITY: usize = 1024;
+const CHANNEL_CAPACITY: usize = 32768;
 
 /// A trait for sending sending.
 #[cfg_attr(test, automock)]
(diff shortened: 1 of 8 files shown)


josecelano:
ACK 6fdbc47d23fb4d246284a27721235134ac34038f

PR #1467, opened 2025-04-15
Overhaul stats: Refactor metrics (part 1)

### LINKED ISSUE #1445: Overhaul stats: Refactor metrics (part 1)
There is a [bug](https://github.com/torrust/torrust-tracker/issues/1439) in the Torrust Demo with the new metrics that I can't reproduce locally.

I will introduce some changes to the `metrics` package to help me find the problem.

I will also make other minor changes in that package and other packages.

### Subtasks

`metrics` package:

- Return errors instead of panicking in the `MetricCollection` struct.
- Add tracing crate.
- Return an optional in `get_counter_value` and `get_gauge_value`. They are returning the default value for that metric type if the metric does not exist. That the behavior we want for `increase_counter` and `set_gauge` but not for these methods. They are currently used only in tests.

`http-tracker-code`, `udp-tracker-core`, `udp-tracker-server` packages:

- Add logs to listeners initialization

Change this:

```rust
pub fn run_event_listener(&mut self, receiver: Receiver<Event>) {

    tokio::spawn(async move { dispatch_events(receiver, stats_repository).await });
}
```

To something like this:

```rust
pub fn run_event_listener(&mut self, receiver: Receiver<Event>) {

    // TARGET should be HTTP_TRACKER_CORE
    tracing::info!(TARGET, "Starting event listener");

    tokio::spawn(async move {
        dispatch_events(receiver, stats_repository).await;

        tracing::info!(TARGET, "Event listener finished");
    });
}
```

NOTE: This won't help identify a panic inside the `dispatch_events` function. We should log the error returned by `tokio::spawn`, but the function is not async. We need to move the function call from the app setup to the app start as described [here](https://github.com/torrust/torrust-tracker/issues/1444).

josecelano:
Refactor metrics (part 1)

### Subtasks

- [x] Extract macro for metric name constructor (to fail at compilation time with empty names)
- [x] Replace metric name constructor by new macro
- [x] Extract macro for label name constructor (to fail at compilation time with empty names)
- [x] Replace label name constructor by new macro
- [x] Return an optional in `get_counter_value` and `get_gauge_value`.
- [x] Return errors instead of panicking in the `MetricCollection` struct.
- [x] Add logs to the listener's initialization (when it starts and finishes).
- [x] Add req kind label to the number of errors metric.
- [x] Add `tracing` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4791,6 +4791,7 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.12",
  "torrust-tracker-primitives",
+ "tracing",
 ]
 
 [[package]]
--- a/packages/http-tracker-core/src/event/mod.rs
+++ b/packages/http-tracker-core/src/event/mod.rs
@@ -1,6 +1,7 @@
 use std::net::{IpAddr, SocketAddr};
 
-use torrust_tracker_metrics::label::{LabelName, LabelSet, LabelValue};
+use torrust_tracker_metrics::label::{LabelSet, LabelValue};
+use torrust_tracker_metrics::label_name;
 use torrust_tracker_primitives::service_binding::ServiceBinding;
 
 pub mod sender;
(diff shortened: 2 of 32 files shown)


josecelano:
ACK f9ad729f2899ddd351506adb240757b73bbaa309

Issue #1385, opened 2025-03-18
Overhaul stats: Allow multiple events listeners (subscribers to events)


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1341#issuecomment-2713624185

In order to start collecting **metrics per socket address** (segregate metrics per tracker service running on a different socket) we need to add more than one listener. There will be **one listener per service** that will update the metrics for that service.

Right now, the "factory" stats function creates both the event sender and the listener. We need to be able to:

- Create one sender that can be shared.
- Be able to create many listeners (subscribe listeners to events). One listener per socket.

For the time being we will keep the logic inside the `statistics` module. However, after finishing [the overhaul of the stats](https://github.com/torrust/torrust-tracker/issues/1327) we will decouple events from stats. In the future events might be used for other purposes, not only collecting metrics.

cc @da2ce7

josecelano:
All sub-issues implemented.

Issue #1405, opened 2025-03-21
Overhaul stats: Implement a graceful shutdown for broadcast channels


josecelano:
Depends on: https://github.com/torrust/torrust-tracker/issues/1449
Relates to: https://github.com/torrust/torrust-tracker/issues/1385

In this [issue](https://github.com/torrust/torrust-tracker/issues/1385) I changed the events channels from `mpsc` to `broadcast`. I wanted to have many listeners. It's working fine. However when you send the halt signal to the applcaition (CRTL+c) you get this erros in the logs:

```output
2025-03-21T16:11:39.493220Z ERROR bittorrent_udp_tracker_core::statistics::event::listener: Error receiving udp tracker core event: Closed
2025-03-21T16:11:39.493352Z ERROR bittorrent_http_tracker_core::statistics::event::listener: Error receiving http tracker core event: Closed
2025-03-21T16:11:39.493361Z ERROR torrust_udp_tracker_server::statistics::event::listener: Error receiving udp tracker server event: Closed
```

The problem is we drop the sender and the listener return an error because the channel is closed. We should implement a gracefully shutdown notifying listeners and giving them time to finish processing the current events.

Since I don't see an option in the `broadcast`  channel to notify listener we need to research other alternatives.

cc @da2ce7

josecelano:
First ideas using AI.
 
- The prompt was straightforward, just ask alternatives to gracefully close Tokio broadcast cahnnels.
- Model: [Claude 3.5 Sonnet | Anthropic](https://openrouter.ai/anthropic/claude-3.5-sonnet)

cc @da2ce7 

---

In Tokio, you can gracefully close a broadcast channel by using a combination of techniques. Here are a few approaches:

1. **Using a dedicated shutdown signal:**
```rust
use tokio::sync::broadcast;
use tokio::sync::broadcast::error::RecvError;

struct Sender<T> {
    tx: broadcast::Sender<T>,
    shutdown_tx: broadcast::Sender<()>,
}

struct Receiver<T> {
    rx: broadcast::Receiver<T>,
    shutdown_rx: broadcast::Receiver<()>,
}

impl<T> Sender<T> {
    fn shutdown(&self) {
        // Send shutdown signal before dropping the sender
        let _ = self.shutdown_tx.send(());
    }
}

async fn listener(mut receiver: Receiver<String>) {
    loop {
        tokio::select! {
            // Check for shutdown signal
            Ok(()) = receiver.shutdown_rx.recv() => {
[...]

PR #1476, opened 2025-04-23
Refactor event listeners initialization

### LINKED ISSUE #1444: Overhaul Stats: Refactor event listeners initialization
All the packages using events have a function to initialize statistics:

```rust
pub fn factory(tracker_usage_statistics: bool) -> (Option<Box<dyn event::sender::Sender>>, statistics::repository::Repository) {
    let mut keeper = statistics::keeper::Keeper::new();

    let opt_event_sender: Option<Box<dyn event::sender::Sender>> = if tracker_usage_statistics {
        let broadcaster = Broadcaster::default();

        keeper.run_event_listener(broadcaster.subscribe());

        Some(Box::new(broadcaster))
    } else {
        None
    };

    (opt_event_sender, keeper.repository)
}
```

That example is from the UDP tracker server package.

The problem is this factory method is called when the `AppContainer` is initialized in:

```rust
impl AppContainer {
    #[instrument(skip(configuration))]
    pub fn initialize(configuration: &Configuration) -> AppContainer {
        // ...
    }
}
```

And the `AppContainer` is initialized in the app setup:

```rust
pub async fn run() -> (Arc<AppContainer>, Vec<JoinHandle<()>>, Registar) {
    let (config, app_container) = bootstrap::app::setup(); // <- AppContainer initialized here

    let app_container = Arc::new(app_container);

    let (jobs, registar) = start(&config, &app_container).await;

    (app_container, jobs, registar)
}
```

I think the `app::setup` should only create the services (constructors) but not run any service, including spawning tasks.

I think we should call the `keeper.run_event_listener(broadcaster.subscribe());` function in the `start` function where we start the jobs.

I think that the separation between building and running is a better design.

One concrete example of why I think this would be good is in the `Keeper::run_event_listener`:

```rust
pub struct Keeper {
    pub repository: Repository,
}

impl Default for Keeper {
    fn default() -> Self {
        Self::new()
    }
}

impl Keeper {
    #[must_use]
    pub fn new() -> Self {
        Self {
            repository: Repository::new(),
        }
    }

    pub fn run_event_listener(&mut self, receiver: Receiver<Event>) {

        tokio::spawn(async move { dispatch_events(receiver, stats_repository).await });
    }
}
```

The task for the event listener is spawned during the app setup (building services), and it's not an async function. If the event dispatcher panics (function `dispatch_events`) there is no way to await for the returned value and handle it properly.

I think that is what is happening in this [bug](https://github.com/torrust/torrust-tracker/issues/1439).

These tasks should also be included as jobs in the `start` function so we can wait for them to finish. If they panic (or finish) the main program can do something like send a notification or retry to launch the listener, ...

josecelano:
Move the spawning of new tasks for event listeners from app container instantiation to job execution.

This will enable us to easily add more event listeners in the future. Additionally, it enhances the way the tracker handles spawned tasks. We should have control over all JoinHandles for all tasks.

This promotes a cleaner task lifecycle. Some good practices:

- There should always be an owner, responsible for the spawned tasks. No orphan tasks.
- It should be possible to gracefully shut down the task.
- The state of the task should be known.
- Sometimes the owner needs to retrieve data (for example, status) from the task (when it starts or while it's active) or send the task additional commands.

### Subtasks

- [x] HTTP Tracker Core event listener
  - [x] Step 1: Separate factory from spawning task in the current `setup::factory` function.
  - [x] Step 2: Split factory function into factory and running listener.
  - [x] Step 3: Add `Keeper` to `HttpTrackerCoreContainer` (so we can run the listener later).
  - [x] Step 4: Move task creation to app start.
  - [x] Step 5. Start event listener in test environment.
- [x] UDP Tracker Core event listener (Step 1 to 4)
- [x] UDP Tracker Server event listener (Step 1 to 4)
 
NOTE: This PR will not include the refactor to allow multiple listeners. See https://github.com/torrust/torrust-tracker/issues/1385.

### DIFF
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -75,7 +75,7 @@ impl Environment<Stopped> {
             .server
             .start(self.container.http_tracker_core_container.clone(), self.registar.give_form())
             .await
-            .unwrap();
+            .expect("Failed to start the HTTP tracker server");
 
         Environment {
             container: self.container.clone(),
@@ -105,7 +105,7 @@ impl Environment<Running> {
         }
 
         // Stop the server
-        let server = self.server.stop().await.expect("Failed to stop the http tracker server");
+        let server = self.server.stop().await.expect("Failed to stop the HTTP tracker server");
 
         Environment {
             container: self.container,
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -4,6 +4,7 @@ use bittorrent_http_tracker_core::container::HttpTrackerCoreContainer;
 use bittorrent_primitives::info_hash::InfoHash;
 use bittorrent_tracker_core::container::TrackerCoreContainer;
 use futures::executor::block_on;
+use tokio::task::JoinHandle;
 use torrust_axum_server::tsl::make_rust_tls;
 use torrust_server_lib::registar::Registar;
 use torrust_tracker_configuration::{logging, Configuration};
@@ -17,6 +18,7 @@ pub struct Environment<S> {
     pub container: Arc<EnvContainer>,
     pub registar: Registar,
     pub server: HttpServer<S>,
+    pub event_listener_job: Option<JoinHandle<()>>,
 }
 
 impl<S> Environment<S> {
@@ -54,22 +56,32 @@ impl Environment<Stopped> {
             container,
             registar: Registar::default(),
             server,
+            event_listener_job: None,
         }
     }
 
+    /// Starts the test environment and return a running environment.
+    ///
     /// # Panics
     ///
     /// Will panic if the server fails to start.    
     #[allow(dead_code)]
     pub async fn start(self) -> Environment<Running> {
+        // Start the event listener
+        let event_listener_job = self.container.http_tracker_core_container.stats_keeper.run_event_listener();
+
+        // Start the server
+        let server = self
+            .server
+            .start(self.container.http_tracker_core_container.clone(), self.registar.give_form())
+            .await
+            .unwrap();
+
         Environment {
             container: self.container.clone(),
             registar: self.registar.clone(),
-            server: self
-                .server
-                .start(self.container.http_tracker_core_container.clone(), self.registar.give_form())
-                .await
-                .unwrap(),
+            server,
+            event_listener_job: Some(event_listener_job),
         }
     }
 }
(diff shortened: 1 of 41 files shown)


josecelano:
Hi @da2ce7, I've completed the refactor for one of the three packages that have events. I will apply the same refactor to the other two.

As I was expecting, there is a problem shutting down the tracker when I wait for the event listener to finish because we don't send a halt message to that task. We can include the join handle in the jobs vector once we implement [this issue](https://github.com/torrust/torrust-tracker/issues/1405).

We should also improve the shutdown message.

image

That message is only for axum servers. We should:

- Show the socket address.
- Stop the loop after 90 seconds.
- Stop the loop when there are no connections.

It continues to display the message even if all Axum servers are successfully shut down, but the main tracker cannot stop because there are other pending jobs, such as event listeners, in this case. I will open a new issue to address this improvement. I think we should show the job that is preventing the tracker from shutting down if that task does not provide that feature, like the axum server. We should show something like:

```
!! shuting tracker !!
Waiting for the HTTP core event listener to finish ...
```

Every second or so.

Maybe we should add this at the main level for all jobs regardless of whether the jobs internally show other messages while shutting down (like axum server).

This is a draft, non-tested implementation from ChatGPT 4o:

```rust
use torrust_tracker_lib::app;
use tokio::time::{sleep, Duration};
use tokio::task::JoinHandle;
use std::sync::{Arc};
use std::sync::atomic::{AtomicBool, Ordering};

#[tokio::main]
async fn main() {
    let (_app_container, raw_jobs) = app::run().await;

    // Create flags for each job
    let mut job_flags = Vec::new();
    let mut wrapped_jobs = Vec::new();

    for (i, job) in raw_jobs.into_iter().enumerate() {
        let running = Arc::new(AtomicBool::new(true));
        let running_clone = running.clone();
        job_flags.push((format!("Job {}", i + 1), running.clone()));

        let handle: JoinHandle<()> = tokio::spawn(async move {
            job.await;
            running_clone.store(false, Ordering::SeqCst);
        });

        wrapped_jobs.push(handle);
    }

    // Spawn the status printer
    let status_handle = tokio::spawn({
        let job_flags = job_flags.clone();
        async move {
            loop {
                sleep(Duration::from_secs(5)).await;

                let still_running: Vec<_> = job_flags
                    .iter()
                    .filter(|(_, flag)| flag.load(Ordering::SeqCst))
                    .collect();

                if still_running.is_empty() {
                    break;
                }

                println!("!! shutting tracker !!");
                for (name, _) in still_running {
                    println!("Waiting for {name} to finish ...");
                }
                println!();
            }
        }
    });

    // Wait for CTRL+C
    tokio::signal::ctrl_c().await.unwrap();
    tracing::info!("Torrust tracker shutting down ...");

    // Wait for all job handles
    for handle in wrapped_jobs {
        let _ = handle.await;
    }

    // Wait for the status task to exit
    let _ = status_handle.await;

    tracing::info!("Torrust tracker successfully shutdown.");
}

```

josecelano:
ACK 6d50a784239945299c63116fe5434da92fd0dd6e

josecelano:
ACK 6d50a784239945299c63116fe5434da92fd0dd6e

PR #1487, opened 2025-04-29
Fix shutdown message and improve it

josecelano:
Fix shutdown message and improve it for HTTP servers (they all use Axum).

### Subtasks

- [x] Hard timeout for shutdown. Stop the loop even when there are alive connections after the grace period.
- [x] Show socket-bound address in logs.
- [x] Fix bug, not showing the message passed as a parameter.
- [x] Event listeners should listen to `CRTL+c` signal. Temporary solution until we implement:
  - https://github.com/torrust/torrust-tracker/issues/1405.
- [x] Add a job manager to give a name to all jobs and identify them when the app shuts down.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4684,7 +4684,6 @@ dependencies = [
  "bittorrent-udp-tracker-core",
  "chrono",
  "clap",
- "futures",
  "local-ip-address",
  "mockall",
  "rand 0.9.1",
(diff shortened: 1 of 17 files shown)


josecelano:
New messages during the shutdown:

image

josecelano:
ACK 8f42271efad18c7fe3563440156fee8b0f719d7f

Issue #1358, opened 2025-03-05
Overhaul stats: Emit events from the `torrent-repository` package and collect metrics


josecelano:
### Context

I've been working on this issue https://github.com/torrust/torrust-tracker/issues/1264 recently.

The issue was a bug with the number of downloads for a given torrent in stats.

If you enable it in the configuration the number of downloads for all torrents is persisted into the database.

The tracker must:

- Update the value in the database every time the counter is increased.
- Load the value from the database every time a torrent is added to the torrent repository (active torrent) because a peer is announcing that torrent.

Torrent are removed from the in-memory torrent repository when:

- The tracker is restarted.
- The torrent is removed because it does not have any peer (peerless torrent). You can enable a cronjob to clean peerless torrents periodically.

### Problem

Internally all [torrent repository implementations](https://github.com/torrust/torrust-tracker/tree/develop/packages/torrent-repository) keep torrent metrics like:

- `complete`: number of seeders.
- `incomplete`: number of leechers.
- `downloaded`: the number of peers that have ever completed downloading.

The following is the one we are using in production:

- I've omitted some parts
- We keep the other implementation only to document discarded implementation and to be able to run benchmark again with newer version.

```
pub struct CrossbeamSkipList<T> {
    pub torrents: SkipMap<InfoHash, T>,
}

impl Repository<EntryMutexStd> for CrossbeamSkipList<EntryMutexStd>
where
    EntryMutexStd: EntrySync,
    EntrySingle: Entry,
{
    fn upsert_peer(&self, info_hash: &InfoHash, peer: &peer::Peer, opt_persistent_torrent: Option<PersistentTorrent>) -> bool {
        if let Some(existing_entry) = self.torrents.get(info_hash) {
            existing_entry.value().upsert_peer(peer)
        } else {
            let new_entry = if let Some(number_of_downloads) = opt_persistent_torrent {
                EntryMutexStd::new(
                    EntrySingle {
                        swarm: PeerList::default(),
                        downloaded: number_of_downloads,
                    }
                    .into(),
                )
            } else {
                EntryMutexStd::default()
            };

            let inserted_entry = self.torrents.get_or_insert(*info_hash, new_entry);

            inserted_entry.value().upsert_peer(peer)
        }
    }

    // ...

    fn get_metrics(&self) -> TorrentsMetrics {
        let mut metrics = TorrentsMetrics::default();

        for entry in &self.torrents {
[...]

PR #1492, opened 2025-04-30
Overhaul stats: Simplify torrent-repository package before adding events

### LINKED ISSUE #1491: Overhaul stats: Simplify `torrent-repository` package before adding events
Relates to: https://github.com/torrust/torrust-tracker/issues/1358

[I will add events to the torrent-repository package](https://github.com/torrust/torrust-tracker/issues/1358). Before doing it, I will make some changes to make the package easier to maintain.

The main problem is that we have numerous implementations of the repository with different data structures, primarily for testing performance. However, only one is used in production. There is no way to use the others because:

1. They can not be injected dynamically.
2. The alternatives are not ready for production. They don't have some of the features that were added to the one we use in production after running the performance tests. You can search for this string:

```
// todo: load persistent torrent data if provided
```

To find what was not implemented in the alternative implementations.

Since I'm going to introduce events (as we did for other packages, such as `http-tracker-core`, `udp-tracker-core`, and `udp-tracker-server`) that will require significant changes to the traits and implementations, it would be overkill to maintain all the implementations.

I suggest:

- Isolating the implementations not used in production and keeping them only for the record or executing future benchmarking tests again.
- Import the `bittorrent_tracker_core::torrent::repository::in_memory::InMemoryTorrentRepository` type from the `tracker-repo`. I think it belongs to this package since it's only one of the implementations of the torrent repository (the one used in production).
- Simplify the repository using concrete types instead of generics.

cc @da2ce7

josecelano:
Simplify `torrent-repository` package [before adding events](https://github.com/torrust/torrust-tracker/issues/1358). It would be overkill to maintain all the implementations.

### Subtasks

- [x] Copy benchmarking code to a new workspace package `torrent-repository-benchmarking`.
- [x] Remove code not used in production from the `torrent-repository` package.
  - [x] Unused repositories
  - [x] Unused entries
- [x] Remove unused traits `RepositoryAsync`, `EntryAsync`, `Entry`, `EntrySync`, etc.
- [x] Remove generics.
- [x] Rename main types
- [x] ~~Import the `bittorrent_tracker_core::torrent::repository::in_memory::InMemoryTorrentRepository` type from the `tracker-repo` package.~~ **DISCARDED**. See https://github.com/torrust/torrust-tracker/issues/1358#issuecomment-2846665573.
- [x] Clean tests


josecelano:
ACK 7215f6e4d9073a781dc2bf2d4a99a5b629530567

Issue #1495, opened 2025-05-02
Overhaul stats: Improve torrent-repository package before adding events


josecelano:
[I've been working on simplifying the `torrent-repository` package](https://github.com/torrust/torrust-tracker/issues/1491). While working on that, I have identified some things that can be improved or fixed.

I will do it before adding the events. Otherwise, it would be a bigger effort to do it later.

I will group the subtasks into three groups, one for each of the main types in the package:

- `PeerList`
- `TrackedTorrent`
- `TorrentRepository`

### Refactoring `PeerList`

- Rename `PeerList` to `PeerAnnouncementMap`?
  - The Peer is the latest peer announcement, not only the IP.
  - It's not implemented with a list, although I chose list as a generic list, not because it's implemented with a "[list](https://en.wikipedia.org/wiki/List_(abstract_data_type))"
- Use the peer socket address instead of peer ID to avoid peers with duplicate socket addresses.
- Use a cache for seeders and leechers counters to avoid iterating the list.
- Increase test coverage.

### Refactoring `TrackedTorrent`

- Document `meets_retaining_policy` behavior.
- Consider moving the `number_of_downloads_increased` counter to `PeerList`.
- Add unit tests.

### Refactoring `TorrentRepository`

- Remove `LockTrackedTorrent` trait and return the error to upper layers.
- Add a method for `len()` or `is_empty()`
- Fix Concurrency edge case in `upsert_peer`

Replace:

```rust
if let Some(existing_entry) = self.torrents.get(info_hash) {
    // ...
} else {
    let inserted_entry = self.torrents.get_or_insert(...)
    // ...
}
```

With:

```rust
let inserted_entry = self.torrents.get_or_insert_with(*info_hash, || {
    TrackedTorrentHandle::new(TrackedTorrent { ... }.into())
});
let mut torrent_guard = inserted_entry.value().lock_or_panic();
torrent_guard.upsert_peer(peer)
```

- Optimization: Early `get()` return in `upsert_peer`

Replace:

```rust
if let Some(existing_entry) = self.torrents.get(info_hash) {
    // mutate entry
} else {
    // insert and mutate
}
```

With:

```rust
let entry = self.torrents.get_or_insert_with(*info_hash, || {
    TrackedTorrentHandle::new(
        TrackedTorrent {
            swarm: PeerList::default(),
            downloaded: opt_persistent_torrent.unwrap_or_default(),
        }
        .into(),
    )
});

[...]

PR #1497, opened 2025-05-05
Overhaul stats: Improve torrent-repository package before adding events (part 1)

josecelano:
Partially implements https://github.com/torrust/torrust-tracker/issues/1495

Improve `torrent-repository` package before adding events.

### Sutasks

- [x] Update README 

#### Refactoring `PeerList`

- [x] Rename `PeerList` to `Swarm`.
- [x] Use `SocketAddr` as key for peers in `Swarm` instead of the `PeerId`.
- [x] Use a cache for seeders and leechers counters to avoid iterating the list.
- [x] Move logic from `TrackerTorrent` to `Swarm`

#### Refactoring `TrackedTorrent` (it was removed)

- [x] Remove the `downloaded` field from `TrackedTorrent`. The logic can be moved to `Swarm`.
- [x] Document `meets_retaining_policy` method behavior. I don't know why peerless torrents are kept when persisted stats (`downloaded` field) are enabled (only if the torrent has been downloaded at least once).

It was removed because after the refactorings, it became just a wrapper over `Swarm`.

#### Refactoring `TorrentRepository`

To be implemented in a new PR.

### New state

There are only two primary types:

- `Swarm`: a collection of peers. Peers are members of the swarm.
- `Swarms`: a collection of swarms.

```
$ tree ./src/
./src/
├── lib.rs
├── swarm.rs
└── swarms.rs

1 directory, 3 files
```


josecelano:
ACK 728de220693828e056b8f5069ddff19589b6825a

Issue #1502, opened 2025-05-07
Overhaul stats: fix bug in downloads counter


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/pull/1497/commits/34c159a161b7c167730f6c139dd3cb608173d37a

A couple of days ago, I made a change in [this commit](https://github.com/torrust/torrust-tracker/pull/1497/commits/34c159a161b7c167730f6c139dd3cb608173d37a).

I changed the `Swarm::meets_retaining_policy` method from:

```
/// Returns true if the torrents meets the retention policy, meaning that
/// it should be kept in the tracker.
pub fn meets_retaining_policy(&self, policy: &TrackerPolicy) -> bool {
    if policy.persistent_torrent_completed_stat && self.metadata().downloaded > 0 {
        return true;
    }

    if policy.remove_peerless_torrents && self.is_empty() {
        return false;
    }

    true
}
```

To:

```
pub fn meets_retaining_policy(&self, policy: &TrackerPolicy) -> bool {
    !(policy.remove_peerless_torrents && self.is_empty())
}
```

I thought this code was not needed:

```rust
if policy.persistent_torrent_completed_stat && self.metadata().downloaded > 0 {
    return true;
}
```

However, it's needed.

One of the metrics returned by the tracker API is the **total number of downloads for all torrents**.

```json
{
  "torrents": 320961,
  "seeders": 189885,
  "completed": 975119, <- this
  "leechers": 231044,
  ...
}
```

That metric is always stored in memory but can optionally persist into the database.

It's important to highlight that the metric represents:

- The total number of downloads for **ALL** torrents ever, when the metric is persisted.
- The total number of downloads for **ALL** torrents since the tracker started, when the metric is not persisted.

It could be mixed up with another internal metric (not exposed via the API), which is the same counter but only for ONE swarm (one torrent).

- The total number of downloads for **ONE** concrete torrent ever, when the metric is persisted.
- The total number of downloads for **ONE** concrete torrent since the tracker started, when the metric is not persisted.

The bug affects the first metric. The exposed via the API.

The problem is that this feature conflicts with removing the peerless torrents.

When removing the peerless torrents config option is enabled, the counter is lost unless it is persisted. Becuase the counter values are stored in the "Swarm" together with the list of peers.

[...]

PR #1503, opened 2025-05-07
Overhaul stats: fix bug in downloads counter

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1502

Revert the change that introduced a bug in [this commit](https://github.com/torrust/torrust-tracker/pull/1497/commits/34c159a161b7c167730f6c139dd3cb608173d37a).

Issue #1502 has been partially implemented.

- Bug fixed.
- Unit tests added to avoid regression.

In a future PR, I will implement a change to avoid keeping the swarms in memory just to track the number of downloads for that torrent.

### DIFF
--- a/packages/torrent-repository/src/swarm.rs
+++ b/packages/torrent-repository/src/swarm.rs
@@ -191,10 +191,20 @@ impl Swarm {
     }
 
     /// Returns true if the swarm meets the retention policy, meaning that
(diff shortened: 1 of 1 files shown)


josecelano:
ACK 32a37d148ca1258d47a9bcf3fbbfb3a3d99a1ba8

PR #1509, opened 2025-05-08
Fix wrong number of downloads when the tracker starts

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1502

As described [here](https://github.com/torrust/torrust-tracker/issues/1502#issuecomment-2863479464), when you start the tracker and stats persistence is enabled, you should see the last value for the number of downloads. Instead, you see something like this:

image

This PR fixes that problem.

It fixes the bug described in https://github.com/torrust/torrust-tracker/issues/1502. However, it does not fix other problems described in the issue (problem 3). I will open a new PR to fix problem 3.

### DIFF
--- a/packages/torrent-repository/src/swarms.rs
+++ b/packages/torrent-repository/src/swarms.rs
@@ -231,7 +231,7 @@ impl Swarms {
             inactive_peers_removed += removed;
         }
 
-        tracing::info!("Inactive peers removed: {inactive_peers_removed}");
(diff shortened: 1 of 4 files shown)


josecelano:
ACK 632185bf2237affb044f5c5f32c458061b460f40

josecelano:
[This will be deployed after merging to the demo tracker](https://github.com/torrust/torrust-demo/issues/47).

PR #1511, opened 2025-05-09
Disable torrent stats importation at start

### LINKED ISSUE #1510: Disable stats importation when the trackers starts
I was working on this issue:

- https://github.com/torrust/torrust-tracker/issues/1502

And one of the problems is that the downloads counter (for all torrents) is not initialised with the latest value when the tracker starts. The current way to keep the counter is to load from the database the number of downloads per torrent. It was implemented only when the torrent is added to the tracker but not when the tracker starts. This gives you a wrong general counter because inactive torrent downloads are not counted. I changed the code to fix it in this PR:

- https://github.com/torrust/torrust-tracker/pull/1509

But after deploying the patch to the demo tracker, the server could not start (unhealthy tracker container).

The reason was that there are **1.893.901.614 torrents**!!! in the demo tracker. The database is 17G: 

```console
ls -alh storage/tracker/lib/database/sqlite3.db
-rwxrwx--- 1 torrust torrust 17G May  8 17:53 storage/tracker/lib/database/sqlite3.db
```

That was probably the reason why the function to import torrents from the database (`pub fn import_persistent(&self, persistent_torrents: &PersistentTorrents)`) was not being used. Maybe it was disabled at some point because we realised it was not feasible with many torrents. However, when it was disabled, we didn't realise that it affected the downloads counter when the tracker is restarted.

I will disable it. Anyway, I was planning to implement an alternative solution [here](https://github.com/torrust/torrust-tracker/issues/1502#issuecomment-2863479464). The alternative solution does not work either because we have to run a query to sum up torrents, which will probably be costly. I guess we need to store directly the global counter (persistent metric). I will think of a new solution and update [that issue](https://github.com/torrust/torrust-tracker/issues/1502).

cc @da2ce7

josecelano:
When the tracker starts, if stats persistence is enabled, all torrents that have ever been downloaded are loaded into memory (`Swarms` type) with their download counter.

That's the current way to count all downloads and expose that metric.

However, it does not work with **millions of torrents** (like in the tracker demo) becuase:

- It's too slow.
- It consumes too much memory (all torrents that have ever been downloaded have to be loaded).

A new solution is needed to keep that metric, but in the meantime, this disables that feature, producing these effects:

- Non-accurate value for downloads when the tracker is restarted.
- Increasing indefinitely the number of torrents in memory even if the "remove peerless torrents" policy is enabled (becuase this feature overrides that policy and peerless torrents are kept in memory).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -61,7 +61,12 @@ pub async fn start(config: &Configuration, app_container: &Arc<AppContainer>) ->
 async fn load_data_from_database(config: &Configuration, app_container: &Arc<AppContainer>) {
     load_peer_keys(config, app_container).await;
     load_whitelisted_torrents(config, app_container).await;
-    load_torrents_from_database(config, app_container);
+    // todo: disabled because of performance issues.
+    // The tracker demo has a lot of torrents and loading them all at once is not
+    // efficient. We also load them on demand but the total number of downloads
+    // metric is not accurate because not all torrents are loaded.
+    // See: https://github.com/torrust/torrust-tracker/issues/1510
+    //load_torrents_from_database(config, app_container);
 }
 
 async fn start_jobs(config: &Configuration, app_container: &Arc<AppContainer>) -> JobManager {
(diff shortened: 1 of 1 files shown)


josecelano:
ACK cb487f36588c681988f5f4c75eacf87a8539dc1c

PR #1528, opened 2025-05-19
Overhaul stats: add event metrics to `torrent-repository` package

### LINKED ISSUE #1522: Overhaul stats: add event metrics to `torrent-repository` package
The events triggered by the torrent-repository package are:

Torrent events:

- `TorrentAdded`
- `TorrentRemoved`

Peer events:

- `PeerAdded`
- `PeerRemoved`
- `PeerUpdated`
- `PeerDownloadCompleted`

And the metrics:

- `torrent_repository_torrents_total`: The total number of torrents.
- `torrent_repository_torrents_downloads_total`: The total number of torrent downloads (since the tracker process started).
- `torrent_repository_peer_connections_total`: The total number of peer connections (one connection per torrent).

See https://github.com/torrust/torrust-tracker/issues/1521

I think it would be convenient to add one counter for each metric:

- `torrent_repository_torrents_added_total`: The total number of torrents added.
- `torrent_repository_torrents_removed_total`: The total number of torrents removed.
- `torrent_repository_peers_added_total`
- `torrent_repository_peers_removed_total`
- `torrent_repository_peers_updated_total`

I want to add them because it would give the admins more information about how the tracker is being used that is not included in the other metrics.

For example, every 15 seconds, 1M torrents can be added and 1.9M torrents removed. If you look only at the current number of torrents, it might look like a flat line of 100K torrents, but it does not represent the high load. With these new metrics, you can better compare the percentage of stable torrents vs. temporary torrents.

- Stable torrents: torrents that always have some peers interested in them.
- Temporary torrents: torrents that are downloaded occasionally.

**IMPORTANT:** These metrics will not be persisted. If we want to keep the counters across tracker restarts, we can add the same metrics in the tracker-core, where we handle persistence.

I don't think this will affect performance. It's easy to implement, and I think the information these metrics can provide is valuable for admins.

### How to test

1. Run the tracker with `cargo run`. 
2. Use the tracker client to make an announce request.

  ```
  cargo run -p torrust-tracker-client --bin udp_tracker_client announce udp://127.0.0.1:6969 443c7602b4fde83d1154d6d9da48808418b181b6 | jq
  ```
3. Use the REST API to get the metric value.

  ```
  curl -s "http://localhost:1212/api/v1/metrics?token=MyAccessToken&format=prometheus" | grep torrent_repository_torrents_added_total
  ```

josecelano:
These new metrics just count the number of times events have occurred.

### DIFF
--- a/packages/torrent-repository/src/statistics/event/handler.rs
+++ b/packages/torrent-repository/src/statistics/event/handler.rs
@@ -8,7 +8,9 @@ use torrust_tracker_primitives::DurationSinceUnixEpoch;
 use crate::event::Event;
 use crate::statistics::repository::Repository;
 use crate::statistics::{
-    TORRENT_REPOSITORY_PEER_CONNECTIONS_TOTAL, TORRENT_REPOSITORY_TORRENTS_DOWNLOADS_TOTAL, TORRENT_REPOSITORY_TORRENTS_TOTAL,
+    TORRENT_REPOSITORY_PEERS_ADDED_TOTAL, TORRENT_REPOSITORY_PEERS_REMOVED_TOTAL, TORRENT_REPOSITORY_PEERS_UPDATED_TOTAL,
+    TORRENT_REPOSITORY_PEER_CONNECTIONS_TOTAL, TORRENT_REPOSITORY_TORRENTS_ADDED_TOTAL,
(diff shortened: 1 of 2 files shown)


josecelano:
ACK b11af88ee3981faa92d26f64b6216d56ec1ff473

PR #1530, opened 2025-05-20
Overhaul stats: add peer inactivity metrics to `torrent-repository` package

### LINKED ISSUE #1523: Overhaul stats: add peer inactivity metrics to `torrent-repository` package
Relates to: https://github.com/torrust/torrust-tracker/issues/1522

[A user asked for metrics about peer inactivity](https://github.com/torrust/torrust-tracker/issues/1368). It was also described here https://github.com/torrust/torrust-tracker/issues/1437

Considering the recent changes and conventions for the metrics names I think we can add a new metric to the `torrent-repository` package called:

- `torrent_repository_inactive_peers_total`

We can use metric label as in other metrics for the peer role (seeder or leecher)

### How to implement it

Create a new cronjob to iterate over all peers every N seconds, get the number of inactive peers and update the metric.

1. Add the new cronjob (it can be copied from the cleanup cronjob) 
2. Execute it every N seconds. It can be a configuration option like the other crobjob.
3. Iterate over all peers in all swarms, counting inactive peers.
4. Update the gauge metric.

**NOTES:**

I'm proposing a new cronjob because:

- The frequency for this cronjob should be higher than the cronjob to clean torrents (the cleanup torrent job is executed every 10 minutes by default). I would update the metric every 15 seconds, which is the interval that Prometheus is polling the API.
- In the cleanup torrent cronjob, all inactive peers are removed, so the new metric would always be zero after running the cronjob. We could update the metric before deleting inactive peers, so we would know how many peers are inactive just before removing them.
- I prefer to keep both features decoupled. Adding a new cronjob is very easy.
- We should not update the metric directly in the Swarms or Swarm types. Metrics are decoupled from the core logic by using events. I would do it in the cronjob main loop.
- I have been thinking of an alternative solution using events. For example, the cronjob can trigger a `PeerDisconnected` (or `InactivePeer`) event. However, it seems to be more complex because peers are added and removed constantly. WE only want to know the total number of inactive peers still in memory. When a peer disconnects, we have to increase the counter, but we have to decrease it if the peer is removed. The hard part is to decrease the inactive peers counter if one of the inactive peers has announced again before it was removed. We have to check that the peer was considered inactive, although not removed yet. I think all that logic is more complex than the solution I've described.

josecelano:
This adds two new metrics:

- [x] `torrent_repository_inactive_peers_total`: The number of inactive peers across all torrents.
- [x] `torrent_repository_inactive_torrents_total`: The number of inactive torrents (inactive swarms).


josecelano:
ACK f3056efcf980907c01e0bd73dca4378e5c434d42

josecelano:
ACK 677deacdc419526122eff62973f2685ac976a5eb

PR #1531, opened 2025-05-21
Overhaul stats: Simplify how the `AnnounceHandler` persist number of downloads in `tracker-core` package

### LINKED ISSUE #1524: Overhaul stats: simplify how the `AnnounceHandler` persist number of downloads in `tracker-core` package
Now that the `torrent-repository is emitting events, we can simplify the announce handler in the tracker-core package.

The `bittorrent_tracker_core::announce_handler::AnnounceHandler` has a method to handle the announce request:

```rust
    pub async fn announce(
        &self,
        info_hash: &InfoHash,
        peer: &mut peer::Peer,
        remote_client_ip: &IpAddr,
        peers_wanted: &PeersWanted,
    ) -> Result<AnnounceData, AnnounceError> {
        self.whitelist_authorization.authorize(info_hash).await?;

        let opt_persistent_torrent = if self.config.tracker_policy.persistent_torrent_completed_stat {
            self.db_torrent_repository.load(info_hash)?
        } else {
            None
        };

        let number_of_downloads_increased = self
            .in_memory_torrent_repository
            .upsert_peer(info_hash, peer, opt_persistent_torrent)
            .await;

        if self.config.tracker_policy.persistent_torrent_completed_stat && number_of_downloads_increased {
            self.db_torrent_repository.increase_number_of_downloads(info_hash)?;
        }

        Ok(self.build_announce_data(info_hash, peer, peers_wanted).await)
    }
```

When `persistent_torrent_completed_stat` is enabled, it updates a DB counter with the number of downloads for that torrent.

I proposed a change [here](https://github.com/torrust/torrust-tracker/issues/1358).

We can now:

- Listen for events from the `torrent-repository` package.
- Filter the event `PeerDownloadCompleted`.
- When we received that event, we call the `db_torrent_repository.increase_number_of_downloads(info_hash)` function.

The downloads will be incremented asynchronously.

**Pros:**

- Response should be faster; the peer does not have to wait until we increase the counter in the DB.
- The announce method will be simpler. It does not have to handle metrics.
- We can simplify the `upsert_peer` method. We don't need to return the boolean to know if downloads were increased.
- The EventBus acts as a buffer to write to the database (to increase counters). We could use batches to increase many counters with one SQL sentence.

**Cons:**

- It will be harder to follow what's happening by only reading the `announce` method. Anyway, we are intensively using events everywhere.
- It can be inefficient to listen to all `torrent-repository` events. Ideally we should be able to listen only to the `PeerDownloadCompleted` event. But I think that would not be possible if we don't use another EventBus.

josecelano:
Simplify how the `AnnounceHandler` persist number of downloads in `tracker-core` package.

It's now done asynchronously instead of synchronously.

**NOTICE:** That `bittorrent_tracker_core::announce_handler::AnnounceHandler::announce` will not return a database error anymore, meaning the peer will receive the announce response even if the tracker can't update the downloads counter.

### Subtasks

Main change:

- [x] Scaffolding for listening to torrent-repository events in the `tracker-core` package.
- [x] Move the SQL update from the `AnnounceHandler` to the event handler (`handle_event`).

Cleanup:

- [x] Fix test. https://github.com/torrust/torrust-tracker/actions/runs/15168471742/job/42652507081?pr=1531#step:7:250
  - It's not the service's responsibility to increment the counter anymore. Add:
    - A unit test in the event handler to check that the counter is increased when the event is received.
    - An integration test to simulate a peer completing downloading.
    - Maybe we should also add a main app E2E test (to make sure the listener is initialised), but we are not testing it for other jobs directly. We have to review the strategy for E2E tests in the main app.
- [x] Refactor: rename `upsert_peer` to `handle_announcement`.
- [x] Refactor: rename `AnnounceHandler::announce` to `AnnounceHandler::handle_announcement`
- [x] Refactor: Don't return the boolean `number of downloads increased` in `upsert_peer`. `upsert_peer` should only be a command decoupled from other secondary tasks. Side effects should be captured from events.

NOTE: I cannot remove the `db_torrent_repository` dependency from the `AnnounceHandler` yet until we implement the new persistent metric for the number of downloads, because it's still used in:

```rust
let opt_persistent_torrent = if self.config.tracker_policy.persistent_torrent_completed_stat {
    self.db_torrent_repository.load(info_hash)?
} else {
    None
};
```


josecelano:
The main change is done. I will clean a little bit before merging.

josecelano:
ACK 67d177b6d4af24608d6be5a80ed10434242c4cd4

PR #1552, opened 2025-06-02
Refactor UDP error event logic before adding new metrics

### LINKED ISSUE #1551: Overhaul stats: Refactor logic related to UDP errors
I will make some changes before adding [the new metric](https://github.com/torrust/torrust-tracker/issues/1456).

josecelano:
I will add a new metric to count which client's software is used when a UDP error is produced (only for connection ID errors, for now).

This PR make some changes before adding the new metric.

### Subtasks

- [x] Wrapper for aquatic parse error. It's not sendable.
- [x] Add the error to the `Event::UdpError` event.
- [x] Move logic to increase the number of wrong connection IDs per IP to the event listener.
- [x] Rename errors.
- [x] Refactor event handler.


josecelano:
ACK a8f3a973c661815b7721d87cc24b828915d0deec

PR #1558, opened 2025-06-04
Add new metric label `server_binding_address_type`

### LINKED ISSUE #1375: Overhaul stats: Increase IPv4 stats for IPv6 IPs with the prefix `::ffff:`
Relates to: https://github.com/torrust/torrust-tracker/issues/1368

When you bind the server socket address to IPv6 IP addresses (`[::]:6969` in the `bind_address` config option), the tracker can handle requests from IPv4 and IPv6 clients. IPv4 clients' addresses are converted to IPv6 IPs.

However, the metrics are increased for IPv6. I think we can add a config option to change the behaviour and increase IPv4-related metrics when the IPv6 address has the `::ffff:` prefix.

I would make it the default behaviour, but that's a breaking change. So the new config option should be disabled.

### IPv4-mapped notation

Image

cc @da2ce7 @devnakx

josecelano:
Add new metric label `server_binding_address_type`.

- Label name: `server_binding_address_type`
- Label values: `plain`, `v4_mapped_v6`

Usage example in Prometheus format:

```
udp_tracker_server_requests_accepted_total{request_kind="connect",server_binding_address_type="plain",server_binding_ip="0.0.0.0",server_binding_port="6969",server_binding_protocol="udp"} 1
```

Example of IPv4-mapped-IPv6 IP: `[::ffff:192.0.2.33]`

It should be included whenever another IP label value is used.

cc @devnakx

### DIFF
--- a/packages/http-tracker-core/src/event.rs
+++ b/packages/http-tracker-core/src/event.rs
@@ -87,6 +87,10 @@ impl From<ConnectionContext> for LabelSet {
                 LabelValue::new(&connection_context.server.service_binding.bind_address().ip().to_string()),
             ),
             (
+                label_name!("server_binding_address_type"),
+                LabelValue::new(&connection_context.server.service_binding.bind_address_type().to_string()),
+            ),
+            (
                 label_name!("server_binding_port"),
                 LabelValue::new(&connection_context.server.service_binding.bind_address().port().to_string()),
             ),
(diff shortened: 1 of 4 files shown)


josecelano:
ACK d4c43bd3a5bc75704d8b8a5b4641f273968aceb4

PR #1566, opened 2025-06-09
Add a new metric to count peers reverting `Completed` state

### LINKED ISSUE #1534: Overhaul stats:  Clients can artificially inflate download counter by repeating announce and completed events
### Problem

Currently, the tracker increments the downloads counter each time it receives an `announce` request with `Started` event followed by a `Completed` event from a client. There is no mechanism in place to verify whether a peer has already reported the `Completed` event for a given torrent. As a result, a malicious or misbehaving client can repeatedly send `announce` requests with `Started` and `completed` events, artificially inflating the downloads counter for a torrent.

### Steps to Reproduce

1. A client sends an `announce` request for a torrent  with the `event=started` parameter.
2. The client sends an `announce` request with the `event=completed` parameter.
3. The client repeats steps 1 and 2 multiple times.
4. Each repetition increases the downloads counter, regardless of whether the peer has already completed the download.

See https://github.com/torrust/torrust-tracker/issues/1532

### Expected Behavior

The tracker should only increment the downloads counter the first time a peer reports a `completed` event for a given torrent. Subsequent `completed` events from the same peer for the same torrent should be ignored.

### Impact

- Download statistics are unreliable and can be manipulated.
- This can affect tracker analytics, ranking, and trust in reported data.

### Suggested Solution

TODO

cc @da2ce7

josecelano:
Add a new metric to count peers reverting the `Completed` state.

### Subtasks

- Refactor `swarm-coordination-registry` package: rename types, variables, and constants to follow the package's new name.
- Add the new metric: `swarm_coordination_registry_peers_completed_state_reverted_total`

### DIFF
--- a/packages/axum-http-tracker-server/src/environment.rs
+++ b/packages/axum-http-tracker-server/src/environment.rs
@@ -144,13 +144,13 @@ impl EnvContainer {
             .expect("missing HTTP tracker configuration");
         let http_tracker_config = Arc::new(http_tracker_config[0].clone());
 
-        let torrent_repository_container = Arc::new(SwarmCoordinationRegistryContainer::initialize(
+        let swarm_coordination_registry_container = Arc::new(SwarmCoordinationRegistryContainer::initialize(
             configuration.core.tracker_usage_statistics.into(),
         ));
 
         let tracker_core_container = Arc::new(TrackerCoreContainer::initialize_from(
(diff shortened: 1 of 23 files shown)


josecelano:
ACK 7be03663946dcbcf2f1ee28d62b1c8d30741cd42

PR #1577, opened 2025-06-11
Overhaul stats: Refactor metrics, use label metrics to calculate global metrics

### LINKED ISSUE #1446: Overhaul stats: Refactor metrics (part 2)
Depends on: https://github.com/torrust/torrust-tracker/issues/1445

The API exposes the "old" metrics (`/stats`) and the new ones (`/metrics`)

Internally, we use only one repository to store both:

For example, in `packages/http-tracker-core/src/statistics/metrics.rs`:

```
pub struct Metrics {
    /// Total number of TCP (HTTP tracker) `announce` requests from IPv4 peers.
    pub tcp4_announces_handled: u64,

    /// Total number of TCP (HTTP tracker) `scrape` requests from IPv4 peers.
    pub tcp4_scrapes_handled: u64,

    /// Total number of TCP (HTTP tracker) `announce` requests from IPv6 peers.
    pub tcp6_announces_handled: u64,

    /// Total number of TCP (HTTP tracker) `scrape` requests from IPv6 peers.
    pub tcp6_scrapes_handled: u64,

    /// A collection of metrics.
    pub metric_collection: MetricCollection,
}

impl Metrics {
    pub fn increase_counter(&mut self, metric_name: &MetricName, labels: &LabelSet, now: DurationSinceUnixEpoch) {
        self.metric_collection.increase_counter(metric_name, labels, now);
    }

    pub fn set_gauge(&mut self, metric_name: &MetricName, labels: &LabelSet, value: f64, now: DurationSinceUnixEpoch) {
        self.metric_collection.set_gauge(metric_name, labels, value, now);
    }
}
```

I kept global and new metrics because I didn't want to make breaking changes in the API. However, we can remove the old ones internally and calculate them from the new ones only when the API replies to a `/stats` request.

That would have these benefits:

- Simplify the `bittorrent_http_tracker_core::statistics::event::handler`
- Improve performance; the tracker does not have to maintain both metrics.

The cons are:

- The old `/stats` endpoint will be slower
- We must implement new methods in the `MetricColelction` type to calculate aggregate data. These methods could be helpful for other tasks in the future.

josecelano:
The idea is to continue exposing the `stats` endpoint but removing the internal code to calculate them. Instead, we calculate the metrics from the labeled metrics.

That will **simplify** the event listeners and **avoid duplicate code** to calculate the same data.

### Endpoints for statistics

`stats` (simple predefined global metrics)

- http://localhost:1212/api/v1/stats?token=MyAccessToken&format=json
- http://localhost:1212/api/v1/stats?token=MyAccessToken&format=prometheus

`metrics` (extendable-labelled metrics)

- http://localhost:1212/api/v1/metrics?token=MyAccessToken&format=json
- http://localhost:1212/api/v1/metrics?token=MyAccessToken&format=prometheus

### Subtasks

- [x] Add a new method to `MetricCollection` to sum the metric's samples.
- [x] Review test coverage.
- [x] Use the new method to calculate metrics in the `stats` endpoint.

### After merging this PR

- Deploy to the tracker demo and verify that both Grafana dashboards appear identical.
- Delete unused code to calculate the metrics in the `stats` endpoint directly in the event listeners.
- Redeploy to the tracker demo and re-verify.

**NOTICE:** We have two Grafana dashboards, one collecting data from the `stats` endpoint and another collecting data from `metrics`, and they look identical now. If there is a bug in the way we calculate aggregate values in the `MetricCollection`, we should see different graphs in the dashboards.

Both dashboards (built from stats on the left, built from metrics on the right):

image

**NOTICE:** Also, the new methods will be helpful for external users once we publish the metrics crate. Consumers of the Torrust Tracker API (`metrics` endpoint) will be able to deserialize and calculate aggregate values from the response.


josecelano:
image

josecelano:
I've been considering removing the traits:

`src/metric_collection/aggregate.rs`:

```rust
pub trait Sum {
    fn sum(&self, metric_name: &MetricName, label_set_criteria: &LabelSet) -> Option<AggregateValue>;
}
```

`src/metric/aggregate/sum.rs`:

```rust
pub trait Sum {
    fn sum(&self, label_set_criteria: &LabelSet) -> AggregateValue;
}
```

I wanted to keep MetricCollection and Metric types unaware (decouple) from the aggregate functions. So:

- You can add more aggregate functions without adding more methods.
- Aggregate functions don't need access to the internals of those types. They only need to get the values (iterate).

However, I tried to do the refactor, and I ended up with something like:

```rust
/// Calculate the sum of metric samples for a given metric name and label criteria.
/// 
/// This function follows the Open/Closed principle by allowing new aggregate functions
/// to be added without modifying the MetricCollection type.
pub fn sum(collection: &MetricCollection, metric_name: &MetricName, label_set_criteria: &LabelSet) -> Option<AggregateValue> {
    // Direct implementation without trait dependency
    if let Some(counter_collection) = collection.counters.metrics.get(metric_name) {
        return Some(counter_collection.sum(label_set_criteria));
    }
    
    if let Some(gauge_collection) = collection.gauges.metrics.get(metric_name) {
        return Some(gauge_collection.sum(label_set_criteria));
    }
    
    None
}
```

The problem with this approach is that the aggregate function will be broken as soon as we add a new type of metric to the metric collection.

It doesn't work for the metric collection, but it could work for the metric. 

Anyway, since we only have one aggregate function, I will keep it as it is for now. We can reconsider other alternatives if we plan to add more aggregate functions in the future.

cc @da2ce7

josecelano:
ACK dcfb5d5d207b9fad0aceba9aa85c4497923cb33c

Issue #1589, opened 2025-06-17
Overhaul stats: Wrong UDP Average Connect Time metric


josecelano:
I've [deployed the latest changes to the tracker demo](https://github.com/torrust/torrust-demo/issues/69) and there is one bug in one of the metrics:

![Image](https://github.com/user-attachments/assets/aa9411d2-9b23-42a8-a9b3-6f8551168a3e)

The problem may be that there are connection requests to a second UDP tracker (port `6868`), but there are not `announce` and `scrape` requests.

```
# HELP udp_tracker_server_performance_avg_processing_time_ns Average time to process a UDP connect request in nanoseconds
# TYPE udp_tracker_server_performance_avg_processing_time_ns gauge
udp_tracker_server_performance_avg_processing_time_ns{request_kind="connect",server_binding_address_ip_family="inet",server_binding_address_ip_type="plain",server_binding_ip="0.0.0.0",server_binding_port="6969",server_binding_protocol="udp"} 18446743710713731000
udp_tracker_server_performance_avg_processing_time_ns{request_kind="scrape",server_binding_address_ip_family="inet",server_binding_address_ip_type="plain",server_binding_ip="0.0.0.0",server_binding_port="6969",server_binding_protocol="udp"} 40927.05282530009
udp_tracker_server_performance_avg_processing_time_ns{request_kind="announce",server_binding_address_ip_family="inet",server_binding_address_ip_type="plain",server_binding_ip="0.0.0.0",server_binding_port="6969",server_binding_protocol="udp"} 29331.002378199326
udp_tracker_server_performance_avg_processing_time_ns{request_kind="connect",server_binding_address_ip_family="inet",server_binding_address_ip_type="plain",server_binding_ip="0.0.0.0",server_binding_port="6868",server_binding_protocol="udp"} 18446743710604753000
```

### Subtasks

- [x] Fix the bug
- [x] Fix also the problem described [here](https://github.com/torrust/torrust-tracker/pull/1590#issuecomment-2983769891).
- [x] Refactor

josecelano:
After the initial debug locally, the simplest case is working:

- Metrics endpoint: http://localhost:1212/api/v1/metrics?token=MyAccessToken&format=prometheus
- Command to simulate one `connect` and `announce`: `cargo run -p torrust-tracker-client --bin udp_tracker_client announce udp://127.0.0.1:6969 000620bbc6c52d5a96d98f6c0f1dfa523a40df82 | jq`

[...]

PR #1590, opened 2025-06-18
Add logs to debug this bug: Wrong UDP Average Connect Time metric

josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1589

I am unable to reproduce the bug described [here](https://github.com/torrust/torrust-tracker/issues/1589) locally.

My plan is to:

- Add unit tests to see if I can reproduce the problem with edge cases.
- Enable debugging with tracing and redeploy to the tracker demo to collect data from the demo.

### DIFF
--- a/cSpell.json
+++ b/cSpell.json
@@ -175,6 +175,7 @@
         "trackerid",
         "Trackon",
         "typenum",
+        "udpv",
         "Unamed",
         "underflows",
         "Unsendable",
--- a/cSpell.json
+++ b/cSpell.json
@@ -34,6 +34,7 @@
         "chrono",
         "ciphertext",
         "clippy",
+        "cloneable",
         "codecov",
         "codegen",
         "completei",
--- a/packages/udp-tracker-server/src/statistics/metrics.rs
+++ b/packages/udp-tracker-server/src/statistics/metrics.rs
@@ -276,3 +276,784 @@ impl Metrics {
             .unwrap_or_default() as u64
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use torrust_tracker_clock::clock::Time;
+    use torrust_tracker_metrics::metric_name;
+
+    use super::*;
+    use crate::statistics::{
+        UDP_TRACKER_SERVER_ERRORS_TOTAL, UDP_TRACKER_SERVER_IPS_BANNED_TOTAL,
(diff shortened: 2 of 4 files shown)


josecelano:
I think the problem could be a race condition between recalculating the average and increasing the total number of requests. Since we need the total number of requests to calculate the running average, it is possible that we calculate the average before the request counter has been incremented. In that case, the initial average would be infinite. However, I have added a test and, although it's a bug, it's not the same problem I'm getting in production.

```rust
    #[tokio::test]
    async fn it_should_handle_moving_average_calculation_before_any_connections_are_recorded() {
        let repo = Repository::new();
        let now = CurrentClock::now();

        // This test checks the behavior of `recalculate_udp_avg_connect_processing_time_ns``
        // when no connections have been recorded yet. The first call should
        // handle division by zero gracefully and return an infinite average,
        // which is the current behavior.

        // todo: the first average should be 2000ns, not infinity.
        // This is because the first connection is not counted in the average
        // calculation if the counter is increased after calculating the average.
        // The problem is that we count requests when they are accepted, not
        // when they are processed. And we calculate the average when the
        // response is sent.

        // First calculation: no connections recorded yet, should result in infinity
        let processing_time_1 = Duration::from_nanos(2000);
        let avg_1 = repo.recalculate_udp_avg_connect_processing_time_ns(processing_time_1).await;

        // Division by zero: 1000 + (2000 - 1000) / 0 = infinity
        assert!(
            avg_1.is_infinite(),
            "First calculation should be infinite due to division by zero"
        );

        // Now add one connection and try again
        let ipv4_labels = LabelSet::from([("server_binding_address_ip_family", "inet"), ("request_kind", "connect")]);
        repo.increase_counter(&metric_name!(UDP_TRACKER_SERVER_REQUESTS_ACCEPTED_TOTAL), &ipv4_labels, now)
            .await
            .unwrap();

        // Second calculation: 1 connection, but previous average is infinity
        let processing_time_2 = Duration::from_nanos(3000);
        let avg_2 = repo.recalculate_udp_avg_connect_processing_time_ns(processing_time_2).await;

        assert!(
            (avg_2 - 3000.0).abs() < f64::EPSILON,
            "Second calculation should be 3000ns, but got {avg_2}"
        );
    }
```

josecelano:
ACK bf9d16a83ec48d2b60074fdc97b93f7c58bb5944

PR #1653, opened 2026-02-06
20260206 sql update example

da2ce7:
@josecelano a simple demo of what a basic update of the sql tables would look like

I can confirm that the torrents table only includes torrents that have at least one download.


josecelano:
This was related to a discussion were we thought we were persisting all the torrents statistics even for torrents with no downloads at all. However @da2ce7 confirmed that the torrents table only includes torrents that have at least one download

PR #1661, opened 2026-03-03
fix: add sleep after HTTP server stop in health check test to avoid race condition

josecelano:
The test `it_should_return_error_when_http_service_was_stopped_after_registration` was flaky on nightly Rust because the OS can keep a TCP port in `CLOSE_WAIT`/`TIME_WAIT` briefly after the server stops, causing the health check to still return `Ok` instead of `Error`.\n\nAdds a 100ms sleep after `stop().await` to let the OS fully release the TCP port before the health check runs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -626,7 +626,7 @@ checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
 
 [[package]]
 name = "bittorrent-http-tracker-core"
-version = "3.0.0-develop"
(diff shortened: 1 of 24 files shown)


josecelano:
ACK 2d65b09ec5fe74e1ed53239100f65f7e3bd29b4b

josecelano:
The testing workflow failed in this [PR](https://github.com/torrust/torrust-tracker/pull/1660).

josecelano:
The testing workflow failed in this [PR](https://github.com/torrust/torrust-tracker/pull/1660).

Issue #1525, opened 2025-05-16
Overhaul persistence


josecelano:
We should review the database layer.

### Context

- User has a request for new driver implementation: [PostgreSQL](https://github.com/torrust/torrust-tracker/issues/462)
  - [It seems adding Postgres is tricky](https://github.com/torrust/torrust-tracker/issues/462#issuecomment-2766305804) with the crate we are using: [r2d2](https://docs.rs/r2d2/latest/r2d2/)
- We don't have migrations. We are using [sqlx](https://docs.rs/sqlx/latest/sqlx/) in the Index.

### Problem

- I'm not sure if the package we are using is still the best fit for this project.
- I think we should include migrations if we start making schema changes more often, and it seems we will. For example, I'm planning to add more tables to persist metrics. See https://github.com/torrust/torrust-tracker/issues/1437#issuecomment-2887185631

### Proposal

In the middle term (before adding more tables), we should:

- Research to confirm if the current crate is the best one. Change it if it's not.
- Add migrations.
- Split the database trait into many different traits per context. Otherwise, it will end up having too many methods.

I think I will implement the metrics persistence with the current drivers because I want to finish the [stats overhaul](https://github.com/torrust/torrust-tracker/issues/1327), but we should consider implementing this epic before continuing to add more tables (complexity) to the current persistence layer, especially if we decide to change it.

cc @da2ce7

josecelano:
### Alternatives

- https://www.sea-ql.org/

josecelano:
An article about Rust, PostgreSQL, and [sqlx](https://github.com/launchbadge/sqlx) by @skerkour

https://kerkour.com/rust-postgres-everything

<img width="770" height="213" alt="Image" src="https://github.com/user-attachments/assets/446d6261-f336-461e-a406-8feeb3741ec4" />

cc @da2ce7

josecelano:
## Implementation Plan

The implementation has been broken down into 9 sequential sub-issues. Each one is independently reviewable and mergeable into `develop` before PostgreSQL support is complete.

The full specs have been merged into `develop` via PR #1702. Links below point to each spec file.

---

### Phase 1 — Make the change easy

| # | Sub-issue | Short description | Spec | Status |
|---|-----------|-------------------|------|--------|
[...]

Issue #462, opened 2023-09-20
Postgres support should be awesome addition


the-rarbg:
Postgres is also very popular anyway we could use postgres instead of mysql

gmemstr:
I dug into this a little bit and came to the conclusion that adding PostgreSQL with the r2d2 library is *tricky*, because the base `postgres` crate *wraps `tokio_postgres`*, and therefore tries to spawn a nested runtime. I figure this mean that the database layer will need refactoring to be fully async so that `tokio_postgres` can be used directly.

josecelano:
> I dug into this a little bit and came to the conclusion that adding PostgreSQL with the r2d2 library is _tricky_, because the base `postgres` crate _wraps `tokio_postgres`_, and therefore tries to spawn a nested runtime. I figure this mean that the database layer will need refactoring to be fully async so that `tokio_postgres` can be used directly.

Hi @gmemstr, thank you for the feedback. I have not done much at the database layer. I only know we were considering alternatives to r2d2:

https://github.com/torrust/torrust-tracker/discussions/616

And other changes. So I guess we need to make changes anyway for other reasons.

Have you consider other alternatives? Do you think r2d2 is still a good choice for this project?

josecelano:
Hi @the-rarbg @gmemstr, I'm working on https://github.com/torrust/torrust-tracker/issues/1525 and expect to add support for PostgreSQL in some days. It will be available in the "develop" branch. We are planning to release version 4.0.0 this year.

josecelano:
Hi @the-rarbg, @gmemstr, support for postgresql has been added in https://github.com/torrust/torrust-tracker/issues/1723. It's available in the "develop" branch and will be included in the next major release.

PR #1684, opened 2026-04-09
feat: add PostgreSQL database driver

DamnCrab:
## Summary

Add PostgreSQL as a third supported database backend, alongside the existing SQLite and MySQL drivers.

## Motivation

The tracker previously only supported SQLite and MySQL. PostgreSQL is widely used in production environments and adds an important deployment option.

### `packages/configuration`
- Added `PostgreSQL` variant to the `Driver` enum (serializes as `"postgresql"`)
- Added PostgreSQL connection string documentation and example
- Added PostgreSQL URL password masking in `mask_secrets()`
- Added unit test for PostgreSQL password masking

### `packages/tracker-core`
- Added `r2d2_postgres = "0.18"` dependency
- New `databases/driver/postgres.rs` — full `Database` trait implementation:
  - All CRUD operations: torrents, whitelist, authentication keys, global downloads
  - `ON CONFLICT DO UPDATE` upsert semantics (equivalent to SQLite/MySQL `INSERT OR REPLACE` / `ON DUPLICATE KEY UPDATE`)
- Added `PostgreSQL` variant to the tracker-core `Driver` enum
- Added `GenericConnectionError` variant to the database `Error` enum for PostgreSQL-specific connection errors
- Added `From<r2d2_postgres::postgres::Error>` conversion for `Error`
- Updated `databases/setup.rs` to handle the new `PostgreSQL` variant
- Updated module-level documentation to reference three drivers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -720,6 +720,7 @@ dependencies = [
  "mockall",
  "r2d2",
  "r2d2_mysql",
+ "r2d2_postgres",
  "r2d2_sqlite",
  "rand 0.10.0",
  "serde",
@@ -801,6 +802,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
 dependencies = [
  "generic-array",
+]
+
+[[package]]
+name = "block-buffer"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
+dependencies = [
+ "hybrid-array",
 ]
 
 [[package]]
@@ -1134,7 +1144,7 @@ version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773f3b9af64447d2ce9850330c473515014aa235e6a783b02db81ff39e4a3dad"
 dependencies = [
- "crypto-common",
+ "crypto-common 0.1.7",
  "inout",
 ]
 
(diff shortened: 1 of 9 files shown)


josecelano:
Relates to:

- https://github.com/torrust/torrust-tracker/issues/1525
- https://github.com/torrust/torrust-tracker/issues/462

Hi @DamnCrab thank you for your PR! I will review it today or tomorrow. PostgreSQL will be a really good feature for the tracker.

josecelano (changes requested):
Thank you for the PR and for the work to add PostgreSQL support. I reviewed it carefully locally and I appreciate the effort that went into getting the new driver working and covered by tests.

After reviewing the implementation, I’m going to reject this PR in its current form.

The main reason is architectural: the PostgreSQL driver currently spawns a new OS thread for every database operation in order to isolate the sync `postgres` client from the tracker’s tokio runtime. Even though the implementation uses an r2d2 connection pool, it does not use a reusable execution pool for the blocking PostgreSQL work, so every query still pays the cost of thread creation and join. I do not think that performance model is acceptable for merge.

There is also a correctness issue in the current PostgreSQL implementation: torrent counters are stored as 32-bit `INTEGER` values and converted through `i32`, which creates a real overflow/panic risk once counters exceed 2,147,483,647.

In addition, I think the PR is still missing a couple of important pieces for a complete PostgreSQL feature:
- a default PostgreSQL container config file parallel to the existing MySQL one
- proper migrations or at least a migration structure/documentation in the persistence layer

My conclusion is that PostgreSQL support should be implemented after, or together with, the persistence redesign tracked in issue #1525:

https://github.com/torrust/torrust-tracker/issues/1525

I think it will be much easier to introduce PostgreSQL properly once persistence is overhauled, rather than adding it on top of the current design with a costly runtime workaround.

I wrote a more detailed local review here for reference:

https://github.com/josecelano/torrust-tracker/blob/pr-1684-review/docs/pr-reviews/pr-1684-postgresql-driver.md

Thank you again for the contribution. I would be happy to review a follow-up version once the persistence redesign is in place or this is reworked on top of it.

PR #1692, opened 2026-04-16
Document MySQL DSN password URL-encoding requirement

### LINKED ISSUE #1606: Document DSN password URL-encoding requirement for MySQL connection string
I'm working on a [new repo](https://github.com/torrust/torrust-tracker-demo/) to automate the tracker deployment (still experimental), and I have realised we can have problems with the DB passwords automatically generated.

When configuring the tracker with a MySQL DSN (e.g. mysql://user:password@host:3306/db), reserved URL characters in the password (notably + and /) must be percent-encoded. Otherwise, the DSN may be parsed incorrectly, and the tracker fails to connect to MySQL despite correct credentials.

In the [official deployment tutorial](https://torrust.com/blog/deploying-torrust-to-production), we use SQLite, so this is not a problem.

### Why this matters

In many deployments, the DSN is built from environment variables (e.g. docker-compose). If MYSQL_PASSWORD contains reserved characters, the URL becomes invalid unless encoded first.

Standard secret generators produce base64 strings that include + and /, which are problematic in the URL password component.

### Symptoms

- Tracker fails to start or cannot connect to MySQL
- Auth errors from MySQL despite the correct user password
- Errors trace back to the DSN parsing

### Proposed improvements

1) Documentation: Add a short note in configuration docs clarifying that credentials in DSNs must be URL-encoded, with a quick example (e.g. Python urllib.parse.quote or shell snippet) and/or recommend URL-safe secrets (A–Z, a–z, 0–9, -, _).

2) Examples: Ensure docker-compose examples either:

- Use URL-safe secrets for MYSQL_PASSWORD, or
- Show how to provide an already-encoded value in the DSN

3) (Optional) Alternative config: Consider supporting separate config fields for user/password (not embedded in the DSN), to avoid URL-encoding pitfalls.

cc @da2ce7

josecelano:
## Summary
Document that MySQL DSN passwords must be percent-encoded when they contain reserved URL characters.

## Changes
- Add URL-encoding guidance to container deployment docs.
- Add URL-encoding note to the default MySQL tracker config example.
- Clarify MySQL DSN format in configuration docs and add encoding requirement for reserved characters.

## Why
Deployments commonly build DSNs from environment variables. Generated secrets can include characters like `+` and `/`, which break DSN parsing unless encoded.

Fixes #1606

### DIFF
--- a/docs/containers.md
+++ b/docs/containers.md
@@ -248,6 +248,10 @@ driver = "mysql"
 path = "mysql://db_user:db_user_secret_password@mysql:3306/torrust_tracker"
 ```
 
+Important: if the MySQL password contains reserved URL characters (for example `+`, `/`, `@`, or `:`), it must be percent-encoded in the DSN password component. For example, if the raw password is `a+b/c`, use `a%2Bb%2Fc` in the DSN.
+
+When generating secrets automatically, prefer URL-safe passwords (`A-Z`, `a-z`, `0-9`, `-`, `_`) to avoid DSN parsing issues.
+
 ### Build and Run:
 
 ```sh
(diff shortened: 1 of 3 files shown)


josecelano:
ACK bcbef65801ebc3798a6d803b57385537fc677d7e

PR #1695, opened 2026-04-17
feat: rework persistence for async sqlx backends and PostgreSQL support

DamnCrab:
## Summary

This PR reworks the tracker persistence layer to make PostgreSQL support fit the current async architecture, instead of adding PostgreSQL as a special case on top of the previous synchronous database abstraction.

The main goals were:

- remove the previously rejected synchronous PostgreSQL execution model
- split the monolithic database interface into narrower persistence capabilities
- move the SQL backends to a shared async substrate
- add PostgreSQL as a first-class backend
- widen persisted download counters to 64-bit internally while preserving BitTorrent protocol compatibility at the encoding boundary

This branch also includes the follow-up fixes from review and the regressions found while running the full local test suite.

## Why

The previous PostgreSQL attempt worked around the sync/async mismatch with a blocking execution model that is not a good fit for the rest of the project.

This rework addresses that first:

- persistence is now async-native
- SQLite, MySQL, and PostgreSQL follow the same structure
- schema ownership is centralized in migrations
- PostgreSQL is added on top of the same persistence model as the existing backends

### Persistence redesign

- replaced the old monolithic `Database` trait with narrower async traits:
  - `SchemaMigrator`
  - `TorrentMetricsStore`
  - `WhitelistStore`
  - `AuthKeyStore`
- added a `Persistence` facade so the rest of the codebase can move to the new structure without introducing a single “god object” again

### Async SQL backends

- migrated SQLite and MySQL from the previous synchronous stack to `sqlx`
- added a PostgreSQL backend using the same async approach
- removed the old blocking PostgreSQL workaround

### Schema and migrations

- migrations are now the source of truth for schema changes
- added PostgreSQL migrations
- added the matching no-op SQLite migration for the counter widening step so migration history stays aligned across backends

### Counter widening and protocol boundaries

- widened persisted download counters from `u32` to `u64`
- kept peer counts as 32-bit where that still matches their meaning
- preserved protocol compatibility by saturating only at the BitTorrent protocol boundary where 32-bit fields are required

### Follow-up fixes included in this branch

- removed a UDP health-check panic path caused by an internal `unwrap()`
- fixed async persistence timing in tests by waiting for eventual DB writes when needed
- hardened database fault-injection tests after the lazy schema initialization changes
- documented the `drop_database_tables()` contract used by tests
- improved diagnostics for persisted-download polling failures

## Validation

I ran the following locally:

- `cargo test --workspace --all-targets`
- database compatibility matrix across:
  - SQLite `3.51.0`
  - MySQL `8.0`, `8.4`
  - PostgreSQL `14`, `15`, `16`, `17`

I also ran real qBittorrent end-to-end tests against the tracker:

- SQLite + UDP
- MySQL 8.0 + HTTP
- PostgreSQL 16 + HTTP

Those runs completed real seeder/leecher transfers and ended with the expected tracker scrape state:

- `complete = 2`
- `downloaded = 1`
- `incomplete = 0`

## Benchmark Notes

I also compared the previous implementation and this branch with the same black-box workloads.

High-level results:

- SQLite is mostly neutral overall, although reload paths are slower
- MySQL is modestly better on announce and sequential write-heavy paths, with slightly slower reloads
- PostgreSQL shows the clearest benefit from the rework

Representative PostgreSQL results from the before/after comparison:

- `whitelist_add_seq`: about `+61%`
- `auth_key_add_seq`: about `+30%`
- `auth_key_reload`: about `+38%`

## Notes

This PR is intentionally broader than “just add a PostgreSQL driver”, because the required PostgreSQL support is tied to the persistence redesign itself.

If you would prefer, I can split this into smaller follow-up PRs after initial review, but I kept the branch together because the persistence trait split, async SQL migration, and PostgreSQL backend are tightly connected.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -16,6 +16,8 @@ jobs:
     env:
       CARGO_INCREMENTAL: "0"
       RUSTFLAGS: "-Cinstrument-coverage"
+      TORRUST_TRACKER_CORE_RUN_MYSQL_DRIVER_TEST: "true"
+      TORRUST_TRACKER_CORE_RUN_POSTGRES_DRIVER_TEST: "true"
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/generate_coverage_pr.yaml
+++ b/.github/workflows/generate_coverage_pr.yaml
@@ -16,6 +16,8 @@ jobs:
     env:
       CARGO_INCREMENTAL: "0"
       RUSTFLAGS: "-Cinstrument-coverage"
+      TORRUST_TRACKER_CORE_RUN_MYSQL_DRIVER_TEST: "true"
+      TORRUST_TRACKER_CORE_RUN_POSTGRES_DRIVER_TEST: "true"
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -137,6 +137,10 @@ jobs:
         name: Run MySQL Database Tests
         run: TORRUST_TRACKER_CORE_RUN_MYSQL_DRIVER_TEST=true cargo test --package bittorrent-tracker-core
 
+      - id: database-postgresql
+        name: Run PostgreSQL Database Tests
+        run: TORRUST_TRACKER_CORE_RUN_POSTGRES_DRIVER_TEST=true cargo test --package bittorrent-tracker-core
+
   e2e:
     name: E2E
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .env
+__pycache__/
+*.pyc
 *.code-workspace
 **/*.rs.bk
 /.coverage/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,17 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "ahash"
-version = "0.7.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
-dependencies = [
- "getrandom 0.2.17",
- "once_cell",
- "version_check",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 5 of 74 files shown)


josecelano (changes requested):
Hi @DamnCrab, thank you for your PR. It looks good, and I think it's a very valuable feature for the community.

I'm going to start a detailed review (I will write comments on code blocks and maybe another set of questions). These are my first impressions:

1. Things to fix before merging:
 
- Sign commits
- Pass all PR checks

2. Things I would recommend doing:

This is a big PR. As you mentioned, it would have been better to implement it progressively. However, this PR give us a good understanding of all things we need to change to support PostgreSQL. Now that we have learnt a lot I think it would be convenient to do an implementation plan with two main parts:

  -  Refactor plan: Refactor the current state to make the new feature easy to implement.
  - Add support for PostgreSQL

“Make the change easy, then make the easy change”.

Steps could be like something like this:

- Add more E2E tests (as you did). For example using a real client like you do. Everything that add tests should be included as soon as possible.
- Align Rust types with DB types (fix problems with misalignment between max u64 and the representation in the DB)
- Split the Database trait 
- Migrate current drivers to sqlx crate (without migrations)
- Introduce automatic migrations
- Add new driver for postgres

Notice that all steps could be merged independently from the new feature.

I understand that doing that now can be a big effort for you, so if you prefer to keep it all in one I will try to merge it as it's because It's a nice feature to have, but it can take me longer to carefully review it.

3. I have some initial questions (I haven't checked the code yet)

A) Are all migrations safe to execute in old databases?
B) Why the postgresql driver has more migrations? Should not all the drivers have the same?
C) I think it's a good idea to use a real client. I have to review that code. However I would prefer to use Rust instead of python. We are already using container for E2E testing in other parts. On the other hand, I wonder if the tracker client would be enough for the tests you are doing (I have not checked them yet). I'm planning tho change the tracker client to accept more arguments so we can simulate to be a bittorrent client. That should be simpler and faster for testing.

That's it for now. I may add some more comments.

josecelano:
Hi @DamnCrab, I prefer not to have python code:

- contrib/dev-tools/qa/run-qbittorrent-e2e.py
- contrib/dev-tools/qa/run-before-after-db-benchmark.py

Those are valuable tests but:

- We should port them to Rust (I do not want to maintain code in two languages for testing if there is no reason for it)
- They have some duplicate code
- Ideally we should run them in CI with workflows (if they are not too expensive)

josecelano:
Hi @DamnCrab, are you planning to continue working on this PR? I don’t want to lose all the work you’ve done so far, so I’m planning to start working on https://github.com/torrust/torrust-tracker/issues/1525 to basically implement what you’ve done, but incorporating my recommendations (of course, giving you credit).

josecelano:
Hi @DamnCrab I started working on adding your changes. First I have planned the changes in this [PR](https://github.com/torrust/torrust-tracker/pull/1702). After merging it I will start with the implementation of the first sub-issue.

josecelano:
Deprecated. See https://github.com/torrust/torrust-tracker/pull/1725

cc @DamnCrab

PR #1730, opened 2026-05-04
fix(tracker-core): fix race condition in global downloads persistence test

josecelano:
## Problem

The integration test `it_should_persist_the_global_number_of_completed_peers_into_the_database` was flaky under high-concurrency environments such as Docker builds (the container CI workflow).

**Root cause:** After calling `increase_number_of_downloads`, the test only awaited `yield_now()`, which was not enough time for the background event listener to finish persisting the download count to the SQLite database. A new `TestEnv` was then created immediately, reading from the same database before the write had completed — causing an intermittent assertion failure.

## Fix

Three changes:

1. **`packages/tracker-core/src/statistics/persisted/downloads.rs`** — Changed `load_global_downloads` visibility from `pub(crate)` to `pub` so the test helper can call it directly.

2. **`packages/tracker-core/tests/common/test_env.rs`** — Added `wait_for_global_downloads_persisted(expected)` helper that polls the database with a 50 ms interval and a 5-second timeout until the persisted download count reaches the expected value.

3. **`packages/tracker-core/tests/integration.rs`** — Called `test_env.wait_for_global_downloads_persisted(1).await` between the `increase_number_of_downloads` call and the simulated restart, ensuring the background DB write completes before the new `TestEnv` is created.

## Verification

- Pre-commit checks (linters + full test suite) passed locally.
- The fix eliminates the race by making the test deterministic instead of relying on timing assumptions.

### DIFF
--- a/packages/tracker-core/src/statistics/persisted/downloads.rs
+++ b/packages/tracker-core/src/statistics/persisted/downloads.rs
@@ -132,7 +132,7 @@ impl DatabaseDownloadsMetricRepository {
     /// # Errors
     ///
     /// Returns an [`Error`] if the underlying database query fails.
-    pub async fn load_global_downloads(&self) -> Result<Option<NumberOfDownloads>, Error> {
+    pub(crate) async fn load_global_downloads(&self) -> Result<Option<NumberOfDownloads>, Error> {
         self.database.load_global_downloads().await
     }
 }
--- a/packages/tracker-core/src/statistics/persisted/downloads.rs
+++ b/packages/tracker-core/src/statistics/persisted/downloads.rs
@@ -132,7 +132,7 @@ impl DatabaseDownloadsMetricRepository {
     /// # Errors
     ///
     /// Returns an [`Error`] if the underlying database query fails.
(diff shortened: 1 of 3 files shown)


josecelano:
ACK 4e34b9ebfb158a3c6c9d20d7c66de7bc0b6480c9

josecelano:
Addressed Copilot review suggestions in follow-up commit e19decfc.

What changed:
- Kept load_global_downloads as pub(crate) in packages/tracker-core/src/statistics/persisted/downloads.rs to avoid widening public API.
- Updated wait_for_global_downloads_persisted to poll via database_stores.torrent_metrics_store.load_global_downloads() in packages/tracker-core/tests/common/test_env.rs.

Validation:
- Ran cargo test --package bittorrent-tracker-core it_should_persist_the_global_number_of_completed_peers_into_the_database (pass).

Status:
- All Copilot review threads on this PR are resolved.

josecelano:
ACK e19decfce73e709f5c53385c2287a606aaa934db

PR #1741, opened 2026-05-07
ci(#1740): fix container workflow caching

### LINKED ISSUE #1740: Fix Container Workflow Caching
## Overview

The `container` workflow (`.github/workflows/container.yaml`) has multiple issues that prevent the GHA Docker layer cache from working reliably:

1. **`actions/checkout` runs after the build step** — On a cold cache, the job will either fail (no `Containerfile`) or use a stale workspace from a previous run
2. **Matrix targets share one cache namespace** — The `debug` and `release` build targets race and overwrite each other's cache entries
3. **`publish_development` and `publish_release` are missing `actions/checkout`** — These jobs cannot find `./Containerfile` on a cold runner
4. **All jobs share the same unscoped GHA cache** — Cross-job and cross-workflow collisions cause unnecessary cache misses
5. **`.dockerignore` is incomplete** — It doesn't exclude non-build files (`docs/`, `.coverage/`, CI configs), causing unnecessary cache busting on every doc/config change

## Root Cause

- Step ordering bug in `container.yaml` `test` job (checkout after build)
- Missing checkout steps in publish jobs
- Unscoped GHA cache entries causing collisions between parallel and sequential jobs
- No cache scoping between different workflows
- `docker-e2e` job in `testing.yaml` builds images without BuildKit cache

## Solution

This issue requires 6 fixes across two workflows. See `docs/issues/fix-container-workflow-caching.md` for detailed analysis and implementation steps:

1. Move `actions/checkout` to the first step in the `test` job
2. Add `scope=container-${{ matrix.target }}` to `cache-from/cache-to` in `test` job
3. Add `docker/setup-buildx-action` + `docker/build-push-action` pre-build step to `docker-e2e` job with `scope=testing-docker-e2e`
4. Extend `.dockerignore` to exclude non-build files (`docs/`, `.coverage/`, CI configs, etc.)
5. Add `actions/checkout` and scoped cache to `publish_development` and `publish_release` jobs
6. Add `--skip-build` flag to Rust E2E runners (`e2e_tests_runner`, `qbittorrent_e2e_runner`)

## Expected Outcome

After these fixes:
- Cold cache builds will not fail due to missing `Containerfile`
- `debug` and `release` builds will have independent, warm caches
- Publish jobs will have reliable caches
- Doc and config changes will not trigger full container rebuilds
- E2E tests will reuse the pre-built image instead of rebuilding it
- A second run should show cache hits for the `cargo chef cook` layer (~5-10 min savings per build)

## Related Issues

- #1726 — Reduce Build Times with sccache

josecelano:
## Summary
- fix checkout ordering and cache scoping in `.github/workflows/container.yaml`
- prebuild the tracker image in `.github/workflows/testing.yaml` and reuse it in E2E runners with `--skip-build`
- tighten `.dockerignore` and document the non-obvious cache-related `Containerfile` steps

## Validation
- `cargo check --bin e2e_tests_runner --bin qbittorrent_e2e_runner`
- `linter yaml`
- `linter all`

This branch contains four separate commits:
- `docs(#1740): add container workflow caching issue spec`
- `ci(container): fix checkout and cache scoping`
- `ci(e2e): reuse prebuilt tracker image`
- `chore(container): reduce context churn and document cache`


josecelano:
ACK 57c1ed615fe07ca54d5ca002c7ca57235930b7e4

PR #1850, opened 2026-06-01
fix(containerfile): add --locked flag to cargo binstall calls for cargo-nextest

josecelano:
## Problem

The `Container` workflow started failing after the last merged PR with:

```
error: Nextest does not support being installed without --locked. To install nextest from source, run:
       cargo install --locked cargo-nextest
```

`cargo-nextest` enforces installation with `--locked` at compile time via the `locked-tripwire` crate. When `cargo-binstall` cannot find a pre-built binary and falls back to source compilation, it calls `cargo install` without `--locked`, triggering this compile-time error.

## Fix

Add `--locked` to both `cargo binstall` invocations in the `Containerfile` (the `chef` and `tester` image stages). This flag is forwarded by `cargo-binstall` to the underlying `cargo install` call when falling back to source compilation.

## Changes

- `Containerfile`: `cargo binstall --no-confirm` → `cargo binstall --no-confirm --locked` (both occurrences)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -6,7 +6,7 @@
 FROM docker.io/library/rust:trixie AS chef
 WORKDIR /tmp
 RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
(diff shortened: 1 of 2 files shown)


josecelano:
ACK 0baeabac889c8c384cab6d0c8b034d076615541e

PR #1863, opened 2026-06-01
feat(container): restrict recipe stage to manifest-only COPY to prevent spurious cook cache invalidation

### LINKED ISSUE #1852: Restrict recipe stage to manifest-only COPY to prevent spurious cook cache invalidation
## Goal

Prevent the `cargo chef cook` (dependency) layers from being invalidated on every source code change by replacing the full-tree `COPY . /build/src` in the `recipe` stage with a manifest-only copy of `Cargo.toml` and `Cargo.lock` files.

## Background

The current `recipe` stage does:

```dockerfile
FROM chef AS recipe
WORKDIR /build/src
COPY . /build/src          # copies the entire source tree
RUN cargo chef prepare --recipe-path /build/recipe.json
```

The `cargo chef prepare` command only reads `Cargo.toml` manifests and `Cargo.lock` to build `recipe.json`. It does not read any `.rs` source files. This is explicitly stated in `cargo-chef`'s own CLI description:

> "Analyze the current project to determine the **minimum subset of files (Cargo.lock and Cargo.toml manifests)** required to build it and cache dependencies"

Because `COPY . /build/src` copies all source files into the recipe stage, Docker invalidates that layer's cache whenever **any tracked file changes** — including `.rs` files, documentation, and scripts. Since the recipe stage is upstream of both `dependencies` and `dependencies_debug` cook stages, this cascades into a full external dependency recompile (~200–400 s) on every PR that only changes application code.

## The Fix

Replace the full-tree copy with a manifest-only copy in the recipe stage — one `COPY` line per workspace member `Cargo.toml`, plus the root `Cargo.lock`. After this change, the recipe stage cache (and therefore the cook layers) is only invalidated when `Cargo.toml` or `Cargo.lock` actually changes — not on every `.rs` edit.

## Maintenance Cost

The manifest-only COPY list must be kept in sync with the workspace member list in the root `Cargo.toml`. The workspace is relatively stable (packages are being extracted to separate repos under EPIC #1669, reducing the list over time). A CI check can catch drift automatically.

## Acceptance Criteria

- [ ] AC1: The `recipe` stage uses only manifest-only COPY lines — no `COPY . /build/src`.
- [ ] AC2: `cargo chef prepare` produces a `recipe.json` equivalent to the current output.
- [ ] AC3: A warm-cache rebuild triggered by a `.rs`-only change no longer invalidates the cook layers.
- [ ] AC4: All Containerfile targets build successfully end-to-end.
- [ ] AC5: Build time improvement for `.rs`-only changes is measured and documented.
- [ ] AC6: A mechanism exists to detect workspace-member drift from the Containerfile manifest list.

## Related

- Parent EPIC: #1840
- Baseline analysis: #1841
- Companion issue: #1851 (reduces context size; this issue fixes the recipe-layer invalidation)
- Full spec: `docs/issues/open/` (after spec PR merge)

josecelano:
## Summary

Closes #1852

Replaces the full-tree `COPY . /build/src` in the `recipe` stage of the `Containerfile` with individual `COPY <manifest> <dest>/` lines for every workspace member.

## Problem

The `cargo chef prepare` command only needs `Cargo.toml` manifests and `Cargo.lock` to produce `recipe.json`. However, the previous `COPY . /build/src` copied the entire source tree into the recipe stage, causing Docker to invalidate that layer — and the expensive downstream `cargo chef cook` dependency layers — on every source-code change.

## Solution

Replaced the single full-tree copy with manifest-only COPY instructions for all 28 workspace members plus root `Cargo.toml` and `Cargo.lock`. Added a maintenance comment explaining the sync requirement.

## Effect

- Cook layers (`dependencies` and `dependencies_debug`) now stay cached for source-only edits (no `Cargo.toml` / `Cargo.lock` changes).
- Cook layers are correctly invalidated when manifests change.
- Warm container builds for typical PRs should avoid the full external dependency recompilation cost (~200–400 s).

## Maintenance

The manifest list must be kept in sync with the workspace members in `Cargo.toml`. A comment block in the Containerfile documents this requirement.

## Checklist

- [x] `linter all` exits `0`
- [x] Pre-commit hook passes
- [x] All workspace member `Cargo.toml` files are listed (verified with `cargo metadata`)
- [x] Maintenance comment added to Containerfile
- [x] Issue spec progress updated

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -41,9 +41,14 @@ WORKDIR /build/src
 # therefore the expensive `cargo chef cook` dependency layers) on every source-code change.
 # By copying only manifests, the cook layers stay cached for source-only edits.
 #
-# MAINTENANCE: Keep this list in sync with the workspace members in the root Cargo.toml.
-# Every new workspace package must have a corresponding COPY line here; every removed or
-# moved package must have its line removed or updated accordingly.
+# MAINTENANCE: Keep this list in sync with all in-repo path crates (packages/, console/,
+# contrib/). This includes the root crate itself plus every crate reachable as a path
+# dependency from the root — i.e. all packages discovered by `cargo metadata --no-deps`
+# whose manifest path is inside this repository. Note: the `[workspace].members` key in
+# the root Cargo.toml only lists packages not auto-discovered via path dependencies; it
+# is a much smaller set and should NOT be used as the authoritative list here.
+# Every new in-repo path crate must have a corresponding COPY line added; every removed
+# or moved crate must have its line updated or removed accordingly.
 COPY Cargo.toml Cargo.lock ./
 COPY console/tracker-client/Cargo.toml console/tracker-client/
 COPY contrib/bencode/Cargo.toml contrib/bencode/
--- a/Containerfile
+++ b/Containerfile
@@ -78,30 +78,106 @@ COPY packages/tracker-core/Cargo.toml packages/tracker-core/
 COPY packages/udp-protocol/Cargo.toml packages/udp-protocol/
 COPY packages/udp-server/Cargo.toml packages/udp-server/
 COPY packages/udp-tracker-core/Cargo.toml packages/udp-tracker-core/
-# Create stub source files for targets that are EXPLICITLY declared in Cargo.toml.
-# `cargo chef prepare` runs `cargo metadata` internally, which fails with an error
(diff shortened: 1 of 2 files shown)


josecelano:
ACK 2af6803432c935cc21ce8fda2ca8c8a5a45c9cb7

Issue #1669, opened 2026-03-06
Overhaul: packages


josecelano:
Relates to: https://github.com/torrust/torrust-tracker/issues/1659

1. Review their descriptions (README)
2. Review their boundaries (are they loosely coupled and highly cohesive?)
3. Review whether we should fix all of them to the same version or evolve their version separately.
4. Review if we should move them to separate repos (the ones that are more decoupled and have more potencial to be reused)
5. Review potencial performance problems for new contributors loading such a huge repo into their IDEs
6. ...

Package diagrams (on internal Miro):

- https://github.com/torrust/torrust-tracker/tree/develop/docs/media/packages, but others, like the following, have not been published:

<img width="2081" height="1310" alt="Image" src="https://github.com/user-attachments/assets/f5265d33-05cb-4abf-b246-995f4895dacb" />

I'm planning to run other tools like:

- https://sr.ht/~jplatte/cargo-depgraph/
- https://github.com/abhigyanpatwari/GitNexus
- https://codescene.io/

cc @da2ce7

josecelano:
Some screenshots from codescene.io:

<img width="1488" height="1156" alt="Image" src="https://github.com/user-attachments/assets/45aeccfc-d43a-43a7-a8cd-10738964b564" />

<img width="1488" height="1156" alt="Image" src="https://github.com/user-attachments/assets/d753f6a0-a171-4bf0-8319-dbce39b61b02" />

<img width="1488" height="1156" alt="Image" src="https://github.com/user-attachments/assets/fcb69d08-46f3-4bb7-939b-797f55e47eb9" />

<img width="1477" height="1302" alt="Image" src="https://github.com/user-attachments/assets/06b1b9c8-5946-4bbe-8753-7ae59951abc7" />

josecelano:
Relates to: https://github.com/torrust/torrust-bittorrent/issues/64#issuecomment-4522399982

PR #1877, opened 2026-06-04
feat(primitives): replace TORRENT_PEERS_LIMIT with AnnouncePolicy::max_peers_per_announce

### LINKED ISSUE #1864: Review and refactor `TORRENT_PEERS_LIMIT`: hardcoded constant vs. config option
## Context

Issue #1859 moved `TORRENT_PEERS_LIMIT` (value: `74`) and `TrackerPolicy` from `torrust-tracker-configuration` into `torrust-tracker-primitives`. That was the right first step to break a dependency cycle, but the constant is still a global hardcoded value spread across several packages.

## Problem

`TORRENT_PEERS_LIMIT` is currently used in at least three distinct roles:

1. **Cap inside `From<i32/u32> for PeersWanted`** — enforces a maximum on client-requested peers at parse time:
   ```rust
   // packages/tracker-core/src/announce_handler.rs
   impl From<i32> for PeersWanted {
       fn from(value: i32) -> Self {
           ...
           PeersWanted::Only { amount: amount.min(TORRENT_PEERS_LIMIT) }
       }
   }
   ```
   Because this is a `From` impl, there is no way to inject a runtime limit — the constant is baked in.

2. **Default return count in `PeersWanted::limit()`** — used when the client requested `AsManyAsPossible`.

3. **Query cap in repository `get_peers` / `get_swarm_peers`** — used in `in_memory.rs` and `swarm/registry.rs` to limit how many peers are returned per query.

The constant living in `primitives` is better than in `configuration`, but it is still a global that couples unrelated packages to a single magic number.

## Questions to Resolve

- Should `TORRENT_PEERS_LIMIT` remain a single global constant, or should each package define its own local default?
- Should the cap be a runtime configuration option (e.g., `TrackerPolicy::max_peers_per_announce`) instead of a compile-time constant? This would make it tunable per deployment without recompilation.
- For the `From<i32> for PeersWanted` trait impl, which cannot accept injected state, is a package-local constant the right answer, or should the trait impl be replaced by a free function / builder that accepts the limit explicitly?
- If it becomes a config option, where does it live in the configuration hierarchy, and how do the repository query methods receive it?

## Possible Approaches

| Approach | Pros | Cons |
|---|---|---|
| Keep global constant in `primitives` (status quo post-#1859) | Simple, no API churn | Magic number, not tunable, couples packages |
| Move constant into each consuming package | Removes cross-package coupling | Duplication, value can drift |
| Make it a `TrackerPolicy` field with a default | Runtime-tunable, operator-visible | Requires plumbing through announce handler and repositories |
| Replace `From` impls with explicit constructors | Removes implicit global dependency | API change for callers |

## Related

- Closes (or follows from) #1859
- Parent EPIC: #1669 — Overhaul: Packages
- `packages/tracker-core/src/announce_handler.rs` — `PeersWanted`, `From` impls
- `packages/tracker-core/src/torrent/repository/in_memory.rs` — `get_peers`, `get_swarm_peers`
- `packages/swarm-coordination-registry/src/swarm/registry.rs` — same
- `packages/axum-http-server/src/lib.rs` — doc reference to the constant

## Expected Output

A decision (either an ADR or a `DECISIONS.md` entry under the #1669 EPIC) recording which approach is chosen, followed by the implementation if the decision is to change the current design.

josecelano:
## Summary

Closes #1864 — sub-task of EPIC #1669.

Move the peer-count cap from a global compile-time constant (`TORRENT_PEERS_LIMIT = 74`) to a configurable field `AnnouncePolicy::max_peers_per_announce` (default 74, serialized via serde). This makes the limit observable in the configuration file and overridable at runtime without recompilation.

## Changes

- **Remove** `TORRENT_PEERS_LIMIT` from `primitives::policy` and its re-export from `primitives::lib`
- **Add** `AnnouncePolicy::max_peers_per_announce: usize` (serde default = 74, stored in `[core.announce_policy]` config section)
- **Refactor** `PeersWanted` in `tracker-core`:
  - Remove `From<i32>` and `From<u32>` impls (were silently applying the global cap at construction time)
  - Add `PeersWanted::from_client_request(i32)` — `≤0` → `AsManyAsPossible`, positive → `Only { amount }`
  - `PeersWanted::limit()` now takes an explicit `max_peers: usize` argument, applied at announce time from `announce_policy.max_peers_per_announce`
- **Fix** `get_torrent_peers` — now takes `max_peers: usize` instead of using the global constant (old implementation had a bug: `max(limit, TORRENT_PEERS_LIMIT)` returned _more_ than requested)
- Update all call sites, tests, configuration fixtures, and docs
- Record decision as **DEC-10** in `docs/issues/open/1669-overhaul-packages/DECISIONS.md`

## Configuration

The new field appears in `[core.announce_policy]`:

```toml
[core.announce_policy]
interval = 120
interval_min = 120
max_peers_per_announce = 74
```

Existing configs without the field continue to work (serde default = 74).

## Testing

- All workspace tests pass (`cargo test --workspace`)
- Pre-push checks pass (nightly format, nightly check, nightly docs, full stable test suite)

### DIFF
--- a/docs/issues/open/1669-overhaul-packages/DECISIONS.md
+++ b/docs/issues/open/1669-overhaul-packages/DECISIONS.md
@@ -20,6 +20,52 @@ the proposal, the reasoning, and a reference to any supporting artifact.
 
 ---
 
+## DEC-10 — Move peer-count cap from a global constant to `AnnouncePolicy::max_peers_per_announce`
+
+**Date**: 2026-06-09
+**Status**: Adopted
+**Related issue**: [#1864](https://github.com/torrust/torrust-tracker/issues/1864)
+
+### Proposal considered
+
+The hardcoded constant `TORRENT_PEERS_LIMIT = 74` in `torrust-tracker-primitives` was
+the sole compile-time control over how many peers the tracker returns per announce
+response. The options evaluated were:
+
+1. **Keep the constant** but expose it as a public API so callers can pass it explicitly.
+2. **Add a runtime field** `max_peers_per_announce: usize` to `AnnouncePolicy` and remove
+   the constant entirely.
+3. **Move the cap to `TrackerPolicy`** alongside the existing cleanup policy fields.
+
+### Alternative chosen
+
+Option 2: add `max_peers_per_announce: usize` (default `74`) to `AnnouncePolicy` and
+remove `TORRENT_PEERS_LIMIT`. The cap is applied inside `AnnounceHandler::build_announce_data`
+via `PeersWanted::limit(max_peers)` at call time, not at `PeersWanted` construction time.
+
+### Why this alternative was adopted
+
+1. **Semantic fit**: `AnnouncePolicy` already governs announce-response behaviour
+   (`interval`, `interval_min`). The peer count cap belongs in the same bucket.
+2. **Runtime configurability**: operators can tune the cap per deployment without
+   recompiling. The previous constant made that impossible.
+3. **Cleaner type boundaries**: `PeersWanted` no longer needs to know about a global
+   limit when constructed; the limit is injected once at the point of use
+   (`build_announce_data`), keeping the type simple and context-free.
+4. **Avoiding `TrackerPolicy` scope creep**: `TrackerPolicy` is about data-retention
+   behaviour (persistence, ghost peers, etc.). Mixing in a response-size limit there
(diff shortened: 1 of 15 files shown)


josecelano:
ACK 388e3d1f05e0cd9562ad80dd43b2a099a1177fb4

PR #1948, opened 2026-06-26
refactor(workspace-coupling): parse imports with syn AST and emit NDJSON

### LINKED ISSUE #1805: Overhaul workspace-coupling report tool: replace regex scanner with `syn` and adopt CLI output contract
## Goal

Replace the regex-based import scanner in the `workspace-coupling` analysis tool with a
`syn`-based Rust AST parser to correctly extract imported items from all `use` statement
forms, and bring the tool's CLI output into compliance with the global CLI output contract
(ADR `20260519000000_define_global_cli_output_contract`) by replacing plain-text `eprintln!`
calls with structured JSON NDJSON records on stderr.

## Background

The `workspace-coupling` tool uses a regex to extract imports:

```
{module_name}::[A-Za-z_][A-Za-z0-9_]*(?:::[A-Za-z_][A-Za-z0-9_]*)?
```

This regex misses at minimum two legitimate patterns:

1. **Brace-import groups**: `use torrust_tracker_contrib_bencode::{BMutAccess, ben_int, ben_map}`
2. **Re-export statements**: `pub use bittorrent_peer_id::{PeerClient, PeerId}`

This causes six confirmed false-negative "Items not extracted" entries in the coupling report.
Patching the regex would leave the tool fragile; replacing it with `syn` AST parsing handles all
valid `use` forms by construction.

Additionally, the tool's `eprintln!` calls violate ADR `20260519000000_define_global_cli_output_contract`
(section 1: all stderr records must be JSON NDJSON). The migration policy (section 10) requires
migration when the file is substantially touched — this rewrite qualifies.

The tool also has no tests. As part of this overhaul, unit tests (for a pure
`parse_imports_from_source` function) and integration tests (via `std::process::Command` against
a fixture workspace) will be added, following TDD order.

## Spec

Full specification: `docs/issues/open/` (see spec file for this issue number).

## Related

- Part of #1669 (EPIC — Overhaul Packages)
- See also: companion issue for `cargo machete --with-metadata` and unused dev dep removal

mvanhorn:
## Summary

The `workspace-coupling` analysis tool found cross-crate imports with a regex scan over source text and printed its diagnostics as ad-hoc stderr lines. Regex matching over Rust source is brittle: it misclassifies imports inside strings or comments, and it has no real model of `use` trees. This replaces the scanner with a `syn`-based AST parser and moves the parsing into a `lib.rs` so it is unit-testable, and adopts a line-delimited JSON (NDJSON) output contract so the tool's results can be consumed programmatically.

## Why this matters

Issue #1805 asks for the regex scanner to be replaced with a proper AST parse and for the tool to emit a structured output contract. Parsing the actual `use` items with `syn` removes the false positives/negatives inherent to text matching, and NDJSON makes the coupling report machine-readable instead of something a human has to eyeball in stderr. To avoid regressing coverage the old text scan happened to have, the parser also reports root-level crate aliases (`use dep_crate as alias;`) and dependency paths that appear only inside macro invocations (`matches!(dep_crate::Thing::A, ...)`), which a naive `syn` visitor would otherwise miss.

## Changes

The import-extraction logic now lives in `src/lib.rs` as pure functions (`parse_imports_from_source` / `try_parse_imports_from_source`) that walk the `syn` AST, including `UseTree` renames and macro token streams, so the binary in `main.rs` only handles file discovery, invocation, and NDJSON emission. `Cargo.toml` gains the `syn` dependency. The change is confined to the `contrib/dev-tools/analysis/workspace-coupling` tool; no tracker code is affected.

## Testing

Added `tests/parse_imports.rs` covering grouped/nested/aliased use trees, root-alias-only imports, macro-embedded dependency paths, and the binary's NDJSON output. Run with `cargo test -p workspace-coupling` (9 tests pass) and `cargo clippy -p workspace-coupling --all-targets -- -D warnings` (clean).

Closes #1805


josecelano:
ACK 12fa21f225bcc84b5a883b40a51189584ca4bbd7

PR #2025, opened 2026-07-22
Add protocol-aware service bindings to server logs

### LINKED ISSUE #1415: Make server listen URL available where server socket address is also available
We use the service socket addresses in many places (logs, events, etc.). However we are only using the server URL in few places like logging. When you start the tracker you can see something like:

```output
2025-03-26T10:35:01.306728Z  INFO run_with_graceful_shutdown{cookie_lifetime=120s}: UDP TRACKER: Started on: udp://0.0.0.0:6868
2025-03-26T10:35:01.306818Z  INFO run_with_graceful_shutdown{cookie_lifetime=120s}: UDP TRACKER: Started on: udp://0.0.0.0:6969
2025-03-26T10:35:01.307005Z  INFO start: HTTP TRACKER: Started on: http://0.0.0.0:7070
2025-03-26T10:35:01.307072Z  INFO start: HTTP TRACKER: Started on: http://0.0.0.0:7171
2025-03-26T10:35:01.307247Z  INFO start: API: Started on http://0.0.0.0:1212
2025-03-26T10:35:01.307368Z  INFO start:start_job: HEALTH CHECK API: Started on: http://127.0.0.1:1313
```

We want to add the service URL in two new places:

1. [The information exposed via the health check API](https://github.com/torrust/torrust-tracker/issues/1409).
2. [The metrics](https://github.com/torrust/torrust-tracker/pull/1414).

We need to add the service URL whenever we are passing the service socket address. We need to store this:

- `udp://127.0.0.1:6969`
- `http://127.0.0.1:7070/announce`
- `https://127.0.0.1:7171/announce`
- `http://127.0.0.1:1212`
- `http://127.0.0.1:1313`

Instead of just this:

- `127.0.0.1:6969`
- `127.0.0.1:7070`
- `127.0.0.1:7171`
- `127.0.0.1:1212`
- `127.0.0.1:1313`

We have to options:

### Option 1

Replace the `SocketAddr` by `Url`

Pros:

- More generic.
- It will support more services running in new protocols.
- We can use the URL directly when needed.

Cons:

- We have to clone the `SocketAddr` from the Url when we only need that. However we are probably already doing that.

In the Tracker client we are already using a `ServiceUrl` which is a Url:

```rust
use reqwest::Url as ServiceUrl;

pub struct Configuration {
    pub udp_trackers: Vec<ServiceUrl>,
    pub http_trackers: Vec<ServiceUrl>,
    pub health_checks: Vec<ServiceUrl>,
}
```

### Option 2

Just add the `scheme` argument together with `SockerAddr`

Cons:

- More params in the function signatures.

I will implement option 1.

@nuts-rice, I think this is what is needed to finish the implementation of issue https://github.com/torrust/torrust-tracker/issues/1409. I'm working on this issue now.

cc @da2ce7 

### IMPORTANT

The tracker can run behind a reverse proxy so, the service public and internal URLs can be different. For example, in the tracker demo:

- Public service URL: https://tracker.torrust-demo.com/announce (SSL)
- Container service URL: http://tracker:7070/announce (NO SSL, how the Index reach the tracker).
- Internal service URL: http://0.0.0.0:7070/announce (NO SSL)

What I'm going to add in this issue is the last one: the internal service URL. It's the URL that the application can know.

For the public service URL, we would need to add it to the configuration.

The internal service URL could also be public is the tracker is not running behing a proxy.

I'm going to use the name `internal_service_url`.

josecelano:
## Summary

Add protocol-aware `service_binding` fields to server operational logs while retaining the existing `server_socket_addr` field for compatibility.

- HTTP tracker and REST API request/response logs now include the post-bind `ServiceBinding`.
- UDP error logs now include the post-bind `ServiceBinding`.
- The REST API router now receives `ServiceBinding` and derives the compatible socket-address field from it.
- Record manual runtime evidence and clarify that both fields describe the same actual bound address, including an OS-assigned port.

## Verification

- `cargo fmt --check`
- `linter all`
- `cargo test -p torrust-tracker-axum-http-server -p torrust-tracker-axum-rest-api-server -p torrust-tracker-udp-server`
- `cargo test --workspace`
- `TORRUST_GIT_HOOKS_LOG_DIR=.tmp ./contrib/dev-tools/git/hooks/pre-commit.sh`

Closes #1415


josecelano:
ACK 58504af7cc8360757018cdab4e503ee5ab75dec2

PR #2050, opened 2026-07-31
Feat/i2p peer support

Frigyes06:


josecelano:
Hi @Frigyes06, than your for you contribution. I was on holiday and I have just seen the PR. I did not know [i2p](https://i2p.net/en/docs/applications/bittorrent/), so it would take me a while to review the PR. It look good!

josecelano (commented):
Hi @Frigyes06 — thanks again for the I2P implementation.

I prepared a rebased, GPG-signed draft PR that supersedes this one: #2059. It preserves your authorship for the implementation commits and includes the full review report plus reproducible manual-test evidence in `docs/pr-reviews-manual/pr-2050/`.

The main items to address before merge are:

1. Percent-decode I2P Destination query values so valid `%3D%3D` Base64 padding is accepted (a disabled regression test is prepared).
2. Add I2P-aware compact-response deserialization; do not parse 32-byte Destination hashes as IPv4 peer entries.
3. Preserve structured I2P parsing errors instead of returning only generic `InvalidParam`.
4. Bound and structurally validate untrusted Destination input before decoding, and avoid reflecting the full value in error responses.

Please use the #2059 branch as the base for follow-up fixes. The report also records planned follow-up work for the tracker client and the REST API redesign.

josecelano:
@da2ce7 in general, we should consider if we want to support the protocol. There are some edge cases that can mislead "clearnet" peers (for example, reusing the `ip` announce GET param to provide the peer destination in the I2P network). We also have to decide how the API should handle this new type of peers.

I don't see any major issue to not support it. I think it could even make the code a little bit more generic to accept other types of peers in the future.

josecelano:
Hi @Frigyes06 — the signed replacement draft, [PR #2059](https://github.com/torrust/torrust-tracker/pull/2059), now includes a dedicated [Destination spoofing analysis](https://github.com/josecelano/torrust-tracker/blob/2050-i2p-peer-support-reviewed/docs/pr-reviews-manual/pr-2050/destination-spoofing-analysis.md).

This identifies a **merge-blocking trust-boundary problem**: a normal HTTP announce does not prove that a client owns the I2P Destination supplied in `ip`. A client can claim a victim Destination, obtain I2P swarm information, and update or remove that peer record.

The unverified-identity model also permits **fabricated-Destination spam**: many announces using distinct but attacker-unowned Destinations can pollute swarms and consume registry, CPU, persistence, cleanup, metrics, and response resources.

Before the final I2P-support PR can merge, please choose and implement one safe policy:

1. **Trusted transport-derived identity:** Support I2P announces only through a dedicated trusted listener. Derive the peer Destination from validated I2P transport context—for example, a trusted server-tunnel identity header or a SAMv3/I2CP adapter—and reject direct/untrusted requests and query/header mismatches.
2. **Disable I2P announces:** Reject I2P Destinations until transport-derived identity enforcement is available.

For Policy 1, also document and test the trust boundary: configured trusted source, loopback-only forwarding listener, missing or malformed identity context, matching and mismatching query Destinations, direct-listener access, and attempted peer-record takeover.

Identity enforcement must be complemented by bounded input, rate limits keyed by trusted identity/source, peer-admission and response limits, expiry/cleanup, and privacy-aware abuse metrics.

Please use PR #2059 as the base for follow-up work. Its updated description and the review documents list all mandatory actions A1–A7. The REST representation remains deferred to epic #144 and an ADR.

josecelano:
i2p support is delay until we release version 4.0.0. We are focused on releasing that new version.

Issue #2195, opened 2026-09-10
Port the Maintainer Merge Wrapper to a Repository-Neutral Python Program


da2ce7:
## Goal

Replace `contrib/dev-tools/git/merge-pull-request.sh` with a Python program beside the vendored engine that performs the same preflight and the same delegation, reads every repository-specific value from Git configuration instead of holding it as a constant, and therefore runs unchanged in any repository that has configured the merge workflow.

## Background

The vendored engine `contrib/dev-tools/git/github-merge.py` bakes in no repository identity. It reads `githubmerge.repository` (line 543), `githubmerge.host` (line 544), `githubmerge.branch` (line 545), `githubmerge.merge-author-email` (line 546), `githubmerge.testcmd` (line 547), `user.ghtoken` (line 548) and `user.signingkey` (line 549) from Git configuration, and its own help text (lines 513-522) documents them as the configuration surface. Its one Torrust-named constant, `SYMLINK_DECLARATION_NAMESPACE` (line 240), names the declaration format rather than a repository, which the comment above it and the format documentation both state; every repository adopting the workflow carries the same value. The engine is configuration-driven all the way down.

The wrapper is not. `merge-pull-request.sh` holds `EXPECTED_REPOSITORY="torrust/torrust-tracker"` (line 11), `TARGET_BRANCH="develop"` (line 12) and `SYMLINK_DECLARATION=".symlinks.json"` (line 13), and `require_repository_configuration` (lines 38-50) exits unless `githubmerge.repository` is exactly equal to the first of them. The same string is asserted in the wrapper's own test suite (`contrib/dev-tools/git/tests/test-merge-pull-request.sh` lines 44, 65, 109, 134, 271), in the sibling engine suite's fixture (`contrib/dev-tools/git/tests/test-github-merge-symlinks.py` line 30), in `contrib/dev-tools/git/README-github-merge.md` (lines 4-6) and in `.github/skills/dev/git-workflow/merge-pull-request/SKILL.md` (lines 21-22, 63, 70).

[...]

PR #2196, opened 2026-09-10
docs(issues): add specification for #2195

da2ce7:
Related to #2195

## Summary

This pull request adds the reviewed specification for issue #2195 and contains no implementation, so it does not resolve the issue.

The specification replaces `contrib/dev-tools/git/merge-pull-request.sh` with `contrib/dev-tools/git/merge-pull-request.py`, a repository-neutral program beside the vendored engine that performs the same preflight and the same delegation. The engine already reads every repository fact from Git configuration; the wrapper does not, holding `EXPECTED_REPOSITORY="torrust/torrust-tracker"`, `TARGET_BRANCH="develop"` and `SYMLINK_DECLARATION=".symlinks.json"` as constants and refusing to run anywhere else. The concrete cost is `torrust/torrust-index`, which carries the symbolic link that made the declaration mechanism necessary but ships no copy of `contrib/dev-tools/git`, so a maintainer there invokes the engine directly and loses the argument check, the clean-tree check, the current-branch check, the repository-local signing message, the terminal guard, `--dry-run`, and the `--symlinks` argument without which no declaration is processed at all.

## What the specification decides

The repository is validated by shape, never by value: `githubmerge.repository` is required and must match `<owner>/<repo>`, and AD2 states why the equality check that is given up was never a safety property, since the engine builds its fetch remote from the same locally configured value and displays the merge details before anything is signed.

The target branch comes from `githubmerge.branch` with a documented default of `develop`, and the declaration path from `githubmerge.symlinks` with a documented default of `.symlinks.json`, validated with the rule the engine applies to `--symlinks` and passed unconditionally, because an off switch would let a repository lift a refusal by editing configuration instead of by committing a reviewed declaration.

Delegation uses `os.execv` on `sys.executable`, so the engine's exit code remains the command's exit code and the engine runs under the interpreter that just executed the preflight. The interpreter-on-`PATH` check becomes a documented and program-checked minimum version, and the bash suite's `script(1)` probe disappears with its dependency once the pseudo-terminal comes from `pty.openpty()`.

The bash suite is replaced by `contrib/dev-tools/git/tests/test-merge-pull-request.py` in the conventions of the existing engine suite, and a behavior mapping table accounts for every behavior of the shell wrapper and every one of its thirteen cases, marking exactly two as dropped with the reason each stops existing and exactly one as deliberately changed.

The specification also plans an ignore rule for `__pycache__/`, which `.gitignore` carries for no Python file today: a cache written beside the program leaves an untracked directory in `contrib/dev-tools/git/`, and the wrapper's own clean-tree check refuses a tree that carries one.

## Files touched

Only `docs/issues/` and `project-words.txt`. Against `develop` the branch adds one document, `docs/issues/open/2195-merge-wrapper-python-port/ISSUE.md`, and three dictionary entries, `execv`, `openpty` and `pycache`; no source, test, workflow or manifest is changed here. The port itself follows in a separate implementation pull request from the reserved `2195-merge-wrapper-python-port` branch.

## Validation performed

Six documentation gates were run at the branch tip `00a4bfe8`, all green: `linter markdown` (5.8 s), `linter cspell` (8.7 s), `linter yaml` (0.4 s), `linter toml` (0.1 s), `linter shellcheck` (1.4 s), and `lychee --offline` over the new document (0.05 s). The branch changes no Rust, so `clippy` and `rustfmt` have nothing to report on it and were not run.

The document moved from `docs/issues/drafts/merge-wrapper-python-port/` to `docs/issues/open/2195-merge-wrapper-python-port/`, which is the same directory depth, so its two relative links, to `contrib/dev-tools/git/README-github-merge.md` and to the `merge-pull-request` skill, are unchanged and still resolve; lychee confirms this independently. `git diff --check` over the branch range against `develop` reports no whitespace defects.

### DIFF
--- a/docs/issues/drafts/merge-wrapper-python-port/ISSUE.md
+++ b/docs/issues/drafts/merge-wrapper-python-port/ISSUE.md
@@ -6,102 +6,208 @@ priority: p2
 epic: null
 github-issue: null
 spec-path: docs/issues/drafts/merge-wrapper-python-port/ISSUE.md
-branch: "{issue-number}-{short-description}"
+branch: "{issue-number}-merge-wrapper-python-port"
 related-pr: null
-last-updated-utc: YYYY-MM-DD HH:MM
+last-updated-utc: 2026-09-10 13:22
 semantic-links:
   skill-links:
     - create-issue
   related-artifacts:
-    - .github/skills/dev/planning/create-issue/SKILL.md
+    - contrib/dev-tools/git/merge-pull-request.sh
+    - contrib/dev-tools/git/github-merge.py
+    - contrib/dev-tools/git/tests/test-merge-pull-request.sh
+    - contrib/dev-tools/git/tests/test-github-merge-symlinks.py
+    - contrib/dev-tools/git/README-github-merge.md
+    - .github/skills/dev/git-workflow/merge-pull-request/SKILL.md
+    - docs/issues/open/2175-merge-tool-symlink-exceptions/ISSUE.md
+    - docs/issues/closed/2022-vendor-and-document-maintainer-merge-workflow/ISSUE.md
 ---
 
 <!-- skill-link: create-issue -->
 
-# Issue #[To be assigned] - Port the maintainer merge wrapper to a repository-neutral Python program
+# Issue #[To be assigned] - Port the Maintainer Merge Wrapper to a Repository-Neutral Python Program
 
 ## Goal
 
-Describe the expected outcome in one or two sentences.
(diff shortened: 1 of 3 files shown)


josecelano:
I've commented on https://github.com/torrust/torrust-tracker/issues/2195#issuecomment-5620660655. If you are using the script in Torrust Index, you should consider build it there and we can copy it later to the Torrust Tracker repo. But I totally discourage using bash or python if not temporarily or justified.

da2ce7:
Closing without merge: #2195 is deferred until the post-4.0.0 Rust automation work under #2003 decides where the merge-workflow preflight lives. The behaviour table and configuration design in this specification are language-independent and remain available on this branch for that port.

